PR #8, opened 2018-04-16
Allow init & unseal in one go

avoidik:


avoidik:
@bonifaido it was too fast I think :) `InitRootToken` and `StoreRootToken` will be set as defaults (to `""` and `false`) if it's required to use non-defaults, then I can revise this PR

bonifaido:
Hi @avoidik, sorry, please open a new PR, until then I have reverted the changes from master.

PR #11, opened 2018-04-16
Allow init & unseal in one go with additional parameters

avoidik:


avoidik:
@bonifaido I've removed that block

PR #12, opened 2018-04-18
Add missing PFlag lookup

avoidik:
Without PFlag lookup `--init` command-line parameter has no effect


avoidik:
@matyix my apologies, reworked the commit

PR #44, opened 2018-05-07
Vault Operator initial version

bonifaido:


Gopkg.lock
```diff
@@ -266,12 +341,28 @@
   revision = "ca39e5af3ece67bbcda3d0f4f56a8e24d9f2dad4"
   version = "1.1.3"
 
+[[projects]]
+  name = "github.com/juju/ratelimit"
```
baluchicken: This maybe will cause problem with licencing

cmd/vault-operator/main.go
```diff
@@ -0,0 +1,25 @@
...
+
+func printVersion() {
+	logrus.Infof("Go Version: %s", runtime.Version())
+	logrus.Infof("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH)
+	logrus.Infof("operator-sdk Version: %v", sdkVersion.Version)
+}
+
+func main() {
+	printVersion()
+	sdk.Watch("vault.banzaicloud.com/v1alpha1", "Vault", "default", 5)
```
baluchicken: Do you really need to trigger an event every 5 seconds?

bonifaido: No, actually that is not necessary, what value would be accepted here?

baluchicken: 0 will trigger an event only when something happens with the watched resource

deploy/rbac.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+---
+
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: default-account-vault-operator
+subjects:
+- kind: ServiceAccount
+  name: default
```
baluchicken: I think this should be "vault-operator"

bonifaido: Yes! Sorry

Issue #78, opened 2018-07-02
Support etcd backend


raoofm:
Not a bug, but a feature request.

bank-vaults support the default cloud storage, like s3 for aws. It is a good idea to package backend operators based on config. For example if vault config contains backend as etcd then an etcd-operator and etcd cluster should be deployed along with vault.

bonifaido:
@raoofm please see https://github.com/banzaicloud/bank-vaults/pull/79 for the current implementation we have drafted.

PR #79, opened 2018-07-03
Automatic etcd provisioning

bonifaido:
## HA setup with etcd

Additionally you have to deploy the [etcd-operator](https://github.com/coreos/etcd-operator) to the cluster as well:

```bash
kubectl apply -f operator/deploy/etcd-operator.yaml
```

Now deploy a HA vault which connects to an etcd storage backend:

```bash
kubectl apply -f operator/deploy/cr-etcd-ha.yaml
```

From now on, if you deploy a Vault CRD into the cluster which has an [Etcd Storage Backend](https://www.vaultproject.io/docs/configuration/storage/etcd.html) defined in its configuration the Vault operator will create an EtcdCluster CRD for the Vault instance, and the etcd-operator will orchestrate the etcd cluster. After the etcd cluster is ready the Vault instance can connect to it and will start up. If the Vault CRD is deleted from the cluster the etcd cluster will be GCd as well. You have to make sure you define backup and restore for the etcd cluster to prevent data loss, this part is not handled by the Vault operator, see [this](https://github.com/coreos/etcd-operator#backup-and-restore-an-etcd-cluster) document for more details. 

Please note that TLS is not added for Vault -> etcd communication yet, for that I will create a separate issue: #80 

Fixes: #78


raoofm:
@bonifaido awesome, thanks. Yes #80 would be a key for security

bonifaido:
Great! #80 goind to be implemented in the following days. I will keep you informed.

PR #81, opened 2018-07-05
etcd: full TLS support on client, peer and operator level

bonifaido:
TLS library package integrated

Fixes: #80

### DIFF
--- a/operator/pkg/stub/handler.go
+++ b/operator/pkg/stub/handler.go
@@ -5,13 +5,14 @@ import (
 	"fmt"
 	"net/url"
 	"reflect"
+	"strings"
 
 	"github.com/banzaicloud/bank-vaults/operator/pkg/apis/vault/v1alpha1"
 	"github.com/banzaicloud/bank-vaults/pkg/kv/k8s"
 	"github.com/banzaicloud/bank-vaults/pkg/tls"
 	"github.com/banzaicloud/bank-vaults/pkg/vault"
 	etcdV1beta2 "github.com/coreos/etcd-operator/pkg/apis/etcd/v1beta2"
-	etcdCRClient "github.com/coreos/etcd-operator/pkg/client"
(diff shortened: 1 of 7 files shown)


operator/pkg/stub/handler.go
```diff
@@ -311,11 +417,27 @@ func etcdForVault(v *v1alpha1.Vault) (*etcdV1beta2.EtcdCluster, error) {
...
+	etcdCluster.APIVersion = etcdV1beta2.SchemeGroupVersion.String()
+	etcdCluster.Kind = etcdV1beta2.EtcdClusterResourceKind
+	etcdCluster.Name = etcdName
+	etcdCluster.Namespace = v.Namespace
+	etcdCluster.Labels = labelsForVault(v.Name)
 	etcdCluster.Spec.Size = 3
-	etcdCluster.Spec.Version = "3.2.13"
+	// See https://github.com/coreos/etcd-operator/issues/1962#issuecomment-390539621
+	// for more details why we have to pin to 3.1.15
+	etcdCluster.Spec.Version = "3.1.15"
```
raoofm: This is relevant here
https://github.com/coreos/etcd/issues/9398#issuecomment-390725948

bonifaido: We have ClientAuth in the ExtUsage of the generated Cert already (https://github.com/banzaicloud/bank-vaults/blob/master/pkg/tls/tls.go#L174) but the issue still rises. This issue is probably more related to ReverseDNS setup in the cluster, the exact error message is the same (with different IP): https://github.com/coreos/etcd-operator/issues/1962#issuecomment-390538835

raoofm: @bonifaido should we upgrade to 3.3.8 and see how it goes? Based on this comment https://github.com/coreos/etcd-operator/issues/1731#issuecomment-403701510

PR #86, opened 2018-07-12
Configure the operator namespace over an environment variable

ccojocar:


ccojocar:
I can default to all if no namespace is set.

PR #88, opened 2018-07-12
More verbosity

ccojocar:


Makefile
```diff
@@ -1,5 +1,6 @@
-REGISTRY := banzaicloud
+REGISTRY := cosmincojocar
```
bonifaido: Could you please change the registry?

ccojocar: sorry! I will fix it.

ccojocar:
Fixed!

PR #89, opened 2018-07-12
Extend the Vault CRD with a configuration for credentials (aka service account)

ccojocar:
- This configuration can be used to mount a dedicated service account with the required roles only in the vault,unsealer and configurator containers.
- Add a script which generates the boilerplate code for Vault CRD
- Regenerate the boilerplate code


bonifaido:
Hi @ccojocar, by " boilerplate code for Vault CRD" you mean an API client for Vault objects?

ccojocar:
Right, and deep copy

ccojocar:
I tested right now with GKE, and it works well.

bonifaido:
Great! Could you please have look at the build, it has some issues with `VaultList` objects.

ccojocar:
@bonifaido It seems that some of the code generated for the client was not compiling. The build is not green.

bonifaido:
Now the build is fine, I will have a look at this soon. Thank you!!

PR #123, opened 2018-08-17
[DOC] Linting and expand AWS section

ashemedai:
Took me a fair amount of investigation to get the unsealing working in our setup. Added documentation that will, hopefully, prevent others from wasting as much time and focus on unseal their own Vault when using AWS with KMS/S3.


matyix:
@ashemedai thx for the PR and helping others to navigate smoothly through the unsealing process.

ashemedai:
Will have another one coming later today. Need to be explicit that the vault-root is not needed for unsealing.

Issue #72, opened 2018-06-17
[operator] add the Prometheus metrics as in the helm chart


bonifaido:

dfengshi:
@bonifaido Hello, we were looking to using prometheus to grab vault metrics from our bank-vaults deployment without using the helm charts in banzai-charts.

would adding this support be as straightforward as adding a statsd sidecar similar to this https://github.com/banzaicloud/banzai-charts/blob/master/vault/templates/deployment.yaml#L139-L148 

but converted to go, and deployed in the podspec here? https://github.com/banzaicloud/bank-vaults/blob/master/operator/pkg/stub/handler.go#L357

If so, I can put in a PR. If not, please let me know if there's any critical info I'm missing.

Thanks!

bonifaido:
Hi @dfengshi, yes you got it right, that is what needs to be done. We would be more than happy if you could send a PR for it :) I'm happy to help!

bonifaido:
One small detail that you have to add the following annotations to the pod as well: https://github.com/banzaicloud/banzai-charts/blob/master/vault/values.yaml#L28

dfengshi:
@bonifaido thanks, the annotations are handled in the deployment yaml rather than the handler.go correct?  For example: https://github.com/banzaicloud/bank-vaults/blob/master/operator/deploy/cr.yaml#L11-L13 Or should the handler set the annotations?

bonifaido:
I think it's the best way is to add it in the Handler regardless if they are in the deployment or not (potentially overwriting them).

PR #125, opened 2018-08-21
prometheus metrics support

dfengshi:
attempt at addressing https://github.com/banzaicloud/bank-vaults/issues/72 similar to the banzai-charts helm chart

@bonifaido Hello, as we discussed in the issue #72 , here's a quick attempt at addressing the issue with a sidecar statsd container.

Please feel free to give copious feedback/help as I'm new to the operator pattern and go in general. Apologies if there are any glaring issues, I'll do my best to address. I do have a few questions about the volume mount and the TLS Env.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,9 @@ The keys that will be stored are:
 - `vault-root`, which is the Vault's root token
 - `vault-unseal-N`, where `N` is a number, starting at 0 up to the maximum defined minus 1, e.g. 5 unseal keys will be `vault-unseal-0` up to including `vault-unseal-4`
 
+HashiCorp [recommends to revoke root tokens](https://www.vaultproject.io/docs/concepts/tokens.html#root-tokens) after the initial set up of Vault has been completed.
+To unseal Vault the `vault-root` token is not needed and can be removed from the storage if it was put there via the `--init` call to `bank-vaults`.
+
 ## Examples
 
 Some examples are in `cmd/examples/main.go`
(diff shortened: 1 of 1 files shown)


bonifaido:
The TLS volumes are needed since the Vault Api is exposed through self signed certificate, so the CA is needed.

dfengshi:
@bonifaido thanks, my bad

bonifaido:
Only the formatting is now what needs to be fixed (go fmt)

operator/pkg/stub/handler.go
```diff
@@ -417,6 +417,28 @@ func statefulSetForVault(v *v1alpha1.Vault) (*appsv1.StatefulSet, error) {
...
+							Ports: []v1.ContainerPort{{
+								Name:          "statsd",
+								ContainerPort: 9125,
+								Protocol:      "UDP",
+							}, {
+								Name:          "prometheus",
+								ContainerPort: 9102,
+								Protocol:      "TCP",
+							}},
+							VolumeMounts: withStatsDVolumeMount(v, withCredentialsVolumeMount(v, []v1.VolumeMount{})),
```
bonifaido: `withCredentialsVolumeMount` is not needed for statsd. It is used to access on-cloud resources such as S3 or KMS for example.

operator/pkg/stub/handler.go
```diff
@@ -484,6 +506,14 @@ func withTLSVolumeMount(v *v1alpha1.Vault, volumeMounts []v1.VolumeMount) []v1.V
 	return volumeMounts
 }
 
+func withStatsDVolumeMount(v *v1alpha1.Vault, volumeMounts []v1.VolumeMount) []v1.VolumeMount {
```
bonifaido: The `statsd-mapping` volume has to be created as in: https://github.com/banzaicloud/banzai-charts/blob/master/vault/templates/deployment.yaml#L178

dfengshi:
@bonifaido Hi Nandor, I've added the method `withStatsDVolume()` with the configmap setup as requested, but I'm fuzzy about what's going on in `deploymentForConfigurer()` and `statefulSetForVault()` vs what's going on in the helm chart. Should I be calling `withStatsDVolume()` in both of those methods? Maybe something like this: `withStatsDVolume(v, withTLSVolume(v, withCredentialsVolume(v, []v1.Volume{` ? 

This might be a bad question, because I might be missing some things in terms of understanding.

I've also added you as a collaborator on my fork in case it might be easier for you to just go ahead and dive in to my broken code rather than wait for me to understand what's missing.

Thanks again for the guidance+patience!

bonifaido:
Hi David, I have pushed a fix to your branch, if you don't mind I will rebase your commits into one, so two commits will arrive in the master branch. I have tested the operator and now it is working and exposing the Prometheus metrics correctly, thank you for your contribution! 👍

dfengshi:
@bonifaido Thanks Nandor!

Issue #134, opened 2018-08-23
statsd sidecar container fails to launch


dfengshi:
hello,

The statsd sidecar container fails to launch with error `statsd_exporter: error: unknown short flag '-s', try --help`

Most likely, the problem originates from here: https://github.com/banzaicloud/bank-vaults/blob/master/operator/pkg/stub/handler.go#L439

which was introduced in #125, where it was taken from the args in the helm chart here: https://github.com/banzaicloud/banzai-charts/blob/master/vault/templates/deployment.yaml#L142

it might need to be a cmd instead, or passed in as an arg with a long flag

bonifaido:
This is issue is coming from a breaking change in the states exporter, which was released yesterday, and we have tested it with the version one day earlier, see https://github.com/prometheus/statsd_exporter/commit/82f75d3b9ad053751cc2a5fe869c321b64a47934

I will push a fix for this in the operator and in the helm chart as well

PR #135, opened 2018-08-23
Fix statsd's new dash style

bonifaido:
Fixes: #134 

The issue was introduced in yesterday's breaking change in statsd exporeter :) https://github.com/prometheus/statsd_exporter/commit/82f75d3b9ad053751cc2a5fe869c321b64a47934


dfengshi:
@bonifaido Thanks Nandor! :)

Issue #154, opened 2018-09-12
upgrading version of vault in existing cluster not behaving as expected


eisjcormier:
**Describe the bug**
I have an existing bankvault cluster. I'm running Vault version 0.10.4. If I update the cr-etc-ha.yaml file to have a newer version of Vault (0.11.0) and deploy it, my pods are not updated to the new version. I have to manually delete my pods and when they are recreated they have the new version of Vault. 

**To Reproduce**
1. update https://github.com/banzaicloud/bank-vaults/blob/master/operator/deploy/cr-etcd-ha.yaml#L7 to have a different version of Vault than you already have deployed.
2. kubectl apply -f cr_etcd_ha.yaml
3. kubectl -n <vault namespace> describe pod vault-0  ( or vault-1)
4. See Image is not the version of Vault specified in the update cr-etcd-ha.yaml file:
~~~
Containers:
  vault:
    Container ID:  docker://xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
    Image:         vault:0.10.4
~~~

running:  kubectl -n <vault namespace> delete pod vault-0 will cause the pod to get recreated and it will have the correct/expect Image version

**Expected behavior**
 If I update the cr-etcd-ha.yaml file with a new version of Vault and deploy it, my pods perform a rolling update to the new version.

bonifaido:
Hi @eisjcormier, which Kubernetes version are you using? I can't reproduce this on my version 1.10 cluster.

bonifaido:
this is how my `StatefulSet` looks like after the update from 0.11.0 to 0.11.1 Vault:
```
Name:               vault
Namespace:          default
CreationTimestamp:  Wed, 12 Sep 2018 20:05:59 +0200
Selector:           app=vault,vault_cr=vault
Labels:             <none>
Annotations:        <none>
Replicas:           1 desired | 1 total
Update Strategy:    RollingUpdate
Pods Status:        1 Running / 0 Waiting / 0 Succeeded / 0 Failed
Pod Template:
  Labels:       app=vault
                vault_cr=vault
  Annotations:  prometheus.io/path=/metrics
                prometheus.io/port=9102
                prometheus.io/scrape=true
  Containers:
   vault:
    Image:       vault:0.11.1
    Ports:       8200/TCP, 8201/TCP
    Host Ports:  0/TCP, 0/TCP
```

eisjcormier:
Hi. Thanks for looking into my issue. 

[...]

PR #157, opened 2018-09-13
Support StatefuleSet upgrade with workaround

bonifaido:
This feature gives us the option to workaround current **StatefulSet limitations** around updates
See: https://github.com/kubernetes/kubernetes/issues/67250
By default it is set to **false**.

Fixes: #154

### DIFF
--- a/cmd/bank-vaults/unseal.go
+++ b/cmd/bank-vaults/unseal.go
@@ -13,11 +13,15 @@ import (
 const cfgUnsealPeriod = "unseal-period"
 const cfgInit = "init"
 const cfgOnce = "once"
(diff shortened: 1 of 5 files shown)


matyix:
Although the Kubernetes [bug report](kubernetes/kubernetes#67250) attached referencing `broken` stage this actually applies to `not ready` pods as well. The rolling upgrade does upgrade all the pods in any other state than the `ready` one - and unluckily it leaves that out from the upgrade. The workaround is to let rolling upgrades all the `not ready` pods in the StatefulSet then switch to an upgraded one (ready moves to not ready) - the controller than kicks on to upgrade the old ready one as well. 

Hope this was clear enough ...

PR #165, opened 2018-09-21
etcd size for quorum fault tolerance

dfengshi:
If etcd is chosen as a back end, the size is hard coded to 3
https://github.com/banzaicloud/bank-vaults/blob/master/operator/pkg/stub/handler.go#L608

This might cause issues if quorum is lost https://github.com/coreos/etcd-operator/issues/1972, having a size of 3 only tolerates the failure of 1 etcd pod.

Added support in the spec for etcd size


bonifaido:
Good built-in checks!

PR #172, opened 2018-10-03
Added ability to enable plugins via bank-vaults

alexandrst88:
Signed-off-by: Oleksandr Stepanov <alexandrst88@gmail.com>

We very like this tool, so we were trying to use it, but faced with the issue, there is not possible to enable plugins via it. 

This PR provides an ability to enable Vault Plugins: https://www.vaultproject.io/docs/internals/plugins.html

Requirements:
- **Vault Plugin Directory must be set** otherwise, vault will throw an error.

Question: 
- We have some concern about using `getOrDefault` function? We are thinking about creating a new function like `getValue` or `getStringValue`, because, fields: command, plugin_name and sha256 are obligatory. 
- Also, should we add an output of existing plugins as it did for `configureSecretEngines` method?

We did tests with this plugin: https://github.com/immutability-io/vault-ethereum. 

If you like this PR, we make additional commits with examples and updated docs.

### DIFF
--- a/.licensei.toml
+++ b/.licensei.toml
@@ -16,4 +16,7 @@ ignored = [
   "github.com/davecgh/go-spew", # ISC license
   "github.com/howeyc/gopass", # ISC license
   "github.com/oracle/oci-go-sdk", # UPL-1.0
+
+  "github.com/Masterminds/sprig",
+  "github.com/magiconair/properties"
 ]
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -79,7 +79,7 @@ required = [
 [[constraint]]
   name = "k8s.io/api"
   branch = "release-1.10"
-  
+
 [[constraint]]
   name = "k8s.io/apimachinery"
   branch = "release-1.10"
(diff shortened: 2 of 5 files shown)


hatemosphere:
Follow-up comments:
1. Non-default Vault plugins use RPC to communicate with Vault itself, so readOnlyRootFileSystem security option in Kubernetes should be turned off to allow socket writes to /tmp folder
2. Non-default Vault plugin binaries should be built-in to Vault Docker image.

We can add all the important information to README if you guys think it's actually a proper design to configure non-default plugins with bank-vaults

bonifaido:
Hi @alexandrst88 and @hatemosphere !

This feature is a very nice contribution, good idea, thank you!

A few remarks and answering your questions, observations:

- Could you please add the two undetected licensed projects to the ignored block of https://github.com/banzaicloud/bank-vaults/blob/master/.licensei.toml#L9 ? This breaks the build in which we have a license check, both of those are MIT, so we can ignore them:

```
"github.com/Masterminds/sprig",
"github.com/magiconair/properties"
```

- Yes, you're right, instead of `getOrDefault` I suggest to create a `getOrError` or whichever you like to get a value or return an error in case of its absence for the obligatory values.

- Yes, please return any outputs as you like in` configurePluings()`.

- README extensions about this feature would be great.

I think this will need to be integrated to the operator as well some day, but that one should be a seperate PR for sure.

alexandrst88:
@bonifaido please, review, additional changes to examples and `pkg/vault/operator_client.go`, thanks :)

bonifaido:
Looks very much as I expected! Thank you! @alexandrst88 @hatemosphere

bonifaido:
A release has been cut with including your submission: https://github.com/banzaicloud/bank-vaults/releases/tag/0.3.1 it takes some time for the images to arrive to docker hub.

Issue #163, opened 2018-09-21
Fluentd logging sidecar


gowthamsubbu:
@bonifaido We are planning to introduce a fluentd side car in the vault deployment to manage Vault logs (audit, sys) and we can create a PR to get this feature added. Please let us know if you have any questions or concern.

bonifaido:
Hi @gowthamsubbu, all PRs are welcome, this sounds like a nice addition. The only requirement from our side is that it should be optional, so users can turn it off if they don't want to use it.

bonifaido:
Fixed by #174.

PR #174, opened 2018-10-03
operator: add generic fluentd logging sidecar

dfengshi:
Hello,

Here are some additions to the operator to allow a generic fluentd sidecar container. This was brought up in this issue https://github.com/banzaicloud/bank-vaults/issues/163 . We plan on using this sidecar to handle audit logging.

Everything related to this is purely optional and disabled by default.

Please let me know if you have any questions.

### DIFF
--- a/operator/pkg/apis/vault/v1alpha1/types.go
+++ b/operator/pkg/apis/vault/v1alpha1/types.go
@@ -53,6 +53,7 @@ type VaultSpec struct {
 	Image             string                 `json:"image"`
 	BankVaultsImage   string                 `json:"bankVaultsImage"`
 	StatsDImage       string                 `json:"statsdImage"`
+	FluentDImage      string                 `json:"fluentdImage"`
 	Annotations       map[string]string      `json:"annotations"`
 	Config            map[string]interface{} `json:"config"`
 	ExternalConfig    map[string]interface{} `json:"externalConfig"`
@@ -171,6 +172,22 @@ func (spec *VaultSpec) GetAnnotations() map[string]string {
 	spec.Annotations["prometheus.io/path"] = "/metrics"
 	spec.Annotations["prometheus.io/port"] = "9102"
 	return spec.Annotations
+}
+
+// GetFluentDImage returns the StatsD image to use
+func (spec *VaultSpec) GetFluentDImage() string {
+	if spec.FluentDImage == "" {
+		return "fluent/fluentd:v0.12-onbuild"
+	}
+	return spec.FluentDImage
+}
+
+// GetFluentDEnable returns true if fluentd sidecar is to be deployed
+func (spec *VaultSpec) GetFluentDEnable() bool {
+	if spec.FluentDImage == "" {
+		return false
+	}
+	return true
 }
 
 // ConfigJSON returns the Config field as a JSON string
--- a/operator/pkg/apis/vault/v1alpha1/types.go
+++ b/operator/pkg/apis/vault/v1alpha1/types.go
@@ -183,7 +183,7 @@ func (spec *VaultSpec) GetFluentDImage() string {
 	return spec.FluentDImage
 }
 
-// GetFluentDEnable returns true if fluentd sidecar is to be deployed
+// IsFluentDEnabled returns true if fluentd sidecar is to be deployed
 func (spec *VaultSpec) IsFluentDEnabled() bool {
 	// zero value for bool is false
 	return spec.FluentDEnabled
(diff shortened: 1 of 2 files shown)


operator/pkg/apis/vault/v1alpha1/types.go
```diff
@@ -172,6 +173,22 @@ func (spec *VaultSpec) GetAnnotations() map[string]string {
 	return spec.Annotations
 }
 
+// GetFluentDImage returns the StatsD image to use
```
sagikazarmark: StatsD should be FluentD

operator/pkg/apis/vault/v1alpha1/types.go
```diff
@@ -172,6 +173,22 @@ func (spec *VaultSpec) GetAnnotations() map[string]string {
...
+func (spec *VaultSpec) GetFluentDImage() string {
+	if spec.FluentDImage == "" {
+		return "fluent/fluentd:v0.12-onbuild"
```
sagikazarmark: 0.12 is quite an old version, the latest is 1.2.x if I recall correctly. I'm inclined to say we should use the latest-greatest version.

operator/pkg/apis/vault/v1alpha1/types.go
```diff
@@ -172,6 +173,22 @@ func (spec *VaultSpec) GetAnnotations() map[string]string {
...
+	}
+	return spec.FluentDImage
+}
+
+// GetFluentDEnable returns true if fluentd sidecar is to be deployed
+func (spec *VaultSpec) GetFluentDEnable() bool {
```
sagikazarmark: Something like `IsFluentDEnabled` would be better IMHO.

operator/pkg/apis/vault/v1alpha1/types.go
```diff
@@ -172,6 +173,22 @@ func (spec *VaultSpec) GetAnnotations() map[string]string {
...
+	if spec.FluentDImage == "" {
+		return false
+	}
+	return true
```
sagikazarmark: Maybe just return `spec.FluentDImage != ""`? But maybe even better would be a separate configuration option for enabling the sidecar?

operator/pkg/apis/vault/v1alpha1/types.go
dfengshi: @sagikazarmark  
Yeah, definitely! I was following how `GetTLSDisable()` was named, but I would definitely be willing to rename to `IsFluentDEnabled`

dfengshi:
@sagikazarmark Hi Márk! Thanks for the feedback, changes made

bonifaido:
Hi @dfengshi! Changes LGTM after the fixes!
How do you enable audit logging for Vault for this setup? (Audit is off by default). It would be nice if we could use the configurer for that, but I'm not sure if that is possible without code changes.

bonifaido:
Fixes #163

bonifaido:
Also @dfengshi, could you please rebase on latest master, to get the build fixed? (the license issue).

dfengshi:
Hey @bonifaido! Currently enabling audit logging via ansible. Adding this functionality as part of the configurer definitely makes sense to me.

bonifaido:
@dfengshi that would be really great, I think let's make that feature part of another pull request. What do you think?

dfengshi:
@bonifaido Agreed

operator/pkg/stub/handler.go
```diff
@@ -550,6 +554,51 @@ func withTLSVolumeMount(v *v1alpha1.Vault, volumeMounts []v1.VolumeMount) []v1.V
...
+
+func withAuditLogContainer(v *v1alpha1.Vault, owner string, containers []v1.Container) []v1.Container {
+	if v.Spec.IsFluentDEnabled() {
+		containers = append(containers, v1.Container{
+			Image:           v.Spec.GetFluentDImage(),
+			ImagePullPolicy: v1.PullAlways,
+			Name:            "auditlog-exporter",
+			Env: withTLSEnv(v, true, withCredentialsEnv(v, []v1.EnvVar{
+				{
+					Name:  k8s.EnvK8SOwnerReference,
```
bonifaido: Why is the `EnvK8SOwnerReference` needed for this container?

dfengshi: Having the owner reference as an env variable might be useful to tie logs to the owning object. But I haven't worked on anything to do that yet, and would be okay with removing it for now. What do you think?

bonifaido: If you think it gives more context to the logs (I think it can), then fine by me!

sagikazarmark:
Thanks @dfengshi

PR #177, opened 2018-10-05
Configurer needs to check if Vault is active before configuring

bonifaido:
This solves a race condition when Vault hasn't mounted the auth backends/secret engines before configuring but is mounting them during configuration.

### DIFF
--- a/pkg/vault/operator_client.go
+++ b/pkg/vault/operator_client.go
@@ -15,9 +15,11 @@
 package vault
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"runtime"
 	"strings"
@@ -62,6 +64,7 @@ var _ Vault = &vault{}
 type Vault interface {
 	Init() error
 	Sealed() (bool, error)
+	Active() (bool, error)
 	Unseal() error
 	Leader() (bool, error)
 	Configure() error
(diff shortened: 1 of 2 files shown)


pkg/vault/operator_client.go
```diff
@@ -90,6 +93,20 @@ func (v *vault) Sealed() (bool, error) {
...
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+	resp, err := v.cl.RawRequestWithContext(ctx, req)
+	if err == nil {
+		defer resp.Body.Close()
+	}
+	if resp.StatusCode == http.StatusOK {
+		return true, nil
+	}
+	return false, fmt.Errorf("error checking status: %s", err.Error())
```
Ecsy: Reaching this line err could be nil. Calling Error() on nil is not advisable, causes nil pointer dereference.
I would rather replace the above block from line 101 to:
```go
	if err != nil {
		return false, fmt.Errorf("error checking status: %s", err.Error())
	}
	defer resp.Body.Close()
	if resp.StatusCode == http.StatusOK {
		return true, nil
	}
	return false, fmt.Errorf("error unexpected status code: %d", resp.StatusCode)
```

bonifaido: Without doubt you're right! Updated accordingly, thank you!

PR #184, opened 2018-10-12
Add ability to use existing etcd cluster

alantang888:
I have a special case. Which have to use existing etcd as vault's backend storage. 
So use EtcdSize < 0 to let operator not create new etcd cluster. And not override any setting under etcd storage section.


bonifaido:
Thank you @alantang888, nice contribution!

bonifaido:
The golint issue is fixed now on master.

alantang888:
Rebased.

PR #185, opened 2018-10-15
Change the type of vault service to ClusterIP in order to be able to expose it

ccojocar:


bonifaido:
Thanks. I think this is the way to go, probably we need to implement some kind of exposing reconfigurability to the operator as well.

bonifaido:
So @ccojocar can we make this configurable? So existing users using NodePort won't lost their functionality?

bonifaido:
That sounds feasible.

PR #188, opened 2018-10-18
Add Security Context to Vault CR

### LINKED ISSUE #187: Parse SecurityContext to Vault Statefulset
Currently the `default` Service Account in the target namespace must have the `anyuid` SCC, as well as have the `IPC_LOCK` capability, otherwise the vault container continuously crashes due to permission denied when it tries to `chown` the `/vault` directory so that the `/vault` directory user uid matches the running user uid.

Allowing pods to run as root opens the door to an exploit that allows processes to escape the container engine.

It would be useful to have the ability to add `securityContext: {}` to the Vault Statefulset, specifically the `runAsUser` flag.
```yaml
apiVersion: "vault.banzaicloud.com/v1alpha1"
kind: "Vault"
metadata:
  name: "vault"
spec:
  size: 1
  image: vault:0.11.0
  
  securityContext:
    runAsUser: 100
```
This will allow you to give a `default` Service Account the `nonroot` SecurityContextConstraint, slightly modified to allow for the `IPC_LOCK` Capability.
This prevents pods from being run with the root user, which helps to prevent a privilege escalation exploit.

**Additional context**
> By default, OKD runs containers using an arbitrarily assigned user ID. This provides additional security against processes escaping the container due to a container engine vulnerability and thereby achieving escalated permissions on the host node.

[OKD Docs](https://docs.okd.io/latest/creating_images/guidelines.html#openshift-specific-guidelines)

RBlaine95:
This fixes #187 

**Edit:**
This allows users to customize the `securityContext` of the Vault Statefulset.

```yaml
apiVersion: "vault.banzaicloud.com/v1alpha1"
kind: "Vault"
metadata:
  name: "vault"
spec:
  size: 1
  image: vault:0.11.0
  
  securityContext:
    runAsUser: 100
    fsGroup: 100
```

### DIFF
--- a/operator/pkg/apis/vault/v1alpha1/types.go
+++ b/operator/pkg/apis/vault/v1alpha1/types.go
@@ -16,6 +16,7 @@ package v1alpha1
 
 import (
 	"encoding/json"
+	"k8s.io/api/core/v1"
 	"reflect"
 
 	"github.com/spf13/cast"
@@ -60,6 +61,7 @@ type VaultSpec struct {
 	ExternalConfig    map[string]interface{} `json:"externalConfig"`
 	UnsealConfig      UnsealConfig           `json:"unsealConfig"`
 	CredentialsConfig CredentialsConfig      `json:"credentialsConfig"`
+	SecurityContext   v1.PodSecurityContext  `json:"securityContext,omitempty"`
 	// This option gives us the option to workaround current StatefulSet limitations around updates
 	// See: https://github.com/kubernetes/kubernetes/issues/67250
 	// TODO: Should be removed once the ParallelPodManagement policy supports the broken update.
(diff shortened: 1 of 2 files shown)


bonifaido:
Thank you very much @RBlaine95 !

matyix:
Thanks @RBlaine95 for opening #187 and submitting the PR. LGTM

sepulworld:
Hmm, does this work as intended? When trying to switch to 

```
  securityContext:
    runAsUser: 100
    fsGroup: 1000
```

I see... 

```
chown: /vault/config: Operation not permitted
Could not chown /vault/config (may not have appropriate permissions)
chown: /vault/config: Operation not permitted
unable to set CAP_SETFCAP effective capability: Operation not permitted
```

rblaine95:
@sepulworld, last I checked, yes, it does work as expected as it correctly sets the [Security Context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) configuration according to the [Kubernetes pod spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#podsecuritycontext-v1-core)

PR #196, opened 2018-10-30
Add Pod anti-affinity support

alantang888:
If you want setup pod anti-affinity. You can set `podAntiAffinity` vault with a topologyKey value. 
For example, you can use `failure-domain.beta.kubernetes.io/zone` to force K8S deploy vault on multi AZ.

Detail can reference: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity
And: https://kubernetes.io/docs/setup/multiple-zones/#nodes-are-labeled


alantang888:
checking on golint error =.="

PR #199, opened 2018-11-07
Add etcdAnnotations field to Vault CR to annotate etcdcluster

rblaine95:
This allows users to add `spec.etcdAnnotations` to the Vault Custom Resource object which is used to annotate the `etcdcluster` Custom Resource accordingly.

This is specifically for the `etcd.database.coreos.com/scope=clusterwide` annotation, however CoreOS allows for arbitrary annotations to the object so I figured I'd carry on with that and allow full passthrough of annotations.

Example usage:
```yaml
apiVersion: vault.banzaicloud.com/v1alpha1
kind: Vault
labels:
  version: latest
  group: security
  app: vault
metadata:
  name: vault
spec:
  size: 3
  image: vault:0.11.0
  securityContext:
    runAsUser: 100
  supportUpgrade: true
  etcdAnnotations:
    etcd.database.coreos.com/scope: clusterwide
  unsealConfig:
    kubernetes:
      secretNamespace: default
...
```

This means that you can have a Cluster Wide ETCD Operator in one namespace, and then create an `etcdcluster` in a different namespace and have it automatically annotated.

### DIFF
--- a/operator/deploy/cr-etcd-ha.yaml
+++ b/operator/deploy/cr-etcd-ha.yaml
@@ -12,6 +12,12 @@ spec:
   # By default it is false.
   supportUpgrade: true
 
+  # This option allows you to annotate the ETCD Cluster that Vault Operator creates.
+  # It's specifically to annotate the ETCD Cluster as 'clusterwide' for a cluster wide
+  # ETCD Operator, however it can be used to set any arbitrary annotations on the ETCD Cluster.
+  # etcdAnnotations:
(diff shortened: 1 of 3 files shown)


bonifaido:
Looks good @rblaine95, could you please add the example annotation to the [etcd example](https://github.com/banzaicloud/bank-vaults/blob/master/operator/deploy/cr-etcd-ha.yaml) just as a comment, that this is possible?

rblaine95:
@bonifaido, just added the example as a comment

PR #202, opened 2018-11-14
Set secrets to bank-vaults's environment variables

alantang888:
Add config for set secrets to bank-vaults's environment variables. Use case can reference readme example's `my-mysql` username and password.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,18 @@ auth:
       token_max_ttl: 30m
       secret_id_num_uses: 40
 
+# Pass secret to environment variables. Please reference below `my-mysql` part for usage.
+# `env` is environment variable name
+# `secretName` is secret's name.
+# `secretKey` is secret's data key.
+secretEnvsConfig:
+  - env: ROOT_USERNAME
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -205,6 +205,18 @@ auth:
       token_max_ttl: 30m
       secret_id_num_uses: 40
 
+# Pass secret to environment variables. Please reference below `my-mysql` part for usage.
```
bonifaido: I think this should go to one of the deployment examples in the operator package.

On the other hand, this could be more general, like the Kubernetes Pod env config, so it's possible to pass in secretRefs and env string literals as well like:
```yaml
env:
    - name: REGION
      value: us-east-1
    - name: SECRET_USERNAME
      valueFrom:
        secretKeyRef:
          name: test-secret
          key: username
```
For parsing we can just reuse the `v1.EnvVar` Kubernetes struct.

alantang888: ok, will do it.

alantang888: Changed.

Issue #222, opened 2018-12-06
Running Vault 1.0.0 breaks Configurer


rblaine95:
**Describe the bug**
When using Vault v1.0.0, the configurer has problems configuring Vault.

**To Reproduce**
1. Deploy Vault Operator
2. Deploy a Vault cluster with a Vault image that's v1.0.0 (for example: `quay.io/zenlab/vault:1.0.0-1`)
3. Get the logs for the configurer
```bash
$ oc logs -f vault-configurer-7d6459fbc9-ht9g4 

time="2018-12-06T09:07:05Z" level=info msg="New config file change \"Initial\": CREATE"
time="2018-12-06T09:07:05Z" level=info msg="checking if vault is sealed..."
time="2018-12-06T09:08:09Z" level=info msg="checking if vault is active..."
time="2018-12-06T09:08:09Z" level=info msg="vault is unsealed and active, configuring..."
time="2018-12-06T09:08:09Z" level=error msg="error configuring vault: error configuring plugins for vault: Failed to retrieve list of plugins: Error making API request.\n\nURL: LIST https://vault.myproject:8200/v1/sys/plugins/catalog\nCode: 405. Errors:\n\n* 1 error occurred:\n\t* unsupported operation\n\n"
```

**Expected behavior**
Configurer should be able to get the list of plugins and continue with the configuration of Vault.

**Desktop:**
 - OS: Arch Linux
 - K8s: OKD v3.11.0

**Additional context**
It looks like in previous versions of Vault (pre-1.0.0) the `LIST` operation on the `/v1/sys/plugins/catalog` was a valid operation, but it looks like now it needs to be changed to `GET`.

This only happens with v1.0.0 of Vault.

bonifaido:
Hi @rblaine95, yes this is handled in the new version of the client in a backward compatible way, so I will update the client dependencies and check if it works!

https://github.com/hashicorp/vault/blob/v1.0.0/api/sys_plugins.go#L36

bonifaido:
I have checked and the backward compatibility mechanism doesn't seem to work for me with 0.11.5. I have submitted a PR to the upstream Vault project to resolve this (I hope I fixed it correctly 🤞)

https://github.com/hashicorp/vault/pull/5913

My only concert is that if someone happens to use bank-vaults with 0.11.5 and don't want to upgrade Vault versions in the near future, he/she will be forced to use the pre-Vault 1.0.0 supporting version of bank-vaults.

PR #224, opened 2018-12-06
Update to Vault 1.0.0

bonifaido:
Updates the API client used to talk with Vault 1.0.0.

- [x] Vault 0.11.x still works. (not sure, testing)

Fixes: #222 Depends on: https://github.com/hashicorp/vault/pull/5913

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -408,7 +408,7 @@
   revision = "ef8a98b0bbce4a65b5aa4c368430a80ddc533168"
 
 [[projects]]
-  digest = "1:dbef1dda9aa352f3e62bcaa87be4e33d5e1736416aec983a7868d219249e919d"
+  digest = "1:855a0cb08277efb9b18fee68f6f94259775df36350276257df8c96e9d51dbaaf"
   name = "github.com/hashicorp/vault"
   packages = [
     "api",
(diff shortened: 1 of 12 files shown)


ccojocar:
@bonifaido Any plans to phase out the `configurer` in favour of newly introduced `Cloud Auto Unseal`? 

https://github.com/hashicorp/vault/blob/master/vault/seal

Thanks

bonifaido:
Hi @ccojocar , autoseal is just unsealing, so I guess you mean not the configurer but the `init/unseal` subcommand of bank-vaults. Since the configurer is widely used part and it relies currently on the keys put to persistent places in the format as `bank-vaults init` does we don't plan to retire the `init/unseal` part soon, also we support  more options (and plan to support even more) for init/unseal, a, like Kubernetes secrets for testing purposes, where you don't have a cloud provider underneath. What I can think of right now is that the configurer could configure a Vault instance that wasn't inited/unseald via bank-vaults, but for that first I have to check how the autounseal actually works.

ccojocar:
Thanks @bonifaido for clarifications. Yeah, I mean the init container. It seems that the vault has a similar approach if you check the implementation, but in addition they wrap the key shards.  https://github.com/hashicorp/vault/blob/1a7cd116a77d12af5df6a7cdd61c358a3324fcbb/vault/seal_autoseal.go#L76. 

Looking at the implementation, it seems to me that all key shards are wrapped, then the wrap key is encrypted with the KMS service, and finally the entire encryption material is stored together in the physical storage. 

Wrapping the keys before storing them, it might have a slightly better security since the wrap key does not have the same level of access like the master key shards.

bonifaido:
Hi @ccojocar, I have created a separate issue for this concern of using auto-unseal: https://github.com/banzaicloud/bank-vaults/issues/228

For the original API compatibility issue, I have pushed a fix which uses a Vault client from our branch temporarily, until the upstream issue gets resolved.

PR #225, opened 2018-12-07
Mutating webhook for injecting secrets into Kubernetes containers

bonifaido:


cmd/vault-secrets-webhook/main.go
```diff
@@ -0,0 +1,322 @@
...
+
+	annotations := obj.GetAnnotations()
+	if _, ok := annotations["vault.banzaicloud.com/mutated"]; ok {
+		return false, nil
+	}
+	if annotations == nil {
+		annotations = map[string]string{}
+		obj.SetAnnotations(annotations)
+	}
+	annotations["vault.banzaicloud.com/mutated"] = "true"
```
stoader: should this be rather `vault.banzaicloud.io` or coming from a config?

bonifaido: I think the annotations keys should be fix. We were using vault.banzaicloud.com for the operator, but maybe we should support both!

matyix: In general, we use .io but there might be other changes as well to be in line with other annotations we use (as an example see the pattern used in the [HPA operator](https://github.com/banzaicloud/hpa-operator))

bonifaido: Review comments resolved around annotations.

PR #253, opened 2018-12-21
Remove unnecessary data property.

andersjanmyr:
`secret.Data` already parsed the data property.

### DIFF
--- a/cmd/vault-env/main.go
+++ b/cmd/vault-env/main.go
@@ -70,7 +70,7 @@ func main() {
 				fmt.Fprintf(os.Stderr, "path not found: %s\n", path)
 				os.Exit(1)
 			} else {
-				data := cast.ToStringMap(secret.Data["data"])
+				data := cast.ToStringMap(secret.Data)
 				if value, ok := data[key]; ok {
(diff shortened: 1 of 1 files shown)


bonifaido:
Hi @andersjanmyr, vault-env currently only supports Vault KV Version 2, and in that case, the `data` is nested inside another `data` property as you can see in: https://www.vaultproject.io/api/secret/kv/kv-v2.html#sample-response-1 so we can't remove this attribute access. Do you use KV Version 2 in Vault?

andersjanmyr:
Hi @bonifaido. You are right, I was using Vault KV Version 1. It may be possible to support both versions with a simple if statement checking if `data.data` is available and if not fall back to just `data`.

bonifaido:
Yes that is right, that was our idea as well when we were planning v1 compatibility, if you would like to submit a change for that that is more than welcome!

andersjanmyr:
@bonifaido I made the changes and rebase on top of the new master.

bonifaido:
Okay the issue is with the branch name, the Docker tag is generates based on that, so it can't contain any slashes, I will fix it. For now we can just ignore that failure. Thanks for PR! :-)

PR #256, opened 2019-01-02
Make - extracted "generic" targets

lpuskas:
Generic make targets - used in most of the Banzai Cloud projects have been extracted into a separate file that can be included in project specific Make

### DIFF
--- a/main-targets.mk
+++ b/main-targets.mk
@@ -85,7 +85,6 @@ license-cache: bin/licensei ## Generate license cache
 
 .PHONY: test
 test:
-	set -o pipefail
 	go list ./... | xargs -n1 go test ${GOARGS} -v -parallel 1 2>&1 | tee test.txt
 
 bin/go-junit-report:
--- a/main-targets.mk
+++ b/main-targets.mk
@@ -85,7 +85,8 @@ license-cache: bin/licensei ## Generate license cache
 
 .PHONY: test
 test:
-	set -oe pipefail; go list ./... | xargs -n1 go test ${GOARGS} -v -parallel 1 2>&1 | tee test.txt
+	set -o pipefail
+	go list ./... | xargs -n1 go test ${GOARGS} -v -parallel 1 2>&1 | tee test.txt
 
 bin/go-junit-report:
 	@mkdir -p bin
--- a/main-targets.mk
+++ b/main-targets.mk
@@ -85,7 +85,7 @@ license-cache: bin/licensei ## Generate license cache
 
 .PHONY: test
 test:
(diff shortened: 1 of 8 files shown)


main-targets.mk
```diff
@@ -0,0 +1,121 @@
+# A Self-Documenting Makefile: http://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
+
+
+.PHONY: clean
+clean: ## Clean the working area and the project
+	rm -rf bin/ ${BUILD_DIR}/ vendor/
+	rm -rf ${BINARY_NAME}
```
sagikazarmark: Built binaries should be in the build dir

main-targets.mk
```diff
@@ -0,0 +1,121 @@
+# A Self-Documenting Makefile: http://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
+
+
+.PHONY: clean
+clean: ## Clean the working area and the project
+	rm -rf bin/ ${BUILD_DIR}/ vendor/
```
sagikazarmark: Modern go app now has separate clear and clean targets

main-targets.mk
```diff
@@ -0,0 +1,121 @@
...
+	@mkdir -p bin
+	curl https://raw.githubusercontent.com/golang/dep/master/install.sh | INSTALL_DIRECTORY=bin DEP_RELEASE_TAG=v${DEP_VERSION} sh
+	@mv bin/dep $@
+
+.PHONY: vendor
+vendor: bin/dep ## Install dependencies
+	bin/dep ensure -v -vendor-only
+
+.PHONY: build
+build: GOARGS += -tags "${GOTAGS}" -ldflags "${LDFLAGS}"
```
sagikazarmark: This makes reading a target a bit harder, so unless there is need to override something, adding options to the command is preferred.

main-targets.mk
```diff
@@ -0,0 +1,121 @@
...
+	bin/licensei check
+	./scripts/check-header.sh
+
+.PHONY: license-cache
+license-cache: bin/licensei ## Generate license cache
+	bin/licensei cache
+
+.PHONY: test
+test:
+	go list ./... | xargs -n1 go test ${GOARGS} -v -parallel 1 2>&1 | tee test.txt
```
sagikazarmark: See the modern go app for a better test setup. Don't use parallel 1 unless necessary, go-junit-report can be replaced with gotestsum. (modern go app has a circleci config)

main-targets.mk
lpuskas: thx, added the modification

PR #267, opened 2019-01-10
Allow to control vault's log level with env

szymonpk:
According to vault server -help, -log-level can be set with VAULT_LOG_LEVEL. There is no need to set it explicitly.

Also, fix link to Examples in README.md.


baluchicken:
thx @szymonpk

PR #270, opened 2019-01-10
/operator/deploy/cr-aws.yaml: update config aws/s3

### LINKED ISSUE #269: Error in exmaple manifests; "config.storage.aws" should be "config.storage.s3", causes an error during startup
**Describe the bug**
The manifests in the ["operator/build"](https://github.com/banzaicloud/bank-vaults/tree/master/operator/deploy) folder that refer to AWS have an error in them, the correct storage type for AWS should be `s3`. It's currently: `config.storage.aws` ([link](https://github.com/banzaicloud/bank-vaults/blob/master/operator/deploy/cr-aws.yaml#L32)).

And when deployed as is, the vault container won't start and will print the following: 
> Unknown storage type aws

After changing it to: `config.storage.s3`, It starts up normally. I'll be happy to open a PR and fix it myself if you guys are open to contributers outside of Banzai Cloud.

raoofm:
fixes #269

### DIFF
--- a/operator/deploy/cr-aws.yaml
+++ b/operator/deploy/cr-aws.yaml
@@ -29,7 +29,7 @@ spec:
   # See https://www.vaultproject.io/docs/configuration/ for more information.
   config:
     storage:
-      aws:
+      s3:
         bucket: "my-vault"
     listener:
       tcp:


baluchicken:
thanks @raoofm

PR #274, opened 2019-01-14
Use `path` when configuring Kubernetes auth roles

paigr:
Currently, roles are being written to the default Kubernetes auth path, `auth/kubernetes`. This change will have the method `configureKubernetesRoles` use the value of `path` instead, as done in `kubernetesAuthConfig`.

### DIFF
--- a/pkg/vault/operator_client.go
+++ b/pkg/vault/operator_client.go
@@ -400,7 +400,7 @@ func (v *vault) Configure() error {
 				return fmt.Errorf("error configuring kubernetes auth for vault: %s", err.Error())
 			}
 			roles := authMethod["roles"].([]interface{})
-			err = v.configureKubernetesRoles(roles)
+			err = v.configureKubernetesRoles(path, roles)
 			if err != nil {
 				return fmt.Errorf("error configuring kubernetes auth roles for vault: %s", err.Error())
 			}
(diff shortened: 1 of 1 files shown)


pkg/vault/operator_client.go
```diff
@@ -579,14 +579,14 @@ func (v *vault) configurePolicies() error {
...
-func (v *vault) configureKubernetesRoles(roles []interface{}) error {
+func (v *vault) configureKubernetesRoles(path string, roles []interface{}) error {
 	for _, roleInterface := range roles {
 		role, err := cast.ToStringMapE(roleInterface)
 		if err != nil {
 			return fmt.Errorf("error converting role for kubernetes: %s", err.Error())
 		}
 
-		_, err = v.cl.Logical().Write(fmt.Sprint("auth/kubernetes/role/", role["name"]), role)
+		_, err = v.cl.Logical().Write(fmt.Sprint("auth/%s/role/%s", path, role["name"]), role)
```
bonifaido: Please change `Sprint` to `Sprintf` in this case.

paigr: Ah, of course, my mistake.

paigr:
That was quick! Thanks everyone, I'm pleased to be able to contribute to such a great project.

PR #278, opened 2019-01-16
Add support to config secret engine, which nosub-path after config

alantang888:
Like "ad", "alicloud", "azure", "gcp", "gcpkms", "kv" secret engine

### DIFF
--- a/pkg/vault/operator_client.go
+++ b/pkg/vault/operator_client.go
@@ -38,7 +38,7 @@ import (
 const DefaultConfigFile = "vault-config.yml"
 
 // Secret engine config no need name
-var secretEngineConfigNoNeedName = [...]string{"ad", "alicloud", "azure", "gcp", "gcpkms", "kv"}
+var secretEngineConfigNoNeedName = map[string]bool{"ad": true, "alicloud": true, "azure": true, "gcp": true, "gcpkms": true, "kv": true}
 
 // Config holds the configuration of the Vault initialization
 type Config struct {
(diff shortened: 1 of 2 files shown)


pkg/vault/operator_client.go
```diff
@@ -37,6 +37,9 @@ import (
 // DefaultConfigFile is the name of the default config file
 const DefaultConfigFile = "vault-config.yml"
 
+// Secret engine config no need name
+var secretEngineConfigNoNeedName = [...]string{"ad", "alicloud", "azure", "gcp", "gcpkms", "kv"}
```
bonifaido: Could we use a `map[string]bool` here instead of the slice?

alantang888: Then use `_, ok := secretEngineConfigNoNeedName[secretEngine]` to check? Value just a dummy value?

bonifaido: Yes, it is just a dummy value.

alantang888: Done, just tested with simple config is ok.

pkg/vault/operator_client.go
```diff
@@ -971,3 +979,14 @@ func getMountConfigInput(secretEngine map[string]interface{}) (api.MountConfigIn
 
 	return mountConfigInput, nil
 }
+
+func isConfigNoNeedName(secretEngineType string, configOption string) bool {
+	if configOption == "config" {
+		_, ok := secretEngineConfigNoNeedName[secretEngineType]
+		if ok {
```
bonifaido: You can simply `return ok` here instead of the `if`.

alantang888: oh right!! that's redundant...

PR #279, opened 2019-01-16
Add container resource config

### LINKED ISSUE #273: Resources for containers
At the moment when operator creates a pod for vault and banks-vault containers it doesn't specify any resources requests or limits, so it's hard to manage and calculate the capacity for the pods.

I'd like to be able to specify resources in CRD and have some default resources specified in the operator.

Provide static resource requests and limits for prometheus-exporter and banks-vault container, this way vault container can get its resource requests and limits form limit ranger (without that every container in the pod gets the same requests/limits from limit ranger).

alantang888:
For #273. And I need this too :P

Fixes #273


baluchicken:
Hi @alantang888, thanks for the PR you modified vault_types.go, you may know we are using [operator-sdk](https://github.com/operator-framework/operator-sdk) to build our operator. This project requires to rerun `operator-sdk generate k8s` when you change the go file which contains the type information. Because of a bug in that sdk we need to modify the generated code just a little bit, so I will update your PR.

tomaszkiewicz:
@alantang888 Thanks for a quick implementation :)

I have one comment: when you enable metrics then additional container "prometheus-exporter" is also added to the pod.
I also see that there's additonal pod - "vault-configurer" deployed which also misses some resources.

Would you like me to open another issue for above cases?

Best regards

Łukasz Tomaszkiewicz

alantang888:
Thanks. I'll take a look on `operator-sdk` later.

alantang888:
@tomaszkiewicz `vault-configurer` deployment should follow `bankVaultsResource` resource config.

`prometheus-exporter` should just use very few resource. Should set resource same as `bank-vaults`? Or have an extra spce?? Or create a type call `resources` under spec. Then contain 3 different kind container's ResourceRequirements. Like:
```
spec:
  size: 3
  image: vault:1.0.0

  # A YAML representation of resource ResourceRequirements for vault containers
  # Detail can reference: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container
  resources:
    vault:
      requests:
        memory: "256Mi"
        cpu: "200m"
      limits:
        memory: "2Gi"
        cpu: "2"
    bankVaults:
      requests:
        memory: "64Mi"
        cpu: "100m"
      limits:
        memory: "128Mi"
        cpu: "500m"
    prometheusExporter:
      requests:
        memory: "64Mi"
        cpu: "100m"
      limits:
        memory: "128Mi"
        cpu: "500m"
```

@bonifaido @baluchicken any suggestion??

bonifaido:
Hi,

> vault-configurer deployment should follow bankVaultsResource resource config.

You're completely right, and this resource shouldn't be that huge, let's say 64Mi and 100m or something like that.

> prometheus-exporter should just use very few resources. Should set resource same as bank-vaults? 

I agree and disagree. I think this should be a separate resource since the implementation is not controlled by us, but I agree that it probably has a small footprint.

My suggestion is that we should change these resource fields optional (via changing them to pointers in the struct) and if they are not defined we should always apply a sane default limit for every container that the operator deploys.

Does this make sense?

alantang888:
Agree. the thing is default request and limit values. Above example is roughly estimate for my case.
And structure of spec. eg: 
- Use 3 ResourceRequirements just under spec for 3 different container
- Or create a new type just under spec to contain 3 ResourceRequirements

bonifaido:
I think the second would be better to make things a bit more organized.

alantang888:
I done the code logic. But stuck on `operator-sdk generate k8s` =.=". Can't test it yet...
https://github.com/alantang888/bank-vaults/tree/enhance_resource

I tried run `export GO111MODULE=on; go mod vendor; mv ../vendor/ .` under operator directory. Also clone code-generator to `vendor/k8s.io`. Still no luck...

bonifaido:
Yes (and sorry for the inconvinience) currently the operator-sdk can't generate code in a target directory, only in the `CWD`.
The method to make this work is the following, which I'm adding to the documentation right now:

```bash
cd bank-vaults/operator
ln -s ../vendor vendor
operator-sdk generate k8s
```

Issue #330, opened 2019-02-19
Bank-Vaults configure should support multiple configuration files and directories


bonifaido:
See https://github.com/banzaicloud/bank-vaults/issues/323 for more details.

primeroz:
Is there an example on how to use this feature when running the operator ?

bonifaido:
Hi @primeroz, sorry there is no specific docs for this, but there is a sample ConfigMap that can be added as an external config: https://github.com/banzaicloud/bank-vaults/blob/master/operator/deploy/config.yaml

PR #339, opened 2019-02-25
configurer: support multiple configuration files

bonifaido:
Fixes: #330

### DIFF
--- a/cmd/bank-vaults/configure.go
+++ b/cmd/bank-vaults/configure.go
@@ -16,7 +16,6 @@ package main
 
 import (
 	"bytes"
-	"path"
 	"path/filepath"
 	"text/template"
 	"time"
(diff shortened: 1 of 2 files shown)


cmd/bank-vaults/configure.go
```diff
@@ -69,65 +69,71 @@ var configureCmd = &cobra.Command{
...
 
-			err = viper.ReadConfig(buffer)
-			if err != nil {
-				logrus.Fatalf("error reading vault config file: %s", err.Error())
+				err := configTemplate.ExecuteTemplate(buffer, templateName, nil)
+				if err != nil {
+					logrus.Fatalf("error executing vault config template: %s", err.Error())
+				}
+
+				println("buffer:", buffer.String())
```
baluchicken: I am not sure:) if this is needed or you wanted a logrus.info

bonifaido: You're right, thanks, fixed! :)

Issue #341, opened 2019-02-25
[vault-env] clean up the environment #feature


dennybaa:
Just tried `vault-env` in a raw mode without Kubernetes, i.e. I set `VAULT_{ADDR,TOKEN}`. What seems strange to me that these guys are not wiped, I do understand that this should be one-time use token with TTL... But is there any consideration why this is not done?

bonifaido:
Hi Denis, no this is not intentional, they are just not cleared. Are you willing to open up a PR for this?

dennybaa:
Well, maybe, since I have got almost no golang experience. Though should be trivial, I'll ping you back later this week if I succeed or don't.

bonifaido:
No worries, happy to help.

PR #343, opened 2019-03-01
[vault-env] added VAULT_* environment sanitization

dennybaa:
Hello @bonifaido,

I tried to play around with golang and have come up with this. It addresses #341 issue. Please let me know your considerations.

### DIFF
--- a/cmd/vault-env/main.go
+++ b/cmd/vault-env/main.go
@@ -26,6 +26,39 @@ import (
 	"github.com/spf13/cast"
 )
 
+type sanitizedEnviron []string
+
+var sanitizeEnvmap = map[string]bool {
+	"VAULT_TOKEN": true,
+	"VAULT_ADDR": true,
+	"VAULT_CACERT": true,
+	"VAULT_CAPATH": true,
+	"VAULT_CLIENT_CERT": true,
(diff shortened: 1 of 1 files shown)


cmd/vault-env/main.go
```diff
@@ -26,6 +26,45 @@ import (
...
+	"VAULT_MFA",
+	"VAULT_ROLE",
+	"VAULT_PATH",
+}
+
+// Appends variable an entry (name=value) into the environ list.
+// VAULT_* variables are not populated into this list.
+func (environ *sanitizedEnviron) append(name interface{}, value interface{}) {
+	name, value := name.(string), value.(string)
+	if sanitizeEnvmap == nil {
```
bonifaido: I think you can fill this map in var declaration, so this list and the loop is not needed.

dennybaa: actually might, but I don't know how... btw fixed the compilation issues I believe.

dennybaa: you mean?
```go
var keys = map[string]bool{
  "VAULT_x": true,
}
```

bonifaido: Exactly!

PR #345, opened 2019-03-05
example: kv secret engine used for leader election logic

bonifaido:


vault-config.yml
```diff
@@ -111,6 +111,17 @@ secrets:
     options:
       version: 2
 
+  # Mounts kv with extra configuration
+  - path: leaderelecetion
```
raoofm: s/leaderelecetion/leaderelection

bonifaido:
Thanks @raoofm, fixed!

PR #347, opened 2019-03-06
Env update

hydratim:
Adds rudimentary support for update / write style requests to the webhook.

### Reason:
It wasn't possible to generate certificates via the webhook as this required a write to the `intermediate-ca/issue/cert-name` .

### Solution:
Added write support to vault-env via a prepend to the vault URI.
The proposed prepend that has been implemented is `>>`

### Example:
If you wanted to issue a new cert you would use the URI `>>vault:intermediate-ca/issue/cert-name`.

### Future:
It might also be good to support data in the write (it is currently an empty write). This could be done via an append to the path. (E.G. `>>vault:intermediate-ca/issue/cert-name{}` where the {} is json data)

### DIFF
--- a/cmd/vault-env/main.go
+++ b/cmd/vault-env/main.go
@@ -97,7 +97,7 @@ func main() {
 			path = split[0]
 
 			var key string
-			if len(split) > 0 {
+			if len(split) > 1 {
 				key = split[1]
 			}
 			
--- a/cmd/vault-env/main.go
+++ b/cmd/vault-env/main.go
@@ -86,10 +86,10 @@ func main() {
 		value := split[1]
 		var update bool
 		if strings.HasPrefix(value, ">>") {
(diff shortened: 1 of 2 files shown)


bonifaido:
Hi @hydratim, I really like this, this was on the roadmap already :) The syntax is tricky make it good here, but for the initial solution I'm happy with `>>` at first.

Some notes, please rebase your commits to the same author, and that author must have a GitHub user as well, otherwise to sign with the CLAassitant can become tricky (your first commit is good, but the second two are made with a different email I guess).

hydratim:
Thanks, will do.

hydratim:
All rebased.

bonifaido:
I still see two different e-mail addresses.

hydratim:
Yeah, don't know why that didn't clear, anyway I've added it to my account and it's green now.

PR #349, opened 2019-03-06
Update defaultKeyBits to 2048

raoofm:
Was that intentional to have 2038? If yes then I can close this pr. Isn't 2048 the standard?


bonifaido:
This should be a typo only I guess. Thanks for fixing!

raoofm:
thanks @bonifaido

PR #369, opened 2019-03-25
Add support for creating vault configurations as Secrets

primeroz:
Uses the same pattern as the configmaps, it will be useful when used alongside
sealed-secrets controller (or similar) to push sensitive vault configurations into git


primeroz:
Wrong user commit - CLA Issues

Issue #372, opened 2019-03-25
make service ports configurable


jurgenweber:
**Is your feature request related to a problem? Please describe.**
https://github.com/banzaicloud/bank-vaults/blob/master/operator/pkg/controller/vault/vault_controller.go#L464-L492

I need the ability to add ports to the default 8200/8201 which seem to be hard coded. This is because I have added multiple TCP listeners to vault, one for in cluster communication and one for external cluster communication both having their only TLS certificates to handle such.

**Describe the solution you'd like**
I want to be able to change the currently hard coded default ports and add ports to the service.

**Describe alternatives you've considered**
Right now the only option is to do it manually.

**Additional context**
Thanks. :)

jurgenweber:
working and live in 0.4.15-rc.3.

PR #373, opened 2019-03-26
add the ability for custom ports

jurgenweber:
https://github.com/banzaicloud/bank-vaults/issues/372

### DIFF
--- a/operator/deploy/cr-customports.yaml
+++ b/operator/deploy/cr-customports.yaml
@@ -1,0 +1,134 @@
+apiVersion: "vault.banzaicloud.com/v1alpha1"
+kind: "Vault"
+metadata:
+  name: "vault"
+spec:
+  size: 1
+  image: vault:1.0.0
+  bankVaultsImage: banzaicloud/bank-vaults:latest
+
+  # Support for custom Vault (and sidecar) pod annotations
+  annotations:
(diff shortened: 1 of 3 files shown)


operator/deploy/cr-customports.yaml
```diff
@@ -0,0 +1,134 @@
...
+    kubernetes:
+      secretNamespace: default
+
+  # A YAML representation of a final vault config file.
+  # See https://www.vaultproject.io/docs/configuration/ for more information.
+  config:
+    storage:
+      file:
+        path: "/vault/file"
+    listener:
```
bonifaido: I would change this to have multiple listeners as discussed on Slack, please add port 8300 as well.

operator/pkg/controller/vault/vault_controller.go
```diff
@@ -463,6 +463,7 @@ func etcdForVault(v *vaultv1alpha1.Vault) (*etcdV1beta2.EtcdCluster, error) {
 
 func serviceForVault(v *vaultv1alpha1.Vault) *corev1.Service {
 	ls := labelsForVault(v.Name)
+	ports := getServicePorts(v)
```
bonifaido: Please add these ports to the vault container as well.

jurgenweber:
ok, updated.

PR #376, opened 2019-03-26
append to the slice, not overwrite the slice

jurgenweber:
fix bug in `0.4.15-rc.1` that I introduced when setting custom ports, the slice is overwriting. Need to append to the declared slice.


jurgenweber:
updated and ready to go now

PR #378, opened 2019-03-27
add --once flag to configure and respect VAULT_TOKEN in dev mode

bonifaido:


cmd/bank-vaults/configure.go
```diff
@@ -165,10 +167,14 @@ var configureCmd = &cobra.Command{
...
 					if err = v.Configure(config); err != nil {
 						logrus.Errorf("error configuring vault: %s", err.Error())
+
+						exitIfNecessary(1)
 						return
 					}
 
 					logrus.Infof("successfully configured vault")
+
+					exitIfNecessary(0)
```
primeroz: Tested with 2 files and it exited after the first one was successfully applied 

Is this in the right Spot ? should it not be outside the `for` ?

```
docker run --rm -t -i -e VAULT_TOKEN=XXXXXXX -e VAULT_ADDR=http://127.0.0.1:8200 --net=host -v /tmp:/tmp banzaicloud/bank-vaults:configure-once configure --once --mode dev --vault-config-file=/tmp/config.yaml --vault-config-file=/tmp/config2.yaml
INFO[0000] New config file change /tmp/config.yaml      
INFO[0000] checking if vault is sealed...               
INFO[0000] vault is unsealed, configuring...            
INFO[0000] successfully configured vault   
```

bonifaido: Thanks @primeroz, just fixed this with some refactoring.

Issue #362, opened 2019-03-15
Publish a Prometheus Metric for the outcome of bank-vaults configure


primeroz:
**Describe the solution you'd like**
bank-vaults configure will , at the moment, only report an issue in applying the policies to the logs. 
In a GitOps driven scenario where the policies are updated and pulled by the cluster we need to know if the last run of bank-vaults configure was successful and the number of policies that were applied. 

Exposing a prometheus metric would make this information useful for alerting and graphing purposes

Would be great to also report the status in the "vault" object

**Describe alternatives you've considered**
Alerting by parsing logs

bonifaido:
Fixed by: https://github.com/banzaicloud/bank-vaults/pull/379

PR #379, opened 2019-03-27
Prometheus metrics for configurator mode 

primeroz:
Add first implementation of prometheus metric about last configuration run

https://github.com/banzaicloud/bank-vaults/issues/362

prometheus code is shared between unseal and configure operation so I added a flag to the prometheus struct to select which metrics to register. 

it exposes 2 gauge , always increasing, metrics 
 * vault_config_successful - 1 for each successfully applied file
 * vault_config_failed - 1 for each failed applied file

ToDo:
* Port 9091 is not set on the statefulset nor the configurator


bonifaido:
The CI test started to be very flaky, I think this is because of newly introduced code paths, so I will increase some timeouts.

pbalogh-sa:
@primeroz , please set 9091 in deploymentForConfigurer in vault_controller.go

primeroz:
@pbalogh-sa I don't see it set for the statefulset either, i will set it there too

pbalogh-sa:
Looks good @primeroz

primeroz:
Pull Request ready for review, few things i changed in the last commit 

* Added a new label to the _main_ vault service: **global_service**=true 
  This is needed for CoreOS Prometheus Operator servicemonitor to only pick up the main service and avoid duplicate metrics 
* Port 9091 to the Vault Services 
* Vault Configurator Service

Issue #382, opened 2019-03-28
vault 1.1.0 no longer required statsd exporter


jurgenweber:
**Is your feature request related to a problem? Please describe.**
in vault 1.1.0 a prometheus scrape endpoint has been introduced.
https://github.com/hashicorp/vault/blob/master/website/source/docs/configuration/telemetry.html.md#prometheus

**Describe the solution you'd like**
Add the ability to disable the side car (to maintain backwards compaitibility) in favour of the native support.

matyix:
We will keep this option as well - makes no harm, someone might still prefer this - but behind a flag.

primeroz:
Let's keep the statsd optional please. 

At least for now is much easier for me to get a monitoring setup out of the box without having to sort out tokens for prometheus.

bonifaido:
I absolutely agree Prometheus setup with the token is harder currently.

PR #386, opened 2019-03-29
Add metrics policy for prometheus

matyix:


matyix:
Implicitly fixes #382

matyix:
@jurgenweber @bonifaido there are quite a few ways to add a Vault token for Prometheus, let's discuss this on Slack what would be the preferred way of automation - meanwhile, you can add the token into a Secret to make it work.

PR #408, opened 2019-04-05
Make Watch function work on multiple Directories

### LINKED ISSUE #405: watchConfigurations in configure.go only watch the first directory in the list of vaultconfigFiles
**Describe the bug**
When specifying multiple configs we want the Watcher to watch all directories and feed back the configuration file that changed in the "configurations" channel.

At the moment only the first file in the vaultConfigFiles get a watcher though and so only that file changes trigger a reconfiguration form the vault-configure

```
diff --git a/cmd/bank-vaults/configure.go b/cmd/bank-vaults/configure.go
index ed9347b..bd9b265 100644
--- a/cmd/bank-vaults/configure.go
+++ b/cmd/bank-vaults/configure.go
@@ -87,6 +87,7 @@ var configureCmd = &cobra.Command{
                }

                if !runOnce {
+                       logrus.Info("######## launching watcher #####")
                        go watchConfigurations(vaultConfigFiles, configurations)
                } else {
                        close(configurations)
@@ -135,6 +136,9 @@ var configureCmd = &cobra.Command{

 func watchConfigurations(vaultConfigFiles []string, configurations chan *viper.Viper) {                                                                                   
        watcher, err := fsnotify.NewWatcher()
+       logrus.Info("######## Watcher created #####")
+       logrus.Infof("CONFIGS: %#v", vaultConfigFiles)
+       logrus.Infof("CONFIGS length: %s", len(vaultConfigFiles))
        if err != nil {
                logrus.Fatal(err)
        }
@@ -150,9 +154,16 @@ func watchConfigurations(vaultConfigFiles []string, configurations chan *viper.V                                                                      
                        for {
                                select {
                                case event := <-watcher.Events:
+                               logrus.Info("######## NEW EVENT #####")
+                               logrus.Infof("Configfile name: %s", configFile)
+                               logrus.Infof("Basepath name: %s", filepath.Base(event.Name))                                                                               
+                               logrus.Infof("event name: %s", event.Name)
+                               logrus.Infof("event op: %s", event.Op)
                                        // we only care about the config file or the ConfigMap directory (if in Kubernetes)                                                
                                        if filepath.Clean(event.Name) == configFile || filepath.Base(event.Name) == "..data" {                                             
+                                       logrus.Info("Matched filename")
                                                if event.Op&fsnotify.Write == fsnotify.Write || event.Op&fsnotify.Create == fsnotify.Create {                              
+                                         logrus.Info("Matched Operation")
                                                        configurations <- parseConfiguration(configFile)                                                                   
                                                }
                                        }
@@ -161,7 +172,7 @@ func watchConfigurations(vaultConfigFiles []string, configurations chan *viper.V                                                                       
                                }
                        }
                }()
-
+               logrus.Infof("WATCHER Added: %s", configDir)
                watcher.Add(configDir)
                <-done
        }
```
when running with 2 files shows the watcher only on the first file 
```
INFO[0000] ######## launching watcher #####             
INFO[0000] config file has changed: /p/dir1/link.yaml   
INFO[0000] checking if vault is sealed...               
INFO[0000] ######## Watcher created #####               
INFO[0000] CONFIGS: []string{"/p/dir1/link.yaml", "/p/dir2/real.yaml"} 
INFO[0000] CONFIGS length: %!s(int=2)         
          
INFO[0000] WATCHER Added: /p/dir1/            
          
INFO[0000] vault is unsealed, configuring...            
INFO[0000] successfully configured vault                
INFO[0000] config file has changed: /p/dir2/real.yaml   
INFO[0000] checking if vault is sealed...               
INFO[0000] vault is unsealed, configuring...            
INFO[0000] successfully configured vault      
```

**To Reproduce**
1. Start a local vault `docker run --rm -t -i -e SKIP_SETCAP=true -e VAULT_DEV_ROOT_TOKEN_ID=XXXX --net=host vault:1.1.0 server -dev -dev-listen-address=0.0.0.0:8200`
2. Create 2 config files in /tmp/dir1 and /tmp/dir2
3. Start bank-vaults configurator with "--vault-config-file=/tmp/dir1/file.yaml --vault-config-file=/tmp/dir2/file.yaml" 
`docker run --rm -t -i -e VAULT_TOKEN=XXXX -e VAULT_ADDR=http://127.0.0.1:8200 --net=host -v /tmp:/tmp banzaicloud/bank-vaults:watch-files configure --fatal --mode dev --vault-config-file=/tmp/dir1/file.yaml --vault-config-file=/tmp/dir2/file.yaml`
4. Test that changes to dir1/file.yaml trigger reconfigurations while dir2/file.yaml does not 

**Expected behavior**
All unique directories should be watched for changes

**Additional context**
I tried to mess with the "<-done" but every change to it caused infinite loops

primeroz:
Fixes #405 

* Watch multiple directories
* Reload on WRITE on Real Files
* Reload on CREATE on Kubernetes re-link of ..data for any number of files in the configMap
* The Events watching does not need to run in its own GO thread since the whole function is already running as such
* We don't have access to the Filename in the event watching but just the event.Name
  * For real files we can safely use that
  * For Kubernetes Configmaps we need to path the basepath to a saved map of DIR: FILES and just reload all files

### DIFF
--- a/cmd/bank-vaults/configure.go
+++ b/cmd/bank-vaults/configure.go
@@ -181,7 +181,6 @@ func watchConfigurations(vaultConfigFiles []string, configurations chan *viper.V
 				logrus.Infof("File has changed: %s", event.Name)
 				configurations <- parseConfiguration(filepath.Clean(event.Name))
 			} else if event.Op&fsnotify.Create == fsnotify.Create && filepath.Base(event.Name) == "..data" {
-				logrus.Infof("Files : %v", configFileDirs[filepath.Dir(event.Name)])
 				for _, fileName := range configFileDirs[filepath.Dir(event.Name)] {
 					logrus.Infof("ConfgMap has changed, reparsing: %s", fileName)
 					configurations <- parseConfiguration(fileName)
--- a/cmd/bank-vaults/configure.go
+++ b/cmd/bank-vaults/configure.go
@@ -176,14 +176,14 @@ func watchConfigurations(vaultConfigFiles []string, configurations chan *viper.V
 		select {
 		case event := <-watcher.Events:
 			// we only care about the config file or the ConfigMap directory (if in Kubernetes)
-			// For real Files we only need to watch thw WRITE Event # TODO: Sometimes it triggers 2 WRITE when a file is edited and saved
+			// For real Files we only need to watch the WRITE Event # TODO: Sometimes it triggers 2 WRITE when a file is edited and saved
 			// For Kubernetes configMaps we need to watch for CREATE on the "..data"
 			if event.Op&fsnotify.Write == fsnotify.Write && stringInSlice(filepath.Clean(event.Name), vaultConfigFiles) {
 				logrus.Infof("File has changed: %s", event.Name)
 				configurations <- parseConfiguration(filepath.Clean(event.Name))
 			} else if event.Op&fsnotify.Create == fsnotify.Create && filepath.Base(event.Name) == "..data" {
 				for _, fileName := range configFileDirs[filepath.Dir(event.Name)] {
-					logrus.Infof("ConfgMap has changed, reparsing: %s", fileName)
+					logrus.Infof("ConfigMap has changed, reparsing: %s", fileName)
 					configurations <- parseConfiguration(fileName)
 				}
 			}
--- a/cmd/bank-vaults/configure.go
+++ b/cmd/bank-vaults/configure.go
@@ -158,14 +158,15 @@ func watchConfigurations(vaultConfigFiles []string, configurations chan *viper.V
 		// we have to watch the entire directory to pick up renames/atomic saves in a cross-platform way
 		configFile := vaultConfigFile
 		configDir, _ := filepath.Split(configFile)
+		configDirTrimmed := strings.TrimRight(configDir, "/")
 
 		files := make([]string, 0)
-		if len(configFileDirs[strings.TrimRight(configDir, "/")]) != 0 {
-			files = configFileDirs[strings.TrimRight(configDir, "/")]
+		if len(configFileDirs[configDirTrimmed]) != 0 {
+			files = configFileDirs[configDirTrimmed]
 		}
 		files = append(files, configFile)
 
-		configFileDirs[strings.TrimRight(configDir, "/")] = files
+		configFileDirs[configDirTrimmed] = files
 
 		logrus.Infof("Watching Directory for changes: %s", configDir)
 		watcher.Add(configDir)
--- a/cmd/bank-vaults/configure.go
+++ b/cmd/bank-vaults/configure.go
@@ -167,7 +167,7 @@ func watchConfigurations(vaultConfigFiles []string, configurations chan *viper.V
 
 		configFileDirs[strings.TrimRight(configDir, "/")] = files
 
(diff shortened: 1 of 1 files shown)


cmd/bank-vaults/configure.go
```diff
@@ -133,38 +135,62 @@ var configureCmd = &cobra.Command{
...
-							configurations <- parseConfiguration(configFile)
-						}
-					}
-				case err := <-watcher.Errors:
-					logrus.Error(err)
-				}
-			}
-		}()
+		files := make([]string, 0)
+		if len(configFileDirs[strings.TrimRight(configDir, "/")]) != 0 {
```
bonifaido: I think `configFileDirs[strings.TrimRight(configDir, "/")]` could be extracted to a `var` here.

cmd/bank-vaults/configure.go
```diff
@@ -133,38 +135,62 @@ var configureCmd = &cobra.Command{
...
+		case event := <-watcher.Events:
+			// we only care about the config file or the ConfigMap directory (if in Kubernetes)
+			// For real Files we only need to watch thw WRITE Event # TODO: Sometimes it triggers 2 WRITE when a file is edited and saved
+			// For Kubernetes configMaps we need to watch for CREATE on the "..data"
+			if event.Op&fsnotify.Write == fsnotify.Write && stringInSlice(filepath.Clean(event.Name), vaultConfigFiles) {
+				logrus.Infof("File has changed: %s", event.Name)
+				configurations <- parseConfiguration(filepath.Clean(event.Name))
+			} else if event.Op&fsnotify.Create == fsnotify.Create && filepath.Base(event.Name) == "..data" {
+				for _, fileName := range configFileDirs[filepath.Dir(event.Name)] {
+					logrus.Infof("ConfgMap has changed, reparsing: %s", fileName)
```
bonifaido: Typo: `ConfgMap`

cmd/bank-vaults/configure.go
```diff
@@ -133,38 +135,62 @@ var configureCmd = &cobra.Command{
...
+		logrus.Infof("Watching Directory for changes: %s", configDir)
 		watcher.Add(configDir)
-		<-done
 	}
+
+	for {
+		select {
+		case event := <-watcher.Events:
+			// we only care about the config file or the ConfigMap directory (if in Kubernetes)
+			// For real Files we only need to watch thw WRITE Event # TODO: Sometimes it triggers 2 WRITE when a file is edited and saved
```
bonifaido: For spurious writes, we could employ a rate limiter here like https://github.com/uber-go/ratelimit but if this doesn't cause any issues right now I'm happy with the current solution :)

bonifaido: Typo: `thw`

cmd/bank-vaults/configure.go
primeroz: done

cmd/bank-vaults/configure.go
primeroz: done

cmd/bank-vaults/configure.go
primeroz: i don't think is a huge deal for now, and i could only reproduce it when editing a file directly in VIM ... i don't think it will happen in kubernetes. 

We might wat to find out why it is happening though

primeroz: done

PR #409, opened 2019-04-06
operator: allow setting of etcd resources

pickledrick:
As etcd is quite an important component should we choose to use it for vault storage it would be great if we could set resources on this component in the operator to be passed down to the etcd operator.


pickledrick:
Thanks a lot @bonifaido @matyix  for this operator, A perfect case for one! It has saved me a lot of time to get up and going with Vault.

matyix:
Happy that we could help @pickledrick - please let us know if you face issues, etc. There is an active community around the operator at https://community-banzaicloud.slack.com/messages/CFJJW9L94, join us there if you are interested in roadmaps, features/issues we are currently working, or if you have any questions you might not want to address on GitHub.

Issue #403, opened 2019-04-04
Add consul-template support


jurgenweber:
**Current State:**

The current webhook only supports the KV store and KPI for our use case we would need support for the AWS and database secrets backends.

Doing it using Seth Vargo’s method is a lot of work, taking all that manual config and CnPing it everywhere. I have been looking at the code for the webhook (https://github.com/banzaicloud/bank-vaults/blob/master/cmd/vault-secrets-webhook/main.go, https://github.com/banzaicloud/bank-vaults/blob/master/cmd/vault-env/main.go) and I want to explain what I have worked out/reverse engineered so I know we are on the same page and then ask some questions. 

the general workflow:
• you look for pods with the right EnVars/ConfigMaps/Secrets with the correct prefix ‘vault:’ or ‘>>vault:’
• you then inject an init container running vault that gets a vault token and copies it into the /vault/ volume
• you inject another init container that copies `vault-env` into the /vault/ volume
• you then inspect every container in the pod, add the required env vars for `vault-env`
• you then concat the container command and args and turn that into the args and make `/vault/vault-env` the command
• you then run /vault/vault-env <original command> <original args>
• `vault-env` parses the env/configmaps/secrets updates the env/configmaps and secrets per the correct strings (value: vault:secret/data/accounts/aws#AWS_SECRET_ACCESS_KEY)
• `vault-env` then executes the binary (<original command> <oringinal args>)

KV/KPI are only supported because when you make the request to vault you are only handling the returned data for those backends. To support other backends we need to add smarts to handle the returning of passwords and usernames or a data object for that secrets backend.

`vault-env` is only run once when the container starts, meaning only those env/ secrets/configmaps are only manipulated once. This means it does not handle expiring credentials/a TTL on a returned credential because this would mean manipulating the current running containers or triggering a replica set update.

I would like to suggest a new feature of adding `consul-template` as an additive to `vault-env`, not as a replacement but as something that can be used in conjunction to solve a different problem. The difference are:
`vault-env`:
• limited support for secret backends.
• does not support TTL.
• has cool env/secret/config map updating mechanism.
• takes over the live container.
• does not deal with files.

`consul-template`
• supports all backends out of the box with its templating.
• runs as a side car daemon and handles TTL on secrets.
• writes config to a file system (preferably a memory resident one so it disappears on tear down) for the app to pick up.
• can HUP/run a command to tell the process to take the updated config.
• does not deal with env/secrets/config maps.

**Implementation:**
[...]

Issue #380, opened 2019-03-28
Have the operator watch an externally managed Secret and reload vault when it change


primeroz:
**Is your feature request related to a problem? Please describe.**
Thanks to the multiple listener support , [this config](https://github.com/banzaicloud/bank-vaults/blob/master/operator/deploy/cr-customports.yaml) is now easy to have an external endpoint use letsencrypt certificate for e2e encryption

Those certs are renewed frequently and when that happen Vault need to restarted to pick up the change ( In the future we could signal HUP vault but that will require pid sharing namespace to be widely available )

**Describe the solution you'd like**
Have the operator watch the secret created by "cert-manager" , or any other way really, and trigger a reload of the vault statefulset when it change.

It could even just look for any secret matching a set of labels, similar to how configs are managed.

**Describe alternatives you've considered**
I am gonna try run "https://github.com/stakater/Reloader" to achieve the same behavior

primeroz:
Unfortunately using "Reloader" is not feasible due to the annotations not propagating to the statefulset itself 

see https://github.com/banzaicloud/bank-vaults/issues/357

bonifaido:
This annotation feature is no big deal I think, if this is the only blocker :)

bonifaido:
My only concern is adding another deployment/pod into the mix (the Reloader). I think the operator could manage this. What do we mean by restarting? Deleting the pod? (That would be easy and more secure IMHO). Or exec-ing into the pod to send a signal?

primeroz:
Yeah sorry , i was going to run the "reloader" on my own while waiting for this to be implemented.
The reason why i added the comment is that i mentioned the "reloader" as the alternatives i've considered when opening the bug.

I think the operator should just trigger a roll-restart of the statefulset set . 

Looking at how reloader is doing it ( https://github.com/stakater/Reloader/blob/master/internal/pkg/callbacks/rolling_upgrade.go#L109) with my limited knowledge of go and kubernetes api , it seem they are just hitting an Update on the statefulset resource ? 

If i was going to do it with kubectl and manifests i would just use an annotation with a checksum of the secrets and update that to let kubernetes trigger a roll restart 

Regardless i would just let kubernetes statefulset deal with the roll restart 

I would only do the SIGHUP to vault using a sidecar container  , but since that requires the PID SHARED NAMESPACE which is still Beta in 1.14 i doubt is a good idea right now

PR #410, opened 2019-04-08
Add consul-template support

jurgenweber:
https://github.com/banzaicloud/bank-vaults/issues/403


cmd/vault-secrets-webhook/main.go
```diff
@@ -109,6 +111,53 @@ func getInitContainers(vaultConfig vaultConfig) []corev1.Container {
...
+
+	containers = append(containers, corev1.Container{
+		Name:			"consul-template",
+		Image:			viper.GetString("vault_ct_image"),
+		Args: 			[]string{"-config","/etc/ct-config/config.hcl"},
+		ImagePullPolicy: corev1.PullIfNotPresent,
+		SecurityContext: &corev1.SecurityContext{
+			Capabilities: &corev1.Capabilities{
+				Add: []corev1.Capability{
+					"SYS_PTRACE",
```
pbalogh-sa: Why does it need for SYS_PTRACE?

jurgenweber: as per Seth Vargo's example; https://github.com/sethvargo/vault-kubernetes-workshop/blob/master/k8s/db-sidecar.yaml#L59-L61

It is only for when you have a version of k8s that supports the sharePorcessNamespace key/feature gate. I have not completed the logic for handling that yet in all the places. eg. here. I am hoping tomorrow I can work out a way to detect the feature in the api server, etc.

pbalogh-sa: OK, thx

jurgenweber: I am going to go for a manual switch, there is no convenient way to find out what feature-gates have been turned on. In 1.12 it is a enabled by default (https://v1-12.docs.kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/) so I will use this as the min version to auto have it on and have a manual override for those in older versions.

primeroz: should we have a global flag for "PID SHARE PROCESS NAMESPACE" Rather than a specific CT annotation ? 

This will be a useful flag for #380 as well where bank-vaults might just SIGHUP vault to reload tls secrets on disk

pbalogh-sa: @primeroz , it's a good idea.

jurgenweber:
@bonifaido Yeah, I have plans to do a CR and ConfigMap example. Maybe in it's own directory where all the current examples are.

Before we merge.... I need to make it all up and watch it work. :) Give me a few days. Early next week!

bonifaido:
No problem, take your time, and we are here/on Slack to help :)

cmd/vault-secrets-webhook/consul-template.md
```diff
@@ -0,0 +1,99 @@
+# Using consul-template in the mutating webhook
```
pbalogh-sa: @jurgenweber , We've just updated the docs structure. Would you to move this documentation to docs/mutating-webhook directory?

cmd/vault-secrets-webhook/consul-template.md
```diff
@@ -0,0 +1,99 @@
...
+* You have an application or tool that requires to read its configuration from a file.
+* You wish to have secrets that have a TTL and expire.
+* You do not wish to be limited on which vault secrets backend you use.
+
+below features only available with 'ShareProcessNamespace' available.
+* You wish to be able to expire tokens/revoke tokens i
+  (to do this you need to have a ready/live probe that can send a HUP to consul-template when the current details fail).
+
+## General concept
+Based on various publications and tools ([Kuernetes Authenticator](https://github.com/sethvargo/vault-kubernetes-authenticator), [consul-template](https://github.com/hashicorp/consul-template)) released and maintained by Hashicorp and [Seth Vargo](https://github.com/sethvargo).
```
pbalogh-sa: Typo

cmd/vault-secrets-webhook/main.go
```diff
@@ -431,6 +629,12 @@ func mutatePodSpec(obj metav1.Object, podSpec *corev1.PodSpec, vaultConfig vault
 func initConfig() {
 	viper.SetDefault("vault_image", "vault:latest")
 	viper.SetDefault("vault_env_image", "banzaicloud/vault-env:latest")
+	viper.SetDefault("vault_ct_image", "registry.hub.docker.com/sethvargo/consul-template:0.19.6.dev-alpine")
```
bonifaido: Can't we use the image from the official repository? https://hub.docker.com/r/hashicorp/consul-template/tags

cmd/vault-secrets-webhook/main.go
```diff
@@ -30,20 +30,25 @@ import (
...
-	addr       string
-	role       string
-	path       string
-	skipVerify string
-	useAgent   bool
+	addr                  string
+	role                  string
+	path                  string
+	skipVerify            string
+	tlsConfigMap          string
```
pbalogh-sa: @jurgenweber , please use tlsSecret instead of tlsConfigMap, becasue vault-tls is a secret

cmd/vault-secrets-webhook/main.go
```diff
@@ -169,17 +254,55 @@ func vaultSecretsMutator(ctx context.Context, obj metav1.Object) (bool, error) {
...
-	vaultConfig.skipVerify = annotations["vault.security.banzaicloud.io/vault-skip-verify"]
-	vaultConfig.useAgent, _ = strconv.ParseBool(annotations["vault.security.banzaicloud.io/vault-agent"])
+
+	if val, ok := annotations["vault.security.banzaicloud.io/vault-skip-verify"]; ok {
+		vaultConfig.skipVerify = val
+	} else {
+		vaultConfig.skipVerify = viper.GetString("vault_skip_verify")
+	}
+
+	if val, ok := annotations["vault.security.banzaicloud.io/vault-tls-configmap"]; ok {
```
pbalogh-sa: please use vault-tls-secret instead of vault-tls-configmap

cmd/vault-secrets-webhook/main.go
jurgenweber: maybe, I would have to test it still works. I can do that today.

But since the 'setup' I have taken from Seth Vargo's various publications I assume/think there is some 'stuff' in there that makes it work for him. I will admit it is a bit weird. This is why I wanted 0.20.0 but they have broken the functionality as shown in channel.

jurgenweber: ok, I changed it to `hashicorp/consul-template:0.19.6-dev-alpine` which seems to work.

bonifaido: Thanks

PR #411, opened 2019-04-08
Add support for gerenating wildcard certificates

baluchicken:


primeroz:
Did we document anywhere how to enable this ?

PR #417, opened 2019-04-11
Operator: add common unseal options

### LINKED ISSUE #396: Initialisation Fails when using Kube2IAM
The bank-vaults container in the first running pod restarts because when it tries to store the unseal keys in S3 it errors with access denied. The vault container is then initialised and the restarted bank-vaults container is unable to unseal it because the unseal keys have not made it to S3.

This is using S3 as storage with DynamoDB as HA storage and a customer managed key for S3 encryption. I have verified that the IAM role is able to put objects on S3 using the correct key for encryption.

My thoughts are perhaps bank-vaults is too fast to try and initialise vault before Kube2IAM can apply the role.

Logs from first bank-vaults:
```
time="2019-04-02T09:14:34Z" level=info msg="initializing vault"
time="2019-04-02T09:14:36Z" level=fatal msg="error initializing vault: error storing unseal key 'vault-unseal-0': error writing key 'vault-operator-vault-unseal-0' to s3 bucket 'vault': 'AccessDenied: Access Denied\n\tstatus code: 403'"
```

Logs from vault:
```
2019-04-02T09:14:34.621Z [INFO]  core: security barrier not initialized
2019-04-02T09:14:35.086Z [INFO]  core: security barrier not initialized
2019-04-02T09:14:35.244Z [INFO]  core: seal configuration missing, not initialized
2019-04-02T09:14:35.265Z [INFO]  core: security barrier initialized: shares=5 threshold=3
```

Logs from second bank-vaults:
```
time="2019-04-02T09:14:37Z" level=info msg="vault is already initialized"
time="2019-04-02T09:14:37Z" level=info msg="checking if vault is sealed..."
time="2019-04-02T09:14:37Z" level=info msg="vault sealed: true"
time="2019-04-02T09:14:37Z" level=error msg="error unsealing vault: unable to get key 'vault-unseal-0': error getting object for key 'vault-operator-vault-unseal-0': NoSuchKey: The specified key does not exist.\n\tstatus code: 404"
```

The time stamps from the logs support my theory.

Also see state of pods at present:
```
NAME                                          READY   STATUS    RESTARTS   AGE
vault-0                                         2/3     Running             0          32m
vault-1                                          2/3     Running            0          32m
vault-2                                          2/3     Running            1          32m
```

matyix:
Fixes #396

### DIFF
--- a/operator/pkg/apis/vault/v1alpha1/vault_types.go
+++ b/operator/pkg/apis/vault/v1alpha1/vault_types.go
@@ -301,6 +301,7 @@ type VaultList struct {
 
 // UnsealConfig represents the UnsealConfig field of a VaultSpec Kubernetes object
 type UnsealConfig struct {
+	Options    UnsealOptions           `json:"options,omitempty"`
 	Kubernetes *KubernetesUnsealConfig `json:"kubernetes,omitempty"`
 	Google     *GoogleUnsealConfig     `json:"google,omitempty"`
 	Alibaba    *AlibabaUnsealConfig    `json:"alibaba,omitempty"`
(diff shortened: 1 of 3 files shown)


operator/pkg/apis/vault/v1alpha1/vault_types.go
```diff
@@ -301,16 +301,27 @@ type VaultList struct {
...
 }
 
+// UnsealOptions represents the common options to all unsealing backends
+type UnsealOptions struct {
+	PreFlightChecks bool `json:"preFlightChecks,omitempty"`
+}
+
 // ToArgs returns the UnsealConfig as and argument array for bank-vaults
 func (usc *UnsealConfig) ToArgs(vault *Vault) []string {
+	var args []string
```
baluchicken:
```suggestion
	args := []string{}
```

PR #424, opened 2019-04-15
Handle Configuration errors by re-injecting configuration files into the configurations channel

### LINKED ISSUE #381: configurator: Retry apply a configuration when error is 5XX
While applying a configuration if the configurator encounter an error it just moves to the next file and never try that file again unless it changes ( https://github.com/banzaicloud/bank-vaults/blob/master/cmd/bank-vaults/configure.go#L110 ) 

This make sense for a 4XX error , where maybe the syntax was wrong in the configuration, but not for a 5XX error where the issue was on the server side.

In my case i hit a similar issue twice while upgrading the image of both vault and the vault-configurator at once ( since they both use the same field from the vault CRD ) 

```
+ vault-configurer-7dd4b74cc5-phjx5 › bank-vaults
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:00:59Z" level=info msg="config file has changed: /config/common-config/vault-config.yml"      
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:00:59Z" level=info msg="checking if vault is sealed..."                                       
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:01Z" level=info msg="vault is unsealed, configuring..."                                    
- vault-configurer-7dd4b74cc5-vp8gm

vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:09Z" level=error msg="error configuring vault: error configuring auth methods for vault: error listing auth backends vault: Error making API request.\n\nURL: GET https://vault.vault:8200/v1/sys/auth\nCode: 503. Errors:\n\n* Vault is sealed"            


vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:09Z" level=info msg="config file has changed: /config/dev-config/vault-config.yml"         
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:09Z" level=info msg="checking if vault is sealed..."                                       
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:45Z" level=info msg="vault is unsealed, configuring..."                                    
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:48Z" level=info msg="successfully configured vault"                                        
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:48Z" level=info msg="config file has changed: /config/vault-configurer/vault-config.yml"   
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:48Z" level=info msg="checking if vault is sealed..."                                       
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:48Z" level=info msg="vault is unsealed, configuring..."                                    
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:50Z" level=info msg="Already existing audit devices: map[string]*api.Audit{\"file/\":(*api.Audit)(0xc0002169c0)}\n"
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:50Z" level=info msg="audit device is already mounted: file/\n"                             
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:50Z" level=info msg="successfully configured vault"                                        
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:50Z" level=info msg="config file has changed: /config/kubernetes-dev1/vault-config.yml"    
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:50Z" level=info msg="checking if vault is sealed..."                                       
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:50Z" level=info msg="vault is unsealed, configuring..."                                    
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:53Z" level=info msg="Already existing mounts: map[string]*api.MountOutput{\"cubbyhole/\":(*api.MountOutput)(0xc0000ac680), \"identity/\":(*api.MountOutput)(0xc0000ac8f0), \"kv_sandbox/\":(*api.MountOutput)(0xc0000ad2b0), \"sys/\":(*api.MountOutput)(0xc0000ad5f0)}\n"
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:53Z" level=info msg="Tuning already existing mount: kv_sandbox/\n"                         
vault-configurer-7dd4b74cc5-phjx5 bank-vaults time="2019-03-28T17:01:54Z" level=info msg="successfully configured vault" 
```

I think what happened here is that when the configurator checked for the vault to be sealed it was not, when it got to apply the configuration it was and a 5XX error was returned.

The file /config/common-config/vault-config.yml was never reapplied though

IF error is 5XX then retry after a sleep 

None :)

primeroz:
fixes #381 

* Use backoff for all configuration files up to 30s so not to loop intensively over a syntax error
  * Counters will increase so a prometheus alert could notify of any syntax error that keeps failing to apply
* Any server side error should be sorted out with an _eventually consistent_ approach

I added a new module , https://github.com/jpillora/backoff, but i have no idea how to update _go.mod_ and _go.sum_ 

**UNRELATED** Set Gin Mode to release when building through CI so not to waste logs on debug messages from GIN


bonifaido:
Please run the following command to add it to go.mod

```bash
export GO111MODULE=on
go get github.com/jpillora/backoff
```

primeroz:
A summary of my discussion with @bonifaido on slack for anyone else reviewing this PR

The idea behind this error handler is to get an "eventually consistent" configuration into Vault.

Right now the configurator can hit errors while applying a configuration file which are recoverable and not due to a syntax error in the configuration file. #381 has a few example of those but in general they fall in the following categories:
* An error due to a transient issue on the server side like a failover in progress ( and so no active vault for a short period of time, this time can be shorter or longer depending on the backend used)
* An error due to a dependency between vault resources, this is especially true for Identity backend which requires the knowledge of resources IDs , those are only created once the resource is created

At the moment the configurator threat each file as a unit, any error in that file will cause the file to be skipped and **never be retried** until there is either a change in that file ( the watcher will re-inject the file in the channel) or a brand new configurator pod is create ( for whatever reason)

This Pull request handle this problem by re-injecting a configuration file that errored into the configurations channel, at the end of it, using an exponential backoff sleep up to a minute. 

This way it will lead to an eventually consistent configuration for **all recoverable errors* 

**Un-recoverable** errors though will keep on retrying every minute or so and they will keep on failing. 
Thanks to the configurator metrics exposed for prometheus though an alert is supposed to be raised to the operator to be notified that something in the configuration is wrong and need to be acted on.

This is on purpose , to not have a break out of the apply and increase error, and follow on the idea that having the visibility of a problem is better than a silent failure like we have now.

bonifaido:
First thanks for fixing this old issue 👍

For now, I think this implementation is good enough, it will prevent flaky CI errors from happening for example. In the future, we could improve the Prometheus metrics via separate labels for each configuration files and mark them failed as long as they are failed (but still keep the old metrics) and also pull in a backoff that won't loop forever in case of an unrecoverable error. I will create a separate issue for that.

Issue #423, opened 2019-04-15
code-generation is broken after go modules support was added


naemono:
**Describe the bug**

I was trying to add support for `NodeSelector` and `Tolerations` in the vault-operator, but this requires adjusting the `v1alpha1` custom resource definitions, which also requires re-running codegen, which seems to not work with go modules, as it appears that the operator-sdk doesn't support go modules?

see https://github.com/operator-framework/operator-sdk/pull/1001#issuecomment-464633707

**To Reproduce**
```
$ echo $GOPATH
/Users/mich9038/Go/Work
$ pwd
/Users/mich9038/Go/Work/src/github.com/banzaicloud/bank-vaults
$ GO111MODULE=on go mod download
$ GO111MODULE=on go mod vendor
$ cd operator/ && ln -s ../vendor
$ go version
go version go1.12.4 darwin/amd6
$ operator-sdk generate k8s
Error: chdir /Users/mich9038/Go/Work/src/github.com/banzaicloud/bank-vaults/operator/vendor/k8s.io/code-generator: no such file or directory
Usage:
  operator-sdk generate k8s [flags]

Flags:
      --header-file string   Path to file containing headers for generated files.
  -h, --help                 help for k8s
$ GO111MODULE=on go get -u k8s.io/code-generator
go: finding k8s.io/code-generator latest
go: finding k8s.io/gengo latest
go: finding golang.org/x/exp latest
go: finding golang.org/x/net latest
go: finding github.com/remyoudompheng/bigfft latest
go: finding golang.org/x/tools latest
go: finding gonum.org/v1/netlib latest
go: finding gonum.org/v1/gonum latest
go: finding golang.org/x/sys latest
go: finding golang.org/x/crypto latest
go: finding github.com/BurntSushi/xgb latest
go: finding golang.org/x/mobile latest
go: finding golang.org/x/image latest
package k8s.io/code-generator: build constraints exclude all Go files in /Users/mich9038/Go/Work/pkg/mod/k8s.io/code-generator@v0.0.0-20190413052309-5c40078c1b12
```

**Expected behavior**
Would expect code generation to work

**Desktop (please complete the following information):**
 - OS: MacOS 10.14.4
 - go version go1.12.4 darwin/amd6
 - operator-sdk version: v0.7.0+git

estroz:
@naemono you're going to have to add a top-level `tools.go` file to your repo:

```console
$ cat <<EOF > tools.go
// +build tools

package tools

import (
[...]

PR #427, opened 2019-04-16
Supporting nodeselectors, and tolerations

naemono:
Updating example Custom Resource
Adding needed 'tools.go' file to allow code generation to work with go modules
note discussion in banzaicloud/bank-vaults#423


bonifaido:
Could you please copy the license header from one of the Go files to the newly introduced tools.go?

naemono:
@bonifaido The license header was added, and I also added a  bit more information into the readme.

Let me know if you'd like me to rebase these down to a single commit.  thanks

Issue #228, opened 2018-12-11
Support Vault auto-unseal based deployments


bonifaido:
See https://github.com/banzaicloud/bank-vaults/pull/224 for the original thread.

raoofm:
+1

raoofm:
https://www.vaultproject.io/docs/configuration/seal/awskms.html

bonifaido:
#429 has fixed this issue partially, but we also have to disable unsealing in the operator (via parsing the Vault config) if we detect that auto-unseal is enabled.

primeroz:
a Few info on things to change when the operator detect that autounseal is enabled

* Setting of `secret_shares` and the `threshold_shares` to the same value of 1 ( see [here](https://www.vaultproject.io/api/system/init.html#secret_threshold) ) and the `stored_shares` set to the same as secret shares ( see [here](https://www.vaultproject.io/api/system/init.html#stored_shares) ) value of 1. This will happen automatically and logged as WARNING if not enforce
`vault-0 vault 2019-04-18T13:49:53.470Z [WARN]  core: stored keys supported on init, forcing shares/threshold to 1`
* when starting with autounseal disable the "bank-vaults unseal" loop to avoid errors like 
`vault-0 bank-vaults time="2019-04-18T13:56:41Z" level=error msg="error unsealing vault: unable to get key 'vault-unseal-0': error getting object for key 'vault-unseal-0': storage: object doesn't exist"`
  * Should we just have a sleep loop ?

jhohertz:
I am trying to figure out what's left to do here... if I can help in any way. I am looking to use the awskms autounseal.

Is it mostly just needed to knock out the sidecar that is currently doing the unseal? IE: https://github.com/banzaicloud/bank-vaults/blob/master/operator/pkg/controller/vault/vault_controller.go#L999-L1018

Or would it be preferred that the unseal command of bankvaults simply notice that autounseal is being used, and skip much of it's logic, perhaps mostly just doing initialization steps for new vaults?

I'd really like to replace some existing autounseal based vaults with ones managed by the operator, and am happy to jump in to help if I can get some guidance on approach.

bonifaido:
Hi @jhohertz, thank you for your offer, help is always welcomed!

The second option sounds better to me, I think we should support this with another CLI flag which would be sent down by the operator since it knows the Vault config in the first hand.

jhohertz:
[...]

PR #429, opened 2019-04-18
fix: save recovery key to storage

anasinnyk:
We need to store recovery keys when we make `vault operator init autounseal`


bonifaido:
Hi @anasinnyk, I think this fixes https://github.com/banzaicloud/bank-vaults/issues/228?

anasinnyk:
@bonifaido yes

primeroz:
Should we disable the Unseal functionality of the bank-vaults when running with auto-unseal enabled ? 

Also , when in autounseal mode, i think is a requirement that the `secret_shares` and the `threshold_shares` are set to the same ( see [here](https://www.vaultproject.io/api/system/init.html#secret_threshold) ) and the `stored_shares` are set to the same as secret shares ( see [here](https://www.vaultproject.io/api/system/init.html#stored_shares) )

bonifaido:
Yeah, if auto-unseal is enabled (we can detect this in the operator from the config file) we should disable it.

bonifaido:
We can open another issue from that since this bit can work alone I guess.

primeroz:
Yeah let's extend #228 with the rest of the auto-unseal functionalities

I am just building the image now for testing  , i am not sure we check anything specific in CI for the unseal key and ( in this case ) recovery key storage other than checking that vault is unsealed.

bonifaido:
Nothing is checked other than if Vault is unsealed from this point of view.

bonifaido:
Thanks @anasinnyk !

primeroz:
@anasinnyk even if post merge, do you have an example Configuration file to add to the `operator/deploy` directory with autounseal enabled ?

anasinnyk:
```
    vault:
      customSecrets:
      - secretName: "${var.vault_gcp_sa_secret_name}"
        mountPath: /etc/gcp
      config:
        storage:
          gcs:
            bucket: "${google_storage_bucket.vault_storage.id}"
            ha_enabled: "true"
        seal:
          gcpckms:
            project: "${module.project-factory.project_id}"
            region: "${var.gcp_region}"
            key_ring: "${google_kms_key_ring.key_ring.name}"
            crypto_key: "${google_kms_crypto_key.vault_kms_crypto_key.name}"
      unsealer:
        args:
          - "--store-root-token"
          - "--mode"
          - "google-cloud-kms-gcs"
          - "--google-cloud-kms-key-ring"
          - "${google_kms_key_ring.key_ring.name}"
          - "--google-cloud-kms-crypto-key"
          - "${google_kms_crypto_key.vault_kms_crypto_key.name}"
          - "--google-cloud-kms-location"
          - "${var.gcp_region}"
          - "--google-cloud-kms-project"
          - "${module.project-factory.project_id}"
          - "--google-cloud-storage-bucket"
          - "${google_storage_bucket.vault_storage.id}"
```
I hope it was useful for you

bonifaido:
Thanks!

Issue #437, opened 2019-04-30
Services, ConfigMap and Secrets leaks from operator clash with ResourceQuota


primeroz:
**Describe the bug**
When in a cluster _ResourceQuota_ are enabled there is clash between leaking of secrets and services and the maximum allowed number of such resources that can be created in a namespace

edit: This also happen to configmaps

**To Reproduce**
1. Create a cluster and enable resource quotas in the vault namespace, set Services and secrets to 20 ( arbitrary value)
2. Create a vault cluster using the operator
3. Watch the number of Secrets and Services grow to 20 at which point the operator will start to error out
4. Garbage Collection ( in operator sdk ) happen and clear the resources
5. Issue start again 

You reproduce the issue using [this script](https://gist.github.com/primeroz/61aeac0b70fca53b712e7a4e7f278336) which requires [KinD](https://github.com/kubernetes-sigs/kind/releases)

**Expected behavior**
Secrets and Services should be updated during Reconcile. 
Or alternatively GC should run after every Reconcile.

**Screenshots**
In a cluster with a 2 vault HA and GCS ( irrelevant ) backend
```
k describe resourcequota -n vault default | egrep "services|secrets"
secrets                 22      50
services                20      20
services.loadbalancers  0       0
services.nodeports      0       0

k describe resourcequota -n vault default | egrep "services|secrets"
secrets                 23      50
services                20      20

k get services -n vault
NAME               TYPE        CLUSTER-IP    EXTERNAL-IP   PORT(S)                               AGE                                                               
vault              ClusterIP   10.0.91.44    <none>        8200/TCP,8201/TCP,8300/TCP,9091/TCP   86m                                                               
vault-0            ClusterIP   10.0.90.90    <none>        8200/TCP,8201/TCP,8300/TCP,9091/TCP   86m                                                               
vault-1            ClusterIP   10.0.90.128   <none>        8200/TCP,8201/TCP,8300/TCP,9091/TCP   86m                                                               
vault-configurer   ClusterIP   10.0.89.249   <none>        9091/TCP                              86m        

k get secrets -n vault
NAME                         TYPE                                  DATA   AGE
default-token-kdxmj          kubernetes.io/service-account-token   3      8h
letsencrypt-staging          Opaque                                1      8h
vault-letsencrypt-cert       kubernetes.io/tls                     3      5h30m
vault-operator-token-5jfw2   kubernetes.io/service-account-token   3      6h23m
vault-tls                    Opaque                                3      177m
[...]

PR #438, opened 2019-04-30
operator: fix resource quota issues via checking and upgrading existing objects

bonifaido:
Fixes: #437 

The operator Helm charts has to be updated as well to add RBAC extensions.

### DIFF
--- a/operator/pkg/controller/vault/vault_controller.go
+++ b/operator/pkg/controller/vault/vault_controller.go
@@ -45,7 +45,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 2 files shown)


operator/pkg/controller/vault/vault_controller.go
```diff
@@ -98,6 +97,52 @@ type ReconcileVault struct {
...
+	if err != nil {
+		return err
+	}
+
+	clone := o.DeepCopyObject()
+
+	err = r.client.Get(context.TODO(), key, clone)
+	if apierrors.IsNotFound(err) {
+		err = r.client.Create(context.TODO(), o)
+		if !apierrors.IsAlreadyExists(err) {
```
primeroz: Does this need to be here ?

We just checked that the error was "IsNotFound" so i guess  it is impossible that on Create we get an error of "Already Exists"

operator/pkg/controller/vault/vault_controller.go
```diff
@@ -98,6 +97,52 @@ type ReconcileVault struct {
...
```
primeroz: same as above.

operator/pkg/controller/vault/vault_controller.go
```diff
@@ -268,9 +283,9 @@ func (r *ReconcileVault) Reconcile(request reconcile.Request) (reconcile.Result,
 	if err := controllerutil.SetControllerReference(v, ser, r.scheme); err != nil {
 		return reconcile.Result{}, err
 	}
-	err = r.client.Create(context.TODO(), ser)
-	if err != nil && !apierrors.IsAlreadyExists(err) {
-		return reconcile.Result{}, fmt.Errorf("failed to create service: %v", err)
+	err = r.createObjectIfNotExists(ser)
```
primeroz: Should the service be a `createOrUpdateObject` ? 

that would help to correct drifts and i don't see any issue with updating it since the spec for it should always be the same

operator/pkg/controller/vault/vault_controller.go
```diff
@@ -282,9 +297,9 @@ func (r *ReconcileVault) Reconcile(request reconcile.Request) (reconcile.Result,
 		if err := controllerutil.SetControllerReference(v, ser, r.scheme); err != nil {
 			return reconcile.Result{}, err
 		}
-		err = r.client.Create(context.TODO(), ser)
-		if err != nil && !apierrors.IsAlreadyExists(err) {
-			return reconcile.Result{}, fmt.Errorf("failed to create per instance service: %v", err)
+		err = r.createObjectIfNotExists(ser)
```
primeroz: same as above

operator/pkg/controller/vault/vault_controller.go
```diff
@@ -337,23 +338,21 @@ func (r *ReconcileVault) Reconcile(request reconcile.Request) (reconcile.Result,
...
-	if err != nil && apierrors.IsAlreadyExists(err) {
-		err = r.client.Update(context.TODO(), configurerDep)
-	}
+	err = r.createOrUpdateObject(configurerDep)
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("failed to create/update configurer deployment: %v", err)
 	}
 
-	// Create the Configurer service if it doesn't exist
+	// Create the Configurer ice if it doesn't exist
```
primeroz: typo : service

operator/pkg/controller/vault/vault_controller.go
```diff
@@ -337,23 +338,21 @@ func (r *ReconcileVault) Reconcile(request reconcile.Request) (reconcile.Result,
...
 	configurerSer := serviceForVaultConfigurer(v)
 	// Set Vault instance as the owner and controller
 	if err := controllerutil.SetControllerReference(v, configurerSer, r.scheme); err != nil {
 		return reconcile.Result{}, err
 	}
-	err = r.client.Create(context.TODO(), configurerSer)
-	if err != nil && !apierrors.IsAlreadyExists(err) {
-		return reconcile.Result{}, fmt.Errorf("failed to create service: %v", err)
+
+	err = r.createObjectIfNotExists(configurerSer)
```
primeroz: same as above. Should service be _creatOrUpdate_ ?

primeroz:
> I will run some tests on the image using my KinD test script and will report on the leaking

No leaking happening.

operator/pkg/controller/vault/vault_controller.go
bonifaido: Thanks, updated.

operator/pkg/controller/vault/vault_controller.go
bonifaido: Thanks fixed, ClusterIP was tricky here, but I fixed it.

bonifaido:
Thanks for testing this @primeroz

PR #440, opened 2019-04-30
Enable Controller prometheus metrics

primeroz:
Prometheus metrics for the Controller are now available out of the box, as described [here](https://github.com/operator-framework/operator-sdk/tree/master/doc/user/metrics)

exposed (useful) metrics are
```
controller_runtime_reconcile_queue_length{controller="vault-controller"} 0
controller_runtime_reconcile_time_seconds_bucket{controller="vault-controller",le="0.005"} 0
controller_runtime_reconcile_time_seconds_bucket{controller="vault-controller",le="0.01"} 0
controller_runtime_reconcile_time_seconds_bucket{controller="vault-controller",le="0.025"} 0
controller_runtime_reconcile_time_seconds_bucket{controller="vault-controller",le="0.05"} 0
controller_runtime_reconcile_time_seconds_bucket{controller="vault-controller",le="0.1"} 0
controller_runtime_reconcile_time_seconds_bucket{controller="vault-controller",le="0.25"} 0
controller_runtime_reconcile_time_seconds_bucket{controller="vault-controller",le="0.5"} 0
controller_runtime_reconcile_time_seconds_bucket{controller="vault-controller",le="1"} 0
controller_runtime_reconcile_time_seconds_bucket{controller="vault-controller",le="2.5"} 0
controller_runtime_reconcile_time_seconds_bucket{controller="vault-controller",le="5"} 1
controller_runtime_reconcile_time_seconds_bucket{controller="vault-controller",le="10"} 1
controller_runtime_reconcile_time_seconds_bucket{controller="vault-controller",le="+Inf"} 3
controller_runtime_reconcile_time_seconds_sum{controller="vault-controller"} 32.514783989
controller_runtime_reconcile_time_seconds_count{controller="vault-controller"} 3
controller_runtime_reconcile_total{controller="vault-controller",result="success"} 3
```

Unfortunately though if we want the metrics to be enabled by default this is a **breaking change** since to expose the service using `_, err = metrics.ExposeMetricsPort(context.TODO(), metricsPort)` the operator need to be able to `GET replicasets` and the current RBAC does not allow for that.

This is now added to the example operator-rbac but if anyone was going to update to a version of the operator with this feature enabled it would break on start with a forbidden get request for replicasets.

Alternatively i could use an Environment variable to enable the feature and have it set to false by default. That would be backward compatible but it would also break the rule of having **sane defaults**

### DIFF
--- a/operator/cmd/manager/main.go
+++ b/operator/cmd/manager/main.go
@@ -27,6 +27,7 @@ import (
 	"github.com/banzaicloud/bank-vaults/operator/pkg/controller"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
+	"github.com/operator-framework/operator-sdk/pkg/metrics"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
@@ -38,8 +39,10 @@ import (
 var log = logf.Log.WithName("cmd")
 
 const (
-	operatorNamespace = "OPERATOR_NAMESPACE"
-	livenessPort      = "8080"
+	operatorNamespace       = "OPERATOR_NAMESPACE"
+	livenessPort            = "8080"
+	metricsHost             = "0.0.0.0"
+	metricsPort       int32 = 8383
 )
 
 func printVersion() {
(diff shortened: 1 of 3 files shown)


primeroz:
@bonifaido  Yeah i will update the helm chart.

What should I do about the breaking changes though ? 
Are you happy to have it as it is ? And maybe just add a note in the release notes ? Or should we make it an Environment Variable flag to the operator ?

bonifaido:
I think the release note should be fine, and we should bump one minor version higher (since we are pre 1.0.0).

PR #446, opened 2019-05-07
PSP AllowPrivilegeEscalation

moskitone:
Allow to enable/disable AllowPrivilegeEscalation for injected containers to make it works with psp restricted deployments

Caused by error in replicaset
**is forbidden: unable to validate against any pod security policy: []'**

### DIFF
--- a/cmd/vault-secrets-webhook/main.go
+++ b/cmd/vault-secrets-webhook/main.go
@@ -327,7 +327,7 @@ func parseVaultConfig(obj metav1.Object) vaultConfig {
 
 	if val, ok := annotations["vault.security.banzaicloud.io/psp-allow-privilege-escalation"]; ok {
 		vaultConfig.pspAllowPrivilegeEscalation, _ = strconv.ParseBool(val)
-	} else if viper.GetString("psp_allow_privilege_escalation") != "" {
+	} else {
 		vaultConfig.pspAllowPrivilegeEscalation, _ = strconv.ParseBool(viper.GetString("psp_allow_privilege_escalation"))
 	}
 
--- a/cmd/vault-secrets-webhook/main.go
+++ b/cmd/vault-secrets-webhook/main.go
@@ -36,16 +36,17 @@ import (
 )
 
 type vaultConfig struct {
-	addr                  string
-	role                  string
-	path                  string
-	skipVerify            string
-	tlsSecret             string
-	useAgent              bool
-	ctConfigMap           string
-	ctOnce                bool
-	ctShareProcess        bool
-	ctShareProcessDefault string
+	addr                        string
+	role                        string
+	path                        string
+	skipVerify                  string
+	tlsSecret                   string
+	useAgent                    bool
+	ctConfigMap                 string
+	ctOnce                      bool
+	ctShareProcess              bool
+	ctShareProcessDefault       string
+	pspAllowPrivilegeEscalation bool
 }
 
 var vaultAgentConfig = `
@@ -90,7 +91,8 @@ func getInitContainers(originalContainers []corev1.Container, vaultConfig vaultC
 
 		runAsUser := int64(100)
 		securityContext := &corev1.SecurityContext{
-			RunAsUser: &runAsUser,
+			RunAsUser:                &runAsUser,
+			AllowPrivilegeEscalation: &vaultConfig.pspAllowPrivilegeEscalation,
 		}
 
 		containers = append(containers, corev1.Container{
@@ -116,6 +118,9 @@ func getInitContainers(originalContainers []corev1.Container, vaultConfig vaultC
 					MountPath: "/vault/",
 				},
 			},
+			SecurityContext: &corev1.SecurityContext{
+				AllowPrivilegeEscalation: &vaultConfig.pspAllowPrivilegeEscalation,
+			},
 		})
 	}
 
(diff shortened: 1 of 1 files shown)


pbalogh-sa:
@moskitone , It seems you haven't any PSP or the used ServiceAccout cannot use the defined PSP.   Do you have any PSP defined in you cluster? Does the ServiceAccount of created pods have access to defined PSP? How did you deploy vault-secrets-webhook?

anasinnyk:
@pbalogh-sa we deploy vault-secrets-webhook via helm chart from the banzai-charts repo. We have a service account with psp usage role binding. And we have `allowPrivilegeEscalation` set to `false` in this psp. But when admission mutation webhook add containers to pod, it have `securityContext.allowPrivilegeEscalation = true` by default. And our psp cannot be applied to this pod.

primeroz:
I don't have any log specific to privilege escalation with me but when i get a deployment forbidden by a PSP ( so in the case of a conflict for the _AllowPrivilegeEscalation_ ) i get the error inside the square brackets . 

For example: `message:  "Error creating: pods "prometheus-to-sd-" is forbidden: unable to validate against any pod security policy: [spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used]"   ` 

Your error does look like no PSP was allowed to be used by the ServiceAccount in that namespace.

cmd/vault-secrets-webhook/main.go
```diff
@@ -317,6 +325,12 @@ func parseVaultConfig(obj metav1.Object) vaultConfig {
 		vaultConfig.ctShareProcess = false
 	}
 
+	if val, ok := annotations["vault.security.banzaicloud.io/psp-allow-privilege-escalation"]; ok {
+		vaultConfig.pspAllowPrivilegeEscalation, _ = strconv.ParseBool(val)
+	} else {
```
primeroz: How are we handling default value for this new flag ? 

Should it be 
```
if annotation
else if viper.getString
else false 
```
?

moskitone: strconv.ParseBool returns false in case psp_allow_privilege_escalation is not defined

primeroz: Good to know! 
so i guess this last change "https://github.com/banzaicloud/bank-vaults/pull/446/commits/15df96348939002081043758aaed4c7fb1d28021" was not needed ? 

we still don't handle the case where the flag is missing, right ?

anasinnyk:
@primeroz I debug it a little bit more. And it's really strange :) I think we have some bug in k8s with psp plugin. 
If our mutation webhook add container without security context we have error: 
`is forbidden: unable to validate against any pod security policy: []`
if we set security context allowPrivilegeEscalation to true in our webhook code we see next error:
`is forbidden: unable to validate against any pod security policy: [spec.initContainers[0].securityContext.allowPrivilegeEscalation: Invalid value: true: Allowing privilege escalation for containers is not allowed spec.initContainers[0].securityContext.allowPrivilegeEscalation: Invalid value: true: Allowing privilege escalation for containers is not allowed]`
if we set it to false our pod created without error. You can check it in our test repository if you need it anasinnyk/webhook-admission-mutation

primeroz:
@anasinnyk 
It  is strange indeed, but i guess that is what PSP does, makes things strange! :)

thinking about it the "allowPrivilegeEscalation" should only control if _setuid_ binaries will work as expected inside the container but in most cases, unless a direct conflict exists like when you set it to true explicitly in your pod security context, it should not prevent the Pod from starting. 
This was my case for example with Vault and mlock ( without allowing privilege escalation it would fail to start and lock memory) 

As a test, are you able to add 

```
DefaultAllowPrivilegeEscalation - Sets the default for the allowPrivilegeEscalation option. The default behavior without this is to allow privilege escalation so as to not break setuid binaries. If that behavior is not desired, this field can be used to default to disallow, while still permitting pods to request allowPrivilegeEscalation explicitly.
```
as specified [here](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#privilege-escalation) 
 to your psp ?

cmd/vault-secrets-webhook/main.go
moskitone: so i guess this last change "15df963" was not needed ? - yes,  I've added new commit.

this if-else statement copypasted from annotations["vault.security.banzaicloud.io/vault-agent"]. If annotation and flag are not defined there will be default "false" value

moskitone:
@primeroz 
This is our spec.
```
spec:
  allowPrivilegeEscalation: false
  allowedCapabilities:
  - '*'
  fsGroup:
    rule: RunAsAny
  runAsUser:
    rule: RunAsAny
  seLinux:
    rule: RunAsAny
  supplementalGroups:
    rule: RunAsAny
  volumes:
  - '*'
```

By the way
* spec.defaultAllowPrivilegeEscalation: Invalid value: true: Cannot set DefaultAllowPrivilegeEscalation to true without also setting AllowPrivilegeEscalation to true

anasinnyk:
@primeroz I added `DefaultAllowPrivilegeEscalation` to `false` too, but it's still forbidden with same empty errors

primeroz:
@anasinnyk i will try and replicate the issue but so far i failed at it 

see [here](https://gist.github.com/primeroz/c68a5fdb1a81e43ba0bbaf5ff79b8a14)

I created a KIND Cluster with psp enabled. 

In the default namespace i used your PSP policy and created the deployment from "pod.yaml" and it worked fine 

the deployment in the pod.yaml has securitycontext set with runasuser 

I know this is not the webhook ( and we could jsut extend the example to run the webhook but i had no time to expand the helm chart for it )

as expected though the pod started even with the psp enforcing a value of false for "allowprivileged"

anasinnyk:
@primeroz It's a bug related to mutation webhook. I have a repo for test and you don't need helm for that
https://github.com/anasinnyk/webhook-admission-mutation/tree/master/k8s 
- deploymnt.yaml contain webhook
- service.yaml svc for webhook
- admission - it's a mutationwebhookconfiguration
- test-app.yaml - it's a simple deployment when we add initContaiener without secutiryContext (empty error case) 

PS add annotation "webhook.k8s.macpaw.io/dissable-allow-privilege-escalation = true" and it's add `secutiryContext.allowPrivilegeEscalation=false` to init container. And it's a success case.

primeroz:
I see. 

I LL give a try tomorrow but anyway having a flag to specify the field can't harm as long as we have a default.

I just like to understand things so I will try your example when I am back at a computer

Cheers.

moskitone:
@pbalogh-sa Thanks!

PR #447, opened 2019-05-07
Add support for provisioning vault with auto-unsealing

### LINKED ISSUE #228: Support Vault auto-unseal based deployments

ccojocar:
fixes #228


cmd/bank-vaults/init.go
```diff
@@ -36,41 +41,80 @@ It will not unseal the Vault instance after initialising.`,
...
+	if err != nil {
+		logrus.Fatalf("error connecting to vault: %s", err.Error())
+	}
+
+	vaultConfig, err := vaultConfigForConfig(appConfig)
+	if err != nil {
+		logrus.Fatalf("error building vault config: %s", err.Error())
+	}
+
+	v, err := vault.New(nil, cl, vaultConfig)
```
primeroz: i think that without a store we can't "store" the recovery keys and root token in the backend.

we need those since

* configurer will use the root token to configure vault
* recovery keys need to be available to _human_ operator to recover vault master key in case of KMS key destruction or other disaster recovery scenarios

operator/pkg/controller/vault/vault_controller.go
```diff
@@ -917,6 +917,10 @@ func statefulSetForVault(v *vaultv1alpha1.Vault) (*appsv1.StatefulSet, error) {
 		return nil, err
 	}
 
+	unsealCommand := []string{"bank-vaults", "unseal", "--init"}
+	if v.Spec.IsAutoUnseal() {
+		unsealCommand = []string{"bank-vaults", "init", "--auto-unseal"}
```
primeroz: when we run in _unseal_ mode with the _--init_ option we start the prometheus exporter but when we run in _init_ mode i don't think we do. 

See https://github.com/banzaicloud/bank-vaults/blob/master/cmd/bank-vaults/unseal.go#L84

If we go down this route of just running the _init_ mode we will need to start the prometheus exporter there as well.

pkg/vault/operator_client.go
```diff
@@ -189,6 +190,30 @@ func (v *vault) keyStoreSet(key string, val []byte) error {
...
+	_, err = v.cl.Sys().Init(&api.InitRequest{
+		SecretShares:      v.config.SecretShares,
+		SecretThreshold:   v.config.SecretThreshold,
+		RecoveryShares:    v.config.SecretShares,
+		RecoveryThreshold: v.config.SecretThreshold,
+	})
+	if err != nil {
+		return fmt.Errorf("error initializing vault: %s", err.Error())
+	}
+	return nil
```
primeroz: See above comment, we need to store the Recovery keys the same way we do in the init mode.

https://github.com/banzaicloud/bank-vaults/blob/master/pkg/vault/operator_client.go#L256

Also, if we do split the Init functions at this level we should probably remove the recovery keys from the init function since it will never be used there and only have it here

pkg/vault/operator_client.go
```diff
@@ -189,6 +190,30 @@ func (v *vault) keyStoreSet(key string, val []byte) error {
 	}
 }
 
+// InitAutoUnseal initializes Vault in auto-unseal mode
+func (v *vault) InitAutoUnseal() error {
```
primeroz: Do we actually need to split the Init function ? 

The current INIT function works well in both _auto unseal_ and _bank-vaults unseal_ mode. 

It also already manage to store the _secret keys_ or the _recovery keys_ in the backend based on the mode used since that part is all handled by vault which won't return the secret keys when in _auto unseal_ and it won't return the recovery keys when in _bank vaults_ mode

cmd/bank-vaults/init.go
```diff
@@ -36,41 +41,80 @@ It will not unseal the Vault instance after initialising.`,
...
+		logrus.Fatalf("error creating vault helper: %s", err.Error())
+	}
+
+	if err = v.InitAutoUnseal(); err != nil {
+		logrus.Fatalf("error initialising vault: %s", err.Error())
+	}
+
+	initPeriod := cfg.GetDuration(cfgInitPeriod)
+	for {
+		if err = v.InitAutoUnseal(); err != nil {
```
primeroz: I wonder if this logic should actually be in the _unseal command_ for loop and keep the init command as a one off as it is now.

primeroz:
@ccojocar thinking about it a bit more about it i am convinced we can do this much simpler by just using the 

"v.Spec.IsAutoUnsealing" check in the _unseal_ mode for loop to decide if we need to unseal or we just sleep.

Can we give that implementation a try ?

bonifaido:
I definitely agree with @primeroz here, as I have mentioned [in the cross issue](https://github.com/banzaicloud/bank-vaults/issues/228#issuecomment-490219000).

ccojocar:
I reworked the PR to store the shards when the vault is initialised and also fixed the secrets engine config serialisation. 

I managed to provision latest vault version with auto-unsealing, and also the kv secrets engine gets mounted by the configurator. See the attached logs. 
[vault-autounseal.log](https://github.com/banzaicloud/bank-vaults/files/3156634/vault-autounseal.log)

ccojocar:
> Should the logic of sleeping rather than unsealing be part of the "unseal" command rather than the "init" command ? currently the init command is a one off run of bank-vaults while the unseal command is the daemon mode. should we just handle the decision of performing unsealing or sleep in the foor loop at https://github.com/banzaicloud/bank-vaults/blob/master/cmd/bank-vaults/unseal.go#L94 with v.Spec.IsAutoUnseal ? see comment below

I fine with this we can also moved in the seal command. Right now, I added an option to init command `auto-unseal`, in which case the init runs in a loop to avoid having the side-container exiting.

cmd/bank-vaults/init.go
ccojocar: I refactored it to use the same init method.

pkg/vault/operator_client.go
ccojocar: Done

cmd/bank-vaults/init.go
ccojocar: It is removed now.

pkg/vault/operator_client.go
ccojocar: This is removed.

operator/pkg/controller/vault/vault_controller.go
ccojocar: I can do this in the init command.

ccojocar:
@primeroz @bonifaido I will add an option to `unseal` command e.g. `auto` in which case the for look does not execute any unsealing. Does this sound good to you?

bonifaido:
@ccojocar that sounds feasible, thank you for the updates!

operator/pkg/controller/vault/vault_controller.go
ccojocar: I moved the initialisation into `unseal` command.

ccojocar:
@bonifaido @primeroz The PR is ready for review now after rework. Thanks!

primeroz:
LGTM now, just waiting on CI

PR #448, opened 2019-05-07
[vault-env] Add support for versioning

tczekajlo:
This PR is adding support for versioning. You can choose a version of the secret, e.g. we wanna use version 2.

```
TEST="vault:secret/data/accounts/aws#AWS_SECRET_ACCESS_KEY#2"
```

If we don't explicitly declare version then will be used the latest one.


cmd/vault-env/main.go
```diff
@@ -124,6 +129,18 @@ func main() {
 			} else {
 				var data map[string]interface{}
 				v2Data, ok := secret.Data["data"]
+
+				// Check if a given version of a path is destroyed
+				metadata := secret.Data["metadata"].(map[string]interface{})
```
bonifaido: Correct me if I'm wrong but I think this will break in case of KV 1 backends.

tczekajlo: My bad. I've moved checks to the proper place. Now It works with KV1 and KV2 backends. In case of KV1 version of a secret is ignored.

bonifaido: Great thanks! :)

bonifaido: Could you please sign the CLA in the meantime?

tczekajlo:
You're welcome :)

Issue #450, opened 2019-05-08
[webhook]  vault-env is not injected when vault.security.banzaicloud.io/vault-agent: "false"


moskitone:
Hello 

**Describe the bug**
It seems that logic in workflow is broken after 0.4.11.
When vault.security.banzaicloud.io/vault-agent is set to "false" container copy-vault-env is not injected to pod.

```
apiVersion: v1
items:
- apiVersion: v1
  kind: Pod
  metadata:
    annotations:
      cni.projectcalico.org/podIP: 10.250.220.132/32
      kubernetes.io/psp: 100-default-psp
      vault.security.banzaicloud.io/vault-addr: https://vault8.test.io
      vault.security.banzaicloud.io/vault-agent: "false"
      vault.security.banzaicloud.io/vault-path: kubernetes
      vault.security.banzaicloud.io/vault-role: k8s-test
      vault.security.banzaicloud.io/vault-skip-verify: "true"
    creationTimestamp: 2019-05-08T09:08:43Z
    generateName: whoami-77b47c95d9-
    labels:
      pod-template-hash: 77b47c95d9
      run: whoami
    name: whoami-77b47c95d9-6p52h
    namespace: mo5
    ownerReferences:
    - apiVersion: apps/v1
      blockOwnerDeletion: true
      controller: true
      kind: ReplicaSet
      name: whoami-77b47c95d9
      uid: e0ce0a88-7170-11e9-9cfd-42010afae40a
    resourceVersion: "2549992"
    selfLink: /api/v1/namespaces/mo5/pods/whoami-77b47c95d9-6p52h
    uid: e12a1dfc-7170-11e9-9cfd-42010afae40a
  spec:
    containers:
    - command:
      - /vault/vault-env
      env:
      - name: AWS_SECRET_ACCESS_KEY
        value: vault:secret/data/k8s/test#foo
      - name: VAULT_ADDR
        value: https://vault8.test.io
      - name: VAULT_SKIP_VERIFY
        value: "true"
      - name: VAULT_PATH
        value: kubernetes_devops
      - name: VAULT_ROLE
        value: k8s-test
      image: containous/whoami
      imagePullPolicy: Always
      name: whoami-empty
      resources: {}
      securityContext:
        allowPrivilegeEscalation: false
        procMount: Default
      terminationMessagePath: /dev/termination-log
      terminationMessagePolicy: File
      volumeMounts:
      - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
[...]

PR #451, opened 2019-05-08
fix for [webhook] vault-env is not injected when vault.security.banzaicloud.io/vault-agent: "false"

moskitone:


moskitone:
fix for https://github.com/banzaicloud/bank-vaults/issues/450

PR #457, opened 2019-05-14
configurer: add support for token auth

bonifaido:
Currently, it is not possible to configure roles for token-based authentication, this PR fixes that. One of the possible use cases is to configure a narrow role for scraping Prometheus metrics.


bonifaido:
just fixed this :)

PR #460, opened 2019-05-15
explicitly define the config file in the volume mount so the file is i…

jurgenweber:
…n the container, Here is the error I was getting in the configurer consol logs:

```
time="2019-05-14T23:56:41Z" level=fatal msg="error executing vault config template: template: no template \"/config/vault-configurer/vault-config.yml\" associated with template \"vault-config-file\""
```


jurgenweber:
looks like a local issue.

PR #465, opened 2019-05-15
Extend Annotations support for operator 

### LINKED ISSUE #357: [Operator] Support annotations for all operator created resources
Currently, only the Etcd Ingress and Pod resources are configurable with annotations. The TLS Secrets and ConfigMaps and Services should be also annotable.

primeroz:
fixes #357 

* Split Annotations between
  * common ( Current annotations field in the spec )
  * vaultAnnotations
  * vaultConfigurerAnnotations
* Set prometheus annotations only on the pod templates
* Set annotations on Deployment and Statefulset ( needed by my use case )
* Annotate vault and vault-configurer Services
* Annotate vault-tls secret

* Fix port for prometheus-exporter in vault-service ( unrelated ) 

TODO:
* EtcdAnnotations propagation to Pods 

@bonifaido @pbalogh-sa @matyix This is a Draft pull request due the missing etcd annotations ( will do tomorrow ) and to make sure you are happy with the direction this implementation is getting. 

I mostly need this to support the External Reloader for vault when cert-manager letsencrypt tls is updated

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -8,4 +8,5 @@
 - [Unseal Keys](unseal-keys/README.md)
 - [Getting and Installing](installing/README.md)
 - [Monitoring](monitoring/README.md)
+- [Annotations](annotations/README.md)
 - [Contributing](contributing/README.md)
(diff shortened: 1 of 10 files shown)


bonifaido:
LGTM, I like the implementation, I will wait for the etcd annotations, thank you!

primeroz:
Docs added to describe Annotations

PR #467, opened 2019-05-16
Fix mount config input serialisation

ccojocar:
There is no need to convert the config keys to string. This will lead to decoding errors.

```time="2019-05-16T08:01:17Z" level=error msg="error configuring vault: error configuring secret engines for vault: error parsing config for secret engine: 2 error(s) decoding:\n\n* 'force_no_cache' expected type 'bool', got unconvertible type 'string'\n* 'options' expected a map, got 'string'"```

### DIFF
--- a/pkg/vault/operator_client.go
+++ b/pkg/vault/operator_client.go
@@ -1309,9 +1309,9 @@ func isOverwriteProhibitedError(err error) bool {
 
 func getMountConfigInput(secretEngine map[string]interface{}) (api.MountConfigInput, error) {
 	var mountConfigInput api.MountConfigInput
-	config, err := getOrDefaultStringMapString(secretEngine, "config")
-	if err != nil {
-		return mountConfigInput, fmt.Errorf("error getting config for secret engine: %v", config)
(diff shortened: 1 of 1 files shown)


ccojocar:
@bonifaido I fixed this a while back but it seems that was changed recently. This prevents us to mount the secrets engine when provisioning vault.

ccojocar:
We use the default MountConfigInput https://github.com/hashicorp/vault/blob/b9f054d8abe0f5a6993bbfd13fb8e7e582df42e3/api/sys_mounts.go#L144. I think the type of the fields should be respected.

ccojocar:
The conversion tries to convert every field to a string but they are not flatten. Also they are already serialised.

bonifaido:
We agree that we this needs to be fixed but making it mandatory breaks user configurations, so let's fix it that way.

ccojocar:
I think, instead of returning an error we can return the default if config is not found.

ccojocar:
@bonifaido I pushed a fix which returns the default value instead of an error if the config is not found for a secret engine.

PR #472, opened 2019-05-20
Read entrypoint and cmd from image configuration

tczekajlo:
In a case when we wanna use entrypoint and cmd of an image we can download that information from registry via API and use it as args for vault-env.

If arguments won't be given for vault-env, the vault-env will try to get entrypoint and cmd from image configuration via registry API. In case when arguments are declared, the vault-env acts as usual.

For docker hub we have to declare only `CONTAINER_IMAGE` environment variable, e.g `CONTAINER_IMAGE=tczekajlo/demo:latest`.

For private registry, we have to declare additional env variables.
`REGISTRY_ADDRESS="https://my.registry"`
`REGISTRY_SECRET_PATH="secret/data/registry"` - vault-env will be looking for credentials to the registry in the given path. The path should contain `USERNAME` and `PASSWORD` keys.
`CONTAINER_IMAGE_TRIM="my.registry/"` - trim registry address from image name. 
`REGISTRY_SKIP_VERIFY="true"` - skip SSL verification. 

Fixes: #475


go.mod
```diff
@@ -39,6 +41,7 @@ require (
 	github.com/go-logr/zapr v0.1.0 // indirect
 	github.com/go-openapi/spec v0.19.0 // indirect
 	github.com/go-sql-driver/mysql v1.4.1 // indirect
+	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
```
bonifaido: Instead of glog please use some already used logging library that is available in the project zap in the operator and logrus in the other parts, I'm okay with either of those.

cmd/vault-env/utils.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		fmt.Print(err)
+	}
+
+	reader, err := hub.DownloadBlob(imageName, manifest.Config.Digest)
+	if reader != nil {
+		defer reader.Close()
+	}
+	if err != nil {
+		fmt.Print(err)
```
bonifaido: I think basically all `fmt.Print(err)` should be replaced with some `.Fatal()` call.

cmd/vault-env/entrypoint_cmd.go
```diff
@@ -0,0 +1,77 @@
+package main
```
bonifaido: Could you please add the standard header comments to the files, this is required by the build currently (and thus fails): https://github.com/banzaicloud/bank-vaults/pull/472/files#diff-203b8812e7300d02adf4c89070655c8eL1

cmd/vault-env/entrypoint_cmd.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/banzaicloud/bank-vaults/pkg/vault"
+	"github.com/spf13/cast"
+)
+
+// GetEntrypointCmd returns entrypoint and command of container
+func GetEntrypointCmd(vaultClient *vault.Client) ([]string, []string) {
+	registryUsername := ""
+	registryPassword := ""
+
+	registryUsernameKey := "USERNAME"
```
bonifaido: I think these USERNAME/PASSWORD env var names are too general and could collide with application-level variables.

tczekajlo:
@bonifaido Thanks for comments. Actually, you're right, all the information what is required is already available in POD definition. 

I'll try to prepare changes to reads information from POD definition in this week :)

bonifaido:
Thank you, looking forward to the changes, anyhow great work! 👍

timduhenchanter:
Awesome @bonifaido and @tczekajlo I think a caching mechanism would be really helpful for this as well to look against before reaching out.  

Also, depending on how much free time you guys have *wink wink nudge nudge* there are "enterprise" scenarios where the Docker Registry API is wrapped and not exposed I.E. Kubernetes which is the reason for this approach and JFrog Artifactory, etc... so in the future have the potential to support mainstream private registries.

cmd/vault-env/entrypoint_cmd.go
tczekajlo: Added

tczekajlo:
@bonifaido I've added changes. Now all needed information is read from POD definition. 

Environment variables are required:
`MY_POD_NAME` - value contain pod name
`NAMESPACE` - namespace
`CONTAINER_NAME` - optional, require in a case when there are more than once container in POD.

@timduhenchanter I can't see how caching mechanism would help, can you tell something more about your idea? :)

cmd/vault-env/entrypoint_cmd.go
bonifaido: Yes, please add it to all newly introduced files, as seen in: https://circleci.com/gh/banzaicloud/bank-vaults/1452?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

bonifaido: Anyhow, except that, I'm happy to merge :)

tczekajlo: Ah, I've missed a dot, now should be better :)

timduhenchanter:
@tczekajlo The caching idea is to prevent from `vault-env` potentially slamming a registry API in a large scale environment as it reaches out for the metadata every time a pod is initialized.  So it can look against something locally to the cluster prior to reaching out.  I will expand on this when I have some time and associate it with a future request at some point.  

Thank you for your work on this!

MXfive:
I think implementing this in the webhook controller is a cleaner way to go about this.  Not only would it make caching easier, it also does not unnecessarily give all containers read access to the docker repository.  Docker image / source code theft could be possible because of this unnecessary access.

bonifaido:
Hi @MXfive, there will be some rework on this before releasing this, since we identified a few issues regarding the current architecture, so please stay tuned, these issues are going to be treated next week.

PR #473, opened 2019-05-21
allow for annotations overrides of the consul template image

jurgenweber:
I have run into an edge case where the deployment may have security contexts set and you will need to override the stuff that is hard coded on a case by case basis.

It will require creating a custom image and setting securitycontext runAsUser UID, depending on your preferred approach.

The initContainer podSpec and Container podSpec I think are different so only touching the Container one.

The UID in the hashicorp container is 100:1000 for the user/group `consul-template`


jurgenweber:
also add a pull policy

PR #476, opened 2019-05-24
Add documentation for ct and prometheus

jurgenweber:
Docs


jurgenweber:
glad oyu like it, I have one more I want to do. I will do it this week.

Issue #355, opened 2019-03-11
[webhook] parse image metadata to find out command and args


bonifaido:
https://banzaicloud.com/blog/inject-secrets-into-pods-vault/#comment-4359596025

> A possible fix would be to analyze the image with the Docker API/ContainerD API and parse the implicit command from the metadata backed into the image.

Issue #227, opened 2018-12-11
Clarify / Support default entrypoint with vault webhook


captncraig:
The readme says this about the mutating webhook:

> The command of the container has to be explicitly defined in the resource definition, the container's default ENTRYPOINT and CMD will not work (to overcome this is a work-in-progress).

A brief look at the code, and I really can't say why this would be the case, but I believe you that it is. 

Is there a better explanation to why adding an init container changes the entrypoint of existing primary containers? Is there a milestone or any progress toward working around that limitation? Thanks!

bonifaido:
Hi Craig, this is because the webhook is not just inserting the init-container but also rewrites (or adds if there isn't any) `command` and `args` parameters in the `PodSpec`, directly from the [blog](https://banzaicloud.com/blog/inject-secrets-into-pods-vault/):

> The init-container also changes the command of the container to run vault-env instead of the application directly. vault-env starts up, connects to Vault (using the Kubernetes Auth method, checks that the environment variables have a reference to a value stored in Vault (vault:secret/....) and replaces that with the value read from Vault’s Secret backend. After that, vault-env executes (with syscall.Exec()) the original process, which can use the secret that was originally stored in Vault.

So since we overwrite the "implicit" entrypoint baked in directly to the Docker image of your application, it is required to make it "explicit". I hope this clarifies your question.

bonifaido:
@captncraig Let me know if you have any questions (reopen the issue).

PR #479, opened 2019-05-28
[vault-secret-injector] Pull default docker image and use AWS creds t…

arturo-c:
…o pull if specified.

This PR can be marked as WIP, it works for the specific implementation that we need, but mostly want to get input from maintainers if they think this is a good route.

I can rework this a bit to be more generic in terms of accepting authentication and clean up the code, write some tests possibly.

Partly addresses https://github.com/banzaicloud/bank-vaults/issues/355, https://github.com/banzaicloud/bank-vaults/issues/227

### DIFF
--- a/cmd/vault-secrets-webhook/main.go
+++ b/cmd/vault-secrets-webhook/main.go
@@ -529,6 +529,9 @@ func mutateContainers(containers []corev1.Container, vaultConfig vaultConfig, ns
 					panic(err)
 				}
 				data, err := base64.StdEncoding.DecodeString(*resp.AuthorizationData[0].AuthorizationToken)
+				if err != nil {
+					panic(err)
+				}
 				token := strings.SplitN(string(data), ":", 2)
 				authConfig := types.AuthConfig{
 					Username: token[0],
--- a/cmd/vault-secrets-webhook/main.go
+++ b/cmd/vault-secrets-webhook/main.go
@@ -16,17 +16,24 @@ package main
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"os"
 	"strconv"
 	"strings"
 
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ecr"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
 	whhttp "github.com/slok/kubewebhook/pkg/http"
 	"github.com/slok/kubewebhook/pkg/log"
 	whcontext "github.com/slok/kubewebhook/pkg/webhook/context"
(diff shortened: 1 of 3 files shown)


cmd/vault-secrets-webhook/main.go
```diff
@@ -503,7 +510,55 @@ func mutateContainers(containers []corev1.Container, vaultConfig vaultConfig, ns
...
+		args := append(container.Command)
+		if len(args) == 0 {
+			// Pull docker image and inspect
+			ctx := xcontext.Background()
+			cli, err := client.NewEnvClient()
+			if err != nil {
+				panic(err)
+			}
+			options := types.ImagePullOptions{}
+			if viper.IsSet("aws_access_key_id") && strings.Contains(container.Image, "amazonaws.com") {
```
arturo-c: This can probably be moved to another package, and use an interface that implements an Auth method

cmd/vault-secrets-webhook/main.go
```diff
@@ -503,7 +510,55 @@ func mutateContainers(containers []corev1.Container, vaultConfig vaultConfig, ns
...
+			_, err = cli.ImagePull(ctx, container.Image, options)
+			if err != nil {
+				panic(err)
+			}
+			imageInspect, _, err := cli.ImageInspectWithRaw(ctx, container.Image)
+			if err != nil {
+				panic(err)
+			}
+
+			args = append(args, []string(imageInspect.Config.Entrypoint)...)
```
arturo-c: This is the same logic implemented in kubernetes -> https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes

arturo-c: This is the same logic implemented in kubernetes -> https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes

bonifaido:
Some of this implementation has been done in https://github.com/banzaicloud/bank-vaults/pull/472 but that is Docker Registry specific, I don't know what kind of implementation of the registry AWS uses, but anyhow I would do along that path (doing this in vault-env).

bonifaido:
We are going to move the registry reading to the webhook as you did here @arturo-c, please stay tuned, will come back with the details early next week.

arturo-c:
@bonifaido cool, let me know if I can help

bonifaido:
Hi @arturo-c ! Some effort has been put into moving the whole registry stuff in a centralized way into the webhook, the code is now located here: https://github.com/banzaicloud/bank-vaults/blob/master/cmd/vault-secrets-webhook/registry/registry.go

Could you please rebase your work on top of that?

bonifaido:
Closing this, for now, please feel free to reopen in case the implementation is still needed.

PR #492, opened 2019-06-03
Check auth credentials existence

baluchicken:
Check auth credentials existence

### DIFF
--- a/cmd/vault-secrets-webhook/secret.go
+++ b/cmd/vault-secrets-webhook/secret.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


cmd/vault-secrets-webhook/secret.go
```diff
@@ -78,6 +79,9 @@ func mutateDockerCreds(secret *corev1.Secret, dc *dockerCreds, vaultConfig vault
 		if strings.HasPrefix(string(creds.Auth), "vault:") {
 			logger.Debugf("auth %s %s", key, creds.Auth)
 			split := strings.Split(creds.Auth, ":")
+			if len(split) < 4 {
```
baluchicken:
```suggestion
			if len(split) != 4 {
```

Issue #494, opened 2019-06-04
[operator] distribute the generated Vault CA secret in specified namespaces


bonifaido:
**Is your feature request related to a problem? Please describe.**
The `vault-tls` Secret containing the CA certificate for Vault is available in the namespace where the actual Vault instance got provisioned, however, clients from other namespaces can't mount this Secret to validate the server properly (like the Istio [node-agent](https://github.com/istio/istio/blob/master/security/pkg/nodeagent/caclient/providers/vault/client.go#L109).

**Describe the solution you'd like**
With a specified list of namespaces or using `*` for any namespaces, we should be able to distribute the CA certificate via the Operator to those.

**Describe alternatives you've considered**
N/A

PR #496, opened 2019-06-04
[operator] CA syncing into namespaces

bonifaido:
Fixes: #494


operator/deploy/operator-rbac.yaml
```diff
@@ -72,6 +72,21 @@ rules:
 
 ---
 
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: secret-creator
```
primeroz: being this a cluster role, while i understand this is just and example, i think i would like a more descriptive name than "secret-creator" 

People tend to just copy those examples ( i know i do )

_vault-secrets-creator_ ?

bonifaido: Yes, you are right, renaming :)

operator/pkg/controller/vault/vault_controller.go
```diff
@@ -227,6 +247,45 @@ func (r *ReconcileVault) Reconcile(request reconcile.Request) (reconcile.Result,
...
+				var namespaceList corev1.NamespaceList
+				if err := r.client.List(context.TODO(), &client.ListOptions{}, &namespaceList); err != nil {
+					return reconcile.Result{}, fmt.Errorf("failed to list namespaces: %v", err)
+				}
+
+				for _, namespace := range namespaceList.Items {
+					namespaces = append(namespaces, namespace.Name)
+				}
+			} else {
+				namespaces = v.Spec.CANamespaces
```
primeroz: Should we verify that the namespaces passed as list actually exist and only add add to the namespaces var those that do ? 

It would avoid failing shortly after if any of the specified namespaces do not exist

operator/pkg/controller/vault/vault_controller.go
```diff
@@ -227,6 +247,45 @@ func (r *ReconcileVault) Reconcile(request reconcile.Request) (reconcile.Result,
...
+			}
+
+			for _, namespace := range namespaces {
+				currentSecret.SetNamespace(namespace)
+				err = createOrUpdateObjectWithClient(r.nonNamespacedClient, &currentSecret)
```
primeroz: Dumb question , we normally do run `controllerutil.SetControllerReference` on newly created objects.

Don't we do it here because is inherited from the get of current secret in line https://github.com/banzaicloud/bank-vaults/blob/29d40926c0b5f3856ab77725e635a451544e7e60/operator/pkg/controller/vault/vault_controller.go#L256 ?

operator/deploy/cr-etcd-ha.yaml
```diff
@@ -68,8 +68,8 @@ spec:
         roles:
           # Allow every pod in the default namespace to use the secret kv store
           - name: default
-            bound_service_account_names: default
-            bound_service_account_namespaces: default
+            bound_service_account_names: ["default", "vault-secrets-webhook"]
```
primeroz: Are these changes to the example unrelated to this merge request or i am simply not getting it ?

operator/pkg/controller/vault/vault_controller.go
primeroz: Maybe with a warning log that one of the namespaces was not found.

I feel that in the case of a wrong list of namespaces when a namespace does not exist failing hard is a bit too much since no workload would be expecting to find that secret anyway

operator/deploy/cr-etcd-ha.yaml
bonifaido: Yes, these were missed from a previous PR and somehow it started to cause issues during the CI integration of this branch.

operator/pkg/controller/vault/vault_controller.go
bonifaido: Yes it is inherited, no need to set it again.

operator/pkg/controller/vault/vault_controller.go
bonifaido: Fixing this right now, thanks, good idea!

PR #497, opened 2019-06-04
Fix typo in vault path

pbalogh-sa:
- update mutating-webhook readme
- fix example vault-secret


pbalogh-sa:
acceptance-test will pass after https://github.com/banzaicloud/bank-vaults/pull/496/commits/f96941844987347db15399d9732ee01f4644ed46 is merged

PR #500, opened 2019-06-07
vault-env: added ability to only warn for missing secrets

### LINKED ISSUE #490: feature request: vault-env option to ignore missing secrets
I have a use case for vault-env where I need to ignore errors when a specific secret is not present in Vault.

It would be nice to be able to tweak the vault-env command to skip over missing secrets. Perhaps this could be enabled by providing an annotation.

E.g.
```yaml
vault.security.banzaicloud.io/vault-ignore-missing-secrets: "true"
```

When this is enabled vault-env could then use `logger.Warn` instead of `logger.Fatal` for missing secrets.

stephenmuss:
This change introduces the annotation `vault.security.banzaicloud.io/vault-ignore-missing-secrets`.

When set to `true` vault-env will not exit the process but will instead log a warning.

Closes #490

### DIFF
--- a/cmd/vault-env/main.go
+++ b/cmd/vault-env/main.go
@@ -33,28 +33,29 @@ type sanitizedEnviron []string
 var logger *zap.Logger
 
 var sanitizeEnvmap = map[string]bool{
-	"VAULT_TOKEN":           true,
-	"VAULT_ADDR":            true,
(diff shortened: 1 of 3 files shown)


stephenmuss:
Thank you. Very pleased to contribute.

Issue #504, opened 2019-06-10
operator/deploy/config-jwt-groups.yaml example does not correctly configure JWT backend


robertgates55:
**Describe the bug**
operator/deploy/config-jwt-groups.yaml example does not correctly configure JWT backend

**To Reproduce**
1. Install the `banzaicloud/vault-operator` helm chart (eg latest `0.2.15`)
2. Deploy a working vault
3. Add a `- type: jwt` section to `spec.externalConfig.auth` to add the example from `operator/deploy/config-jwt-groups.yaml` yaml

**Expected behavior**
This should create a `jwt` auth method, populated with the provided config options

**Actual behaviour**
Installs a `jwt` auth method, but with entirely empty config (see screenshot)

**Screenshots**
![image](https://user-images.githubusercontent.com/7437497/59199366-aa943500-8b8d-11e9-9f74-ecebf892526a.png)

primeroz:
Hi

It probably means that bank-vaults is failing to configure the backend.

Can you post the vault-configurer logs and the configuration snippet you are applying ?

I use it myself on a daily basis and it works :)

robertgates55:
So I'm actually using the oidc provider, but copying the same pattern from the `jwt` provider. I've also tried jwt with the same result.

To test again, I've just detroyed the vault instance and re-applied from scratch. The auth section here looks like:

primeroz:
Yeah oidc and jwt should be same. 

I think you missed the config actually. Also the logs would be great.

robertgates55:
(sorry - pressed send too soon!)

primeroz:
Ok. So the problem , at least from my phone debug skill, is that I don't think oidc is implemented as Auth backend in bank vaults.

 I only did implement jwt ( that's all I needed ) which is consistent with the issue you are having.
Auth backend get mounted but not configured.

We can do it pretty easily. I will give it a shot tomorrow or feel free to submit a pr if you feel like it.

robertgates55:
Ah yeah. Looks fairly likely.

Presume this is all we're after? (I've not raised a PR direct to here as I'm not sure on the normal flow)

https://github.com/robertgates55/bank-vaults/commit/3b1d1e26c1596540ee5c9745e61ce5d40744245d

bonifaido:
Looks cool, please submit it to this repository @robertgates55! Thanks

primeroz:
Yeah pretty much but I am on my phone so can't really double check. 

[...]

PR #505, opened 2019-06-11
Add support for oidc auth type

robertgates55:
Resolves Issue #504 

Added switch case for auth type 'oidc'.

### DIFF
--- a/pkg/vault/operator_client.go
+++ b/pkg/vault/operator_client.go
@@ -610,6 +610,23 @@ func (v *vault) configureAuthMethods(config *viper.Viper) error {
 			if err != nil {
 				return fmt.Errorf("error configuring jwt roles on path %s for vault: %s", path, err.Error())
 			}
+		case "oidc":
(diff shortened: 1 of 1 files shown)


pkg/vault/operator_client.go
```diff
@@ -610,6 +610,23 @@ func (v *vault) configureAuthMethods(config *viper.Viper) error {
 			if err != nil {
 				return fmt.Errorf("error configuring jwt roles on path %s for vault: %s", path, err.Error())
 			}
+		case "oidc":
```
primeroz: Since everything is the same as the JWT case can we merge the 2 and use the "authMethodType" in the error messages to differentiate ? 

preety much the same as https://github.com/banzaicloud/bank-vaults/blob/3b1d1e26c1596540ee5c9745e61ce5d40744245d/pkg/vault/operator_client.go#L656

robertgates55: Yes. That'd be nicer. In my rush to make the PR I neglected to apply much _thought_.

Update incoming. 👍

PR #512, opened 2019-06-12
[vault-secrets-webhook] Mutate configMaps

moskitone:
Added option to mutate kubernetes configMaps. 
Added kubernetes annotation vault.security.banzaicloud.io/mutate-configmap.

This would be useful to use with other charts not supporting things like consulTemplate.

### DIFF
--- a/cmd/vault-secrets-webhook/configmap.go
+++ b/cmd/vault-secrets-webhook/configmap.go
@@ -17,11 +17,9 @@ package main
 import (
 	"fmt"
 	"strconv"
-	"strings"
 
 	"github.com/banzaicloud/bank-vaults/pkg/vault"
 	vaultapi "github.com/hashicorp/vault/api"
-	"github.com/spf13/cast"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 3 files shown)


pbalogh-sa:
Mutating secrets is a similar feature, and storing credentials in k8s secret isn't too secure, but we can restrict access to secrets. Putting credentials into configmaps is worse than using secrets In my  opinions is. You can use secret mutation feature instead of mutating configmaps.

moskitone:
I agree with you about security case.
 But there are lots of software/charts already build with configuration files in configMaps. For example https://github.com/helm/charts/tree/master/stable/ingressmonitorcontroller or https://github.com/helm/charts/tree/master/stable/hazelcast .

pbalogh-sa:
The configMap mutation sounds reasonable in that case when you use official helm charts, as you mention. We will check your PR soon.

cmd/vault-secrets-webhook/configmap.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return err
+	}
+	for key, value := range mapData {
+		configMap.Data[key] = string([]byte(value))
+	}
+	return nil
+}
+
+func getDataFromVault(data map[string]string, vaultClient *vault.Client) (map[string]string, error) {
```
bonifaido: Couldn't we reuse the method from the `Secret` handling part? https://github.com/banzaicloud/bank-vaults/blob/master/cmd/vault-secrets-webhook/secret.go#L142

bonifaido:
@moskitone  I agree as well that this is useful, some charts are inherently using ConfigMaps and a lot of people don't like to maintain different forks of those (we neither). I had one review comment about possible code duplication. Thanks for your contribution!

cmd/vault-secrets-webhook/configmap.go
moskitone: I've reused getDataFromVault in Secret handling because there is support for having several secrets in one line. Also added handling for quoted secrets and some other special characters.

PR #516, opened 2019-06-14
use 'reclaim' policy for host-path PVs

benjamink:
While the snippet for setting up a PV at the end of this file is only for example usage/testing it needs to use the `Reclaim` policy instead of `Delete` as `Delete` isn't supported for `hostPath` PVs.  Having this set to `Delete` would cause the contents of the PV to *NOT* be deleted even when the PV was deleted which would cause bringing up subsequent iterations of Vault to fail.


benjamink:
This should be `Recycle` not `Reclaim` - fixing.

bonifaido:
Thank you @benjamink, could you please sign the CLA?

benjamink:
Looks like there are bigger issues with this than the single line I changed that are preventing me from merging even after signing the CLA.

bonifaido:
Let me rerun the tests.

bonifaido:
Sorry, I have messed up one of the tools used in the acceptance test, now it should be ok.

Issue #515, opened 2019-06-14
[vault-secrets-webhook] Mutated pods fail to be scheduled when ResourceQuotas enabled


devlounge:
**Describe the bug**
When ResourceQuotas are set on a namespace, the mutated pod won't be able to run, because the mutated pod does not set any resources on the injected containers.

**Expected behavior**
Provide a way to set the container.Resources for each type of container (initcontainer and sidecars)

Kubernetes: 1.13.4
Version: 0.4.16

```
...
status:
  conditions:
  - lastTransitionTime: "2019-06-14T15:07:38Z"
    message: 'pods "banzai-poc-77899885b9-nvfzr" is forbidden: failed quota: default-quotas:
      must specify limits.memory,requests.cpu'
```

pbalogh-sa:
@devlounge Adding resource quotas to containers makes sense. Thanks

PR #518, opened 2019-06-17
specify global resourcequota everywhere

bonifaido:
Fixes: #515 

- set resource limits on ALL resources
- set a `resourcequota` globally in CI to verify this in the acceptance test
- make the acceptance test more robust

### DIFF
--- a/cmd/vault-secrets-webhook/main.go
+++ b/cmd/vault-secrets-webhook/main.go
@@ -1,4 +1,4 @@
-// Copyright © 2018 Banzai Cloud
+// Copyright © 2019 Banzai Cloud
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


cmd/vault-secrets-webhook/main.go
```diff
@@ -182,6 +195,12 @@ func getContainers(vaultConfig vaultConfig, containerEnvVars []corev1.EnvVar, co
 		SecurityContext: securityContext,
 		Env:             containerEnvVars,
 		VolumeMounts:    containerVolMounts,
+		Resources: corev1.ResourceRequirements{
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("50m"),
```
primeroz: what did you base those numbers on ? 

I am only concerned with the `consul-template` one. I wonder if , given a high number of templates and keys to render, it will run out of memory and so get OOMKilled or it will take too long to render because of only 50m cpu

Maybe we couuld allow to change those for the consul-template since that one is a sidecar container rather than an init one that will run to completion and be forgotten ?

bonifaido: Just made it configurable, thanks, I think I just blindly changed all limits with copy paste.

cmd/vault-secrets-webhook/main.go
```diff
@@ -382,6 +384,18 @@ func parseVaultConfig(obj metav1.Object) vaultConfig {
 		vaultConfig.ctOnce = false
 	}
 
+	if val, err := resource.ParseQuantity(annotations["vault.security.banzaicloud.io/vault-ct-cpu"]); err != nil {
+		vaultConfig.ctCPU = val
```
primeroz: not sure about it, but don't we need to run the `MustParse` here as well ?

Issue #530, opened 2019-06-25
[Operator] Vault CRD does not support priorityClassName


ryandbump:
**Is your feature request related to a problem? Please describe.**
Vault pods are being preempted in when cluster resource pressure is high. 

**Describe the solution you'd like**
I would like to set `priorityClassName` on the vault pods to prevent unwanted preemption. Ideally, I would be able to specify a different `priorityClassName` for the pods created by each controller. 

I think the best way to achieve this, and to enable more features in the future, is to add attributes like `vaultPodSpec`, `configurererPodSpec`, and `etcdPodSpec` to the CRD. The values in these attribute would be merged with the pod specs created in `vault_controller.go`. 

**Describe alternatives you've considered**
Currently the only way to add `priorityClassName` to pods is to edit the pods after they have been created. This is not sustainable.

matyix:
Hey @ryandbump - thanks for submitting the issue! Would you be interested to work on/contribute to this feature? In our usage pattern, we separated Vault (namespace or cluster) from the workload and we did not add `riorityClassName`, though I agree with you, it's a totally valued requirement.

ryandbump:
Yeah I could take stab at it.

baluchicken:
Fixed by #531 
thanks @ryandbump

matyix:
@ryandbump thanks for opening the issue and contributing it. @pbalogh-sa added some usage example as well in #536.

PR #531, opened 2019-06-25
[Operator] Support generic PodSpec fields on CRD

ryandbump:
This PR addresses https://github.com/banzaicloud/bank-vaults/issues/530

This enables users to update the pod spec for created pods without having to add new fields to the CRD. The merge functionality ensures operator defined fields on the PodSpec are always respected. 

In the future, following a deprecation schedule, we could remove all of these fields and achieve the same results by using only the pod spec. 

```go
PodAntiAffinity       string                        `json:"podAntiAffinity"`
NodeAffinity          v1.NodeAffinity               `json:"nodeAffinity"`
NodeSelector          map[string]string             `json:"nodeSelector"`
Tolerations           []v1.Toleration               `json:"tolerations"`
ServiceAccount        string                        `json:"serviceAccount"`
Volumes               []v1.Volume                   `json:"volumes,omitempty"`
VolumeMounts          []v1.VolumeMount              `json:"volumeMounts,omitempty"`
VaultEnvsConfig       []v1.EnvVar                   `json:"vaultEnvsConfig"`
Resources             *Resources                    `json:"resources,omitempty"`
```
https://github.com/banzaicloud/bank-vaults/blob/master/operator/pkg/apis/vault/v1alpha1/vault_types.go#L133-L141

### DIFF
--- a/operator/pkg/apis/vault/v1alpha1/vault_types.go
+++ b/operator/pkg/apis/vault/v1alpha1/vault_types.go
@@ -26,7 +26,7 @@ import (
 	"github.com/Masterminds/semver"
 	"github.com/spf13/cast"
 	v1 "k8s.io/api/core/v1"
-	v1beta1 "k8s.io/api/extensions/v1beta1"
+	"k8s.io/api/extensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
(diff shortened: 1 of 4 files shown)


operator/pkg/apis/vault/v1alpha1/vault_types.go
```diff
@@ -112,8 +112,10 @@ type VaultSpec struct {
 	Annotations                map[string]string     `json:"annotations"`
 	VaultAnnotations           map[string]string     `json:"vaultAnnotations"`
 	VaultLabels                map[string]string     `json:"vaultLabels"`
+	VaultPodSpec               v1.PodSpec            `json:"vaultPodSpec"`
```
primeroz: Did you regenerate the operator deepcopy file with `operator-sdk generate k8s` ? 

I would expect to see changes https://github.com/banzaicloud/bank-vaults/blob/master/operator/pkg/apis/vault/v1alpha1/zz_generated.deepcopy.go too

pbalogh-sa:
> Really nice.
> 
> Could we add some test to the "acceptance.sh" to use both of the new podspec fields ?

We've opened the issue: #533

operator/pkg/apis/vault/v1alpha1/vault_types.go
ryandbump: This was difficult to do. The `operator-sdk` tool does not support go modules yet (at least not in a release, HEAD has module support). and the operator project does not have a go.mod file so even using the `operator-sdk` with module support fails.

PR #532, opened 2019-06-26
Cluster vs namespaced Role seemed to be swapped around, the chart dep…

jurgenweber:
…loys it 'this way' as well (chart Cluster Role link: https://github.com/banzaicloud/banzai-charts/blob/master/vault-operator/templates/role.yaml)

as per conversation with 'Jared E' on slack:

```
{"level":"info","ts":1561520235.3518147,"logger":"metrics","msg":"Metrics Service object created","Service.Name":"vault-operator","Service.Namespace":"default"}
{"level":"info","ts":1561520235.3518553,"logger":"cmd","msg":"Starting the Cmd."}
E0626 03:37:15.353833       1 reflector.go:134] pkg/mod/k8s.io/client-go@v2.0.0-alpha.0.0.20181213151034-8d9ed539ba31+incompatible/tools/cache/reflector.go:95: Failed to list *v1alpha1.Vault: vaults.vault.banzaicloud.com is forbidden: User "system:serviceaccount:default:vault-operator" cannot list vaults.vault.banzaicloud.com at the cluster scope
```

### DIFF
--- a/operator/deploy/operator-rbac.yaml
+++ b/operator/deploy/operator-rbac.yaml
@@ -6,7 +6,7 @@ metadata:
 ---
 
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: ClusterRole
 metadata:
   name: vault-operator
 rules:
@@ -73,7 +73,7 @@ rules:
 ---
 
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
+kind: Role
 metadata:
   name: vault-operator
 rules:


primeroz:
@jurgenweber @bonifaido @pbalogh-sa 

I am not sure about something in this change.

With the swap around ( which make sense ) now some of the verbs ( list and watch ) for the namespaces resource are a role, which is namespace scoped ... that seems like a contradiction to me 
https://github.com/jurgenweber/bank-vaults/blob/906bf409717e762b3f12616043fab07480b9586d/operator/deploy/operator-rbac.yaml#L82

Also the secrets rbac in the Role is already covered by the secrets in the clusterrole.

Should we extend the namespace in the cluster role ( to include list and watch since right now is get only ) and drop the role entirely ?
Do we even need to list and watch namespaces ?

pbalogh-sa:
@primeroz Thanks for your remark! We will review roles and clusterroles.

PR #538, opened 2019-06-28
Add support for "create_only" field to secrets

### LINKED ISSUE #537: Add support for a "create_only" flag to secrets configurations
On the GCP Secrets whenever a roleset configuration, with no changes from previous one, is re-applied it will cause Vault to delete and recreate the google service account invalidating all keys / tokens associated with that SA

This behavior is particularly bad since with a kubernetes operator it will happen *relatively often* that the GCP roleset configuration will be re-applied
* When the operator is upgraded
* When the pod for the vault-configurer is killed and respawned
* When the config file with the gcp secrets changes 

I *believe* this to be a bug with the Vault GCP Secrets plugin and created an issue upstream
Upstream issue: https://github.com/hashicorp/vault-plugin-secrets-gcp/issues/43

Since i am not really sure that i will get an answer very quickly upstream and / or that it will be recognised as a bug i have implemented support for a flag "create_only" that will make the configurator skip a specific secret configuration if it already exists. 

Open upstream bug

**Additional context**
MR to be coming soon

primeroz:
fixes #537 

I have been trying to get something to test this in the acceptance-test.sh but failed so far because the only backend i know to test this with is GCP Secrets and it does require a google configuration to be actually mounted and used

### DIFF
--- a/operator/deploy/cr-gcs-ha.yaml
+++ b/operator/deploy/cr-gcs-ha.yaml
@@ -45,6 +45,12 @@ spec:
         rules: path "secret/*" {
           capabilities = ["create", "read", "update", "delete", "list"]
           }
+      - name: gcp_editor_access
+        rules: |
+          path "gcp/key/editor"
+          {
+            capabilities = ["read"]
+          }
     auth:
       - type: kubernetes
         roles:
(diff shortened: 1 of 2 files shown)


primeroz:
@bonifaido example added.

Issue #364, opened 2019-03-19
kubebuilder.controller Reconciler error . failed to update vault status


raoofm:
**Describe the bug**
kubebuilder.controller Reconciler error . failed to update vault status

**To Reproduce**
1. operator version 0.4.13
2. vault version 1.1.0-beta2

**Operator Logs**
```
{"level":"info","ts":1553012934.9785693,"logger":"cmd","msg":"Go Version: go1.12"}
{"level":"info","ts":1553012934.9787564,"logger":"cmd","msg":"Go OS/Arch: linux/amd64"}
{"level":"info","ts":1553012934.978833,"logger":"cmd","msg":"operator-sdk Version: v0.4.1"}
{"level":"info","ts":1553012934.9788764,"logger":"cmd","msg":"Watched namespace: "}
{"level":"info","ts":1553012934.9836006,"logger":"leader","msg":"Trying to become the leader."}
{"level":"info","ts":1553012934.9845932,"logger":"cmd","msg":"Liveness probe listening on: 8080"}
{"level":"info","ts":1553012936.3835745,"logger":"leader","msg":"No pre-existing lock was found."}
{"level":"info","ts":1553012936.3945363,"logger":"leader","msg":"Became the leader."}
{"level":"info","ts":1553012936.6002703,"logger":"cmd","msg":"Registering Components."}
{"level":"info","ts":1553012936.6008043,"logger":"kubebuilder.controller","msg":"Starting EventSource","controller":"vault-controller","source":"kind source: /, Kind="}
{"level":"info","ts":1553012936.6010134,"logger":"cmd","msg":"Starting the Cmd."}
{"level":"info","ts":1553012936.7014093,"logger":"kubebuilder.controller","msg":"Starting Controller","controller":"vault-controller"}
{"level":"info","ts":1553012936.8016922,"logger":"kubebuilder.controller","msg":"Starting workers","controller":"vault-controller","worker count":1}
{"level":"info","ts":1553012936.8019593,"logger":"controller_vault","msg":"Reconciling Vault","Request.Namespace":"security","Request.Name":"vault"}
{"level":"info","ts":1553012977.1772525,"logger":"controller_vault","msg":"Reconciling Vault","Request.Namespace":"security","Request.Name":"vault"}
[...]

PR #545, opened 2019-07-11
Use objectmatcher to avoid unnecessary updates. 

pepov:
- Also fix ingress apiversion to avoid constant updates because of the mismatch.
 - Add flag to enable verbose logging. (logs with log level 1 or lower will be shown)
 - Lock some dependencies to their original version for compatibility.


bonifaido:
I would say that this is a possible fix for: https://github.com/banzaicloud/bank-vaults/issues/364

PR #546, opened 2019-07-11
Object matcher and vault CR status matcher fixes

pepov:
- The annotation required for object matching must be applied after the patch calculation, but before setting the resource version.
- Pod names must be sorted to avoid unncessary updates.


pepov:
also related to https://github.com/banzaicloud/bank-vaults/issues/364 as it fixes frequent CR updates

Issue #541, opened 2019-07-02
Support for env interpolation in Vault config


ekarlso:
**Describe the bug**
Vault crashes when using env var interpolation

```
Error initializing storage of type mysql: failed to check mysql schema exist: Error 9999: An internal error has occurred. Please retry or report your issues.
```

**To Reproduce**
Have a MySQL storage backend at azure with mysql config for vault.

Create the Vault CR below
```
apiVersion: vault.banzaicloud.com/v1alpha1
kind: Vault
metadata:
  name: vault
  namespace: vault-system
spec:
  size: 2
  image: vault:1.1.0
  bankVaultsImage: banzaicloud/bank-vaults:latest

  unsealConfig:
    azure:
      keyVaultName: myvault-vault-test

  serviceAccount: vault

  config:
    storage:
      mysql:
        username: '${env "ROOT_USERNAME"}'
        password: '${env "ROOT_PASSWORD"}'
        address: "vault.mysql.database.azure.com:3306"
        ha_enabled: "true"
        tls_ca_file: /etc/digicert-ca/ca.pem

    listener:
      tcp: 0.0.0.0:8200
      tls_cert_file: /vault/tls/server.crt
      tls_key_file: /vault/tls/server.key

    api_addr: https://vault.vault-infra:8200
    ui: true
    log_level: "Trace"

  externalConfig:
    policies:
      - name: allow_secrets
        rules: path "secret/*" {
          capabilities = ["create", "read", "update", "delete", "list"]
          }
    auth:
      - type: kubernetes
        roles:
          - name: default
            bound_service_account_names: default
            bound_service_account_namespaces: default
            policies: allow_secrets
            ttl: 1h
  volumes:
    - name: digicert-secret
      secret:
        secretName: digicert-root-ca
  volumeMounts:
    - mountPath: /etc/digicert-ca/
      name: digicert-secret
      readOnly: true
  vaultEnvsConfig:
    - name: ROOT_USERNAME
      valueFrom:
        secretKeyRef:
          name: vault-db
          key: username
    - name: ROOT_PASSWORD
      valueFrom:
        secretKeyRef:
          name: vault-db
          key: password
    - name: VAULT_LOG_LEVEL
      value: Trace

  credentialsConfig:
    env: AZURE_AUTH_LOCATION
    path: /etc/azure/credentials
    secretName: azure-cred
```

**Expected behavior**
Vault should start.

**Versions:**
Bank Vault 0.4.17
Vault 1.1

[...]

PR #552, opened 2019-07-16
Complete MySQL HA example and Vault config templating support

bonifaido:
Fixes: #541

### DIFF
--- a/operator/pkg/controller/vault/vault_controller.go
+++ b/operator/pkg/controller/vault/vault_controller.go
@@ -1027,11 +1027,16 @@ func statefulSetForVault(v *vaultv1alpha1.Vault, externalSecretsToWatchItems []c
 		return nil, fmt.Errorf("more than 1 replicas are not supported without HA storage backend")
 	}
 
+	configSizeLimit := resource.MustParse("1Mi")
+
 	volumes := withTLSVolume(v, withCredentialsVolume(v, []corev1.Volume{
 		{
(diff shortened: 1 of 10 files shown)


operator/pkg/controller/vault/vault_controller.go
```diff
@@ -1120,10 +1141,6 @@ func statefulSetForVault(v *vaultv1alpha1.Vault, externalSecretsToWatchItems []c
 				Args:            []string{"server"},
 				Ports:           containerPorts,
 				Env: withTLSEnv(v, true, withCredentialsEnv(v, withVaultEnv(v, []corev1.EnvVar{
```
primeroz: I am sure i am missing something but , if we are removing the "VAULT_LOCAL_CONFIG" env variable from the vault container and now we are writing the _rendered vault configuration_ into /vault/config/vault.json ... don't we need to start the vault server with "-config=/vault/config/vault.json" ?

primeroz: nevermind. 

I can see this is handled by the vault container automatically

```
The container has a Vault configuration directory set up at /vault/config and the server will load any HCL or JSON configuration files placed here by binding a volume or by composing a new image and adding files. Alternatively, configuration can be added by passing the configuration JSON via environment variable VAULT_LOCAL_CONFIG
```

bonifaido:
Good idea @primeroz I will add it thanks

PR #555, opened 2019-07-18
RBAC - allow Operator to list replicasets

sosoriov:
Bug fix Vault Operator: user forbidden for access metrics in replicaset.app 

Grant access to list replicaset K8s objects.

### DIFF
--- a/operator/deploy/operator-rbac.yaml
+++ b/operator/deploy/operator-rbac.yaml
@@ -40,6 +40,7 @@ rules:
     resources:
       - replicasets
     verbs:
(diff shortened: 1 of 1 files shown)


bonifaido:
Please change your commit to the same email address you are using on GitHub otherwise the CLA check won't pass.

bonifaido:
@sosoriov could you please make this commit with your GitHub user's email?

sosoriov:
Nandor,

Sorry for the late reply.

It should be OK now.

PR #559, opened 2019-07-22
webhook + operator: transit example

bonifaido:
- vault-env: add `VAULT_TOKEN="vault:login"` special value support, this passes through the VAULT_TOKEN to the application from the login process
- added vault transit unseal example supported and automated by operator and the webhook

### DIFF
--- a/cmd/vault-env/main.go
+++ b/cmd/vault-env/main.go
@@ -27,6 +27,10 @@ import (
 	"github.com/spf13/cast"
 )
 
+// The special value for VAULT_ENV which marks that the login token needs to be passed through to the application
+// which was acquired during the new Vault client creation
+const vaultLogin = "vault:login"
+
 type sanitizedEnviron []string
 
 var sanitizeEnvmap = map[string]bool{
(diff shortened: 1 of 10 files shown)


cmd/vault-env/main.go
```diff
@@ -65,23 +62,35 @@ func (environ *sanitizedEnviron) append(iname interface{}, ivalue interface{}) {
...
-	flag.Parse()
+	ignoreMissingSecrets := os.Getenv("VAULT_IGNORE_MISSING_SECRETS") == "true"
 
-	logger, _ = zap.NewProduction()
-	defer logger.Sync()
+	// The login procedure takes the token from a file (if using Vault Agent)
+	// or requests one for itself (Kubernetes Auth), so if we got a VAULT_TOKEN
+	// for the special value with "vault:login"
+	originalVaultTokenEnvVar := os.Getenv("VAULT_TOKEN")
+	if originalVaultTokenEnvVar != "" {
```
pepov: maybe it would be better to check explicitly for the special value and let a regular VAULT_TOKEN pass through to support external auth in case of testing or other means

bonifaido: Thanks, added the check.

Issue #565, opened 2019-07-25
Issues and roadblocks trying to expose Vault via Ingress on GKE


fraenkel:
My goal was to stand up Vault on GKE for both internal and external consumption.
I did figure out a solution that works although I uncovered some issues and limitations along the way that I would like to describe.

The final solution was to do the simplest solution and just let a default vault get stood up with an Ingress with a LetsEncrypt certificate to the api-port on Vault.

- Attempt 1 - Use the generated certificate on the Ingress
The upside/downside is that it requires the CA to be handed to all clients.
**Issue:** The externalTrafficPolicy is set to Cluster. Since we need NodePorts, my preferred setting would be Local.
**Issue:** This doesn't work because there is no way to inject additional DNS SANs in the certificate to allow inbound requests to match.
A negative is that it wouldn't work well for browsers if implemented.
If I wanted to truly lock it down with mutual TLS, Ingress wouldn't work because there would be no way to go from the LB -> Vault.

- Attempt 2 - Use a LoadBalancer on the Service
This has the same issue as above. The certificate does not have the DNS name included in the Certificate.

- Attempt 3 - Disable TLS
To simplify the setup, I wanted to drop TLS and just protect it at the LoadBalancer using LetsEncrypt.
**Issue:** The _DisableTLS_ checks are incorrect. The Vault config for Listeners are a list of maps. The code assumes it's a single map. I never got this to work.

- Attempt 4 - Add an additional listener port
I created an additional TLS port with my LetsEncrypt certificate and exposed it on the Service
**Issue:** The service annotation _cloud.google.com/app-protocols_ is hard coded and doesn't reflect the additional ports.
**Issue:** Cannot override the healthcheck ports. GKE Ingress needs to invoke the liveness probe. If the backend port matches in the Ingress and the health-check the correct path is used. When they differ, it will use / which always fails.

**Issue:** No way to disable the statsd-exporter. We happen to have a statsd receiver on every node, so its unneeded. The prometheus annotations can be overridden  so that is less of an issue.

primeroz:
I run successfully a setup similar to your attempt 4 based on the example from https://github.com/banzaicloud/bank-vaults/blob/master/operator/deploy/cr-customports.yaml 

[...]

PR #569, opened 2019-07-29
operator: network enhancements

bonifaido:
- proper support for multiple Vault listeners (expose all listener ports in the Service)
- make TLS secret work with Ingress as well
- GKE ingress example added

Fixes https://github.com/banzaicloud/bank-vaults/issues/565


fraenkel:
I added a certExtraSANS and an additional listener with TLS disabled.

1. The service lists all my ports correctly, but the annotation is incorrect:
`cloud.google.com/app-protocols: '{"api-port":"HTTPS"}'`
It doesn't match the 8200-port or 8300-port in the list. So it might be an issue when merging but it's still wrong. The vault-operator logs don't show any error.

2. I added the following:
```
certExtraSANS:
    - vault.example.com
```
The resulting vault-tls, only had:
`DNS:vault.vault, IP Address:127.0.0.1`
I had to delete the secret and have it recreate, but now I have more entries,
server.crt, server.key, tls.crt, tls.key, ca.crt.

bonifaido:
Thanks for checking! 
  1. This seems to be wrong, thanks, I have to look into this.
  2. certExtraSANS: this works only if the certificate is not there yet or about to expire (threshold is configurable), so this one seems ok.

bonifaido:
I just fixed the Service annotation issue.

bonifaido:
@fraenkel have you managed to take a look at this after the service annotations fix?

fraenkel:
Sorry, I reviewed what you did which should fix the replacement issue. I would need to verify that the annotation is actually correct.

bonifaido:
Absolutely no pressure, just asking 👍

fraenkel:
It's fixed. The service annotation works correctly once ingress is enabled.

bonifaido:
Do you find this version easier to use?

fraenkel:
Yes, although I believe in the end I am going with 2 listeners, and an extra service (LoadBalancer) which exposes my single listener.
I didn't want to have all the other ports exposed either as NodePorts or accessible via the LoadBalancer.
While the Ingress only allows 443 through, the CA isn't something a browser doesn't complain about, and it causes an extra SSL hop.
In the end I think it will come down to ease vs control. Both work, but it's a choice.

bonifaido:
I still think we can use some kind of annotations to make this possible, I mean we could just simply say that which listeners in the vault config we (don't) want to expose in the Service (deny or allow list), but I would say that should go into a separate PR.

operator/pkg/apis/vault/v1alpha1/vault_types.go
```diff
@@ -414,8 +426,49 @@ func (spec *VaultSpec) GetTLSExpiryThreshold() time.Duration {
...
+	}
+	return true
+}
+
+func (spec *VaultSpec) GetListenerPort(listener int) corev1.ServicePort {
+	listeners := spec.GetListeners()
+	tcp := cast.ToStringMap(listeners[listener]["tcp"])
+	_, port, err := net.SplitHostPort(cast.ToString(tcp["address"]))
+	if err != nil {
+		panic(err)
```
baluchicken: I think we should just return an error here instead of panicking.

operator/deploy/cr-gke-ingress.yaml
```diff
@@ -0,0 +1,199 @@
...
+spec:
+  # https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1
+  # storageClassName: ""
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+
+# ---
```
baluchicken: We should remove this unnecessary Persistent Volume from the example.

operator/pkg/apis/vault/v1alpha1/vault_types.go
primeroz: While i agree with this and i think we should handle the error where this function is called, in line 595 we call this inline with the creation of the service so we don't get a change to check if an `err` was returned but we should definetely panic there since it is used for the main port:8200 service

Would it be better though to return an error here and fetch the port in line 595 with a few more lines to handle the error and , possibly, print a message out ?

operator/pkg/controller/vault/vault_controller.go
```diff
@@ -725,25 +749,32 @@ func getServicePorts(v *vaultv1alpha1.Vault) ([]corev1.ServicePort, []corev1.Con
 	var containerPorts []corev1.ContainerPort
 
 	if len(v.Spec.ServicePorts) == 0 {
-		return []corev1.ServicePort{
+
+		for listener := range v.Spec.GetListeners() {
```
primeroz: Do we need to keep support for https://github.com/banzaicloud/bank-vaults/blob/3f85959bcc38684e448e073dcc4d17b021e5b9b0/operator/deploy/cr-customports.yaml#L28 with this new feature ? 

My understanding is that now every listener defined will be exposed in the service as a port so we should not need this serviceports in the spec anymore 

Should we just deprecate this settings and print out a message if is set but don't use it anymore ?

operator/pkg/controller/vault/vault_controller.go
```diff
@@ -1162,8 +1171,8 @@ func statefulSetForVault(v *vaultv1alpha1.Vault, externalSecretsToWatchItems []c
 				LivenessProbe: &corev1.Probe{
 					Handler: corev1.Handler{
 						HTTPGet: &corev1.HTTPGetAction{
-							Scheme: getVaultURIScheme(v),
-							Port:   intstr.FromString("api-port"),
+							Scheme: getVaultURIScheme(v, 0),
+							Port:   intstr.FromString(v.Spec.GetListenerPort(0).Name),
```
primeroz: Do we rely too much on the fact that listener number 0 is the "main one" using the TLS Secret generated from the operator ? 
while probably this is a safe bet it does feel a bit fragile.

in the https://github.com/banzaicloud/bank-vaults/blob/3f85959bcc38684e448e073dcc4d17b021e5b9b0/operator/deploy/cr-customports.yaml#L77 for example this is not true and , i believe, it would break some connection since the TLS for the first listener is not the one created by vault operator

This would affect many other places where we are expecing listener 0 to be the `internal` one , including line 1331 where the configurator gets the port to connect to

operator/pkg/controller/vault/vault_controller.go
```diff
@@ -725,25 +749,32 @@ func getServicePorts(v *vaultv1alpha1.Vault) ([]corev1.ServicePort, []corev1.Con
...
 	if len(v.Spec.ServicePorts) == 0 {
-		return []corev1.ServicePort{
+
+		for listener := range v.Spec.GetListeners() {
+			port := v.Spec.GetListenerPort(listener)
+			servicePorts = append(servicePorts, []corev1.ServicePort{
 				{
-					Name: "api-port",
-					Port: 8200,
+					Name: port.Name,
```
primeroz: The change of port-name might affect people's current deployment if they are creating any resource that reference those name outside of the vault-operator

Let's make sure to highlight this change as a `soft breaking` change in the release notes

bonifaido: I was thinking about checking if the port number is 8200 and give it the name "api-port" to keep compatibility, any thoughts?

primeroz: Might be enough in 99% cases ( I don't really see any reason to change this port but someone out there surely has a use case for doing it) , in that case also for 8201 and call it cluster-port ? 

I would still make sure to have a clear message in the release notes for this

bonifaido: Fixes both api-port and cluster-port.

operator/pkg/controller/vault/vault_controller.go
bonifaido: Should we index them by ports instead and use 8200 as the main one?

primeroz: that sounds much better.

The only little conflict i see is that with this PR we want to make ports not hardcoded for 8200 as it used to be. 
Now we will assume 8200 is the main one with the extra confusion of being able to change it.

should we just keep enforcing 8200 as main one then ? does this extra flexibility give any advantage ? 

but yeah, i think it would be better to assume the port 8200 as main one rather than index 0

chrisob:
@bonifaido Hate to be "that guy", but is this PR still being worked on? I'm looking forward to being able to rename ports consistently.

bonifaido:
Hi @chrisob, no problem, absolutely valid question. I'm in progress with other things (on other projects as well), but this PR is still in the pipe, and 100% something we would like to see in master.

bonifaido:
I'm trying to resurrect this branch, rebasing on master is already done, now I'm trying to reflect on the PR comments again, and will run some upgrade tests from 0.5.x to this branch, to see what breaks (hopefully nothing, that is the intention).

LinAnt:
@chrisob @bonifaido I can help testing this release if you want to check for Istio compatibility.

bonifaido:
That would be awesome @LinAnt ! Thanks

bonifaido:
Not going to be merged as it is, need to open a new PR for this in the future.

PR #572, opened 2019-07-31
Configurable webhook listen address

leominov:
Giving capabilities to serve requests on 443 port should be more easily than configuring firewall rules.

ref: https://github.com/banzaicloud/banzai-charts/tree/master/vault-secrets-webhook#about-gke-private-clusters


leominov:
Nope, we don't use Helm in our environments :)

PR #580, opened 2019-08-06
helm: build charts with circle

bonifaido:
Build and release charts from this git repository to the Banzai Helm repository.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,7 +131,7 @@ jobs:
       - run:
           name: Wait for minikube
           command: |
-            timeout 180s bash <<EOT
+            timeout 180s bash \<<EOT
               set -o pipefail
               function is_ready()
               {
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -164,3 +165,20 @@ workflows:
...
+          filters:
+            tags:
+              ignore: /.*/
+
+      - helm/publish-chart:
+          context: helm
+          charts-dir: 'charts'
+          filters:
+            tags:
+              only: /\S+\/\d+.\d+.\d+/
```
sagikazarmark: IMHO this should be more restrictive to tags like `charts/something-0.0.0`

.circleci/config.yml
```diff
@@ -1,7 +1,8 @@
-# Golang CircleCI 2.0 configuration file
-#
-# Check https://circleci.com/docs/2.0/language-go/ for more details
 version: 2
```
sagikazarmark: Shouldn't this be `2.1`?

bonifaido: Fixed, thanks!

PR #585, opened 2019-08-07
make the webhook failure policy configurable, also set the default to…

jurgenweber:
… Ignore, having it as fail means if the webhook starts failing all pods/resources will not schedule/update!


jurgenweber:
version bumped

PR #586, opened 2019-08-07
revert breaking change of vault tls path, mounts and env vars

jurgenweber:
partial revert of; https://github.com/banzaicloud/bank-vaults/commit/7338457003d7f019cbd96626d0073fa99bb6c128

does not seem necessary, not sure if this breaks something else but all of the doco (https://github.com/banzaicloud/bank-vaults/search?q=%2Fvault%2Ftls%2Fca.crt&unscoped_q=%2Fvault%2Ftls%2Fca.crt) and implementations is without the '-env', etc.

### DIFF
--- a/cmd/vault-secrets-webhook/main.go
+++ b/cmd/vault-secrets-webhook/main.go
@@ -242,7 +242,7 @@ func getVolumes(agentConfigMapName string, vaultConfig vaultConfig, logger *log.
 	if vaultConfig.tlsSecret != "" {
 		logger.Debugf("Add vault TLS volume to podspec")
 		volumes = append(volumes, corev1.Volume{
-			Name: "vault-env-tls",
+			Name: "vault-tls",
 			VolumeSource: corev1.VolumeSource{
 				Secret: &corev1.SecretVolumeSource{
 					SecretName: vaultConfig.tlsSecret,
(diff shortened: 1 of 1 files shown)


bonifaido:
Yes, this currently breaks the capability of mounting "vault-tls" volume to vault itself, next to its own "vault-tls" volume, which would like to use the transit engine. :) I would reverse the question, how this breaks other stuff?

jurgenweber:
well check the search link; https://github.com/banzaicloud/bank-vaults/search?q=%2Fvault%2Ftls%2Fca.crt&unscoped_q=%2Fvault%2Ftls%2Fca.crt

look at all the references to that path. I have everything setup as that path, so using 0.5 all of my deployments failed.

bonifaido:
I have added some code paths to handle this situation more gracefully (instead of introducing another breaking change ;) )
If there is a `vault-tls` volume mount already, we use the `vault-env-tls` volume mount name and mountPath, so all the old way and the new way will work again.

bonifaido:
@jurgenweber please check this version the transit unseal scenario works again, hopefully, your deployments are too.

cmd/vault-secrets-webhook/main.go
```diff
@@ -83,6 +83,17 @@ type mutatingWebhook struct {
 	k8sClient *kubernetes.Clientset
 }
 
+// If the original Pod contained a Volume "vault-tls", for example Vault instances provisioned by the Operator
+// we need to handle that edge case and choose another name for the vault-tls volume for accessing Vautl with TLS.
```
baluchicken:
```suggestion
// we need to handle that edge case and choose another name for the vault-tls volume for accessing Vault with TLS.
```

bonifaido:
https://github.com/banzaicloud/bank-vaults/releases/tag/0.5.1-rc.1

jurgenweber:
> I have added some code paths to handle this situation more gracefully (instead of introducing another breaking change ;) )
> If there is a `vault-tls` volume mount already, we use the `vault-env-tls` volume mount name and mountPath, so all the old way and the new way will work again.

yeah, that was the sort of solution I was going to suggest today if you didn't manage to get around to it.

jurgenweber:
i'll give it a whirl today.

PR #587, opened 2019-08-07
[webhook]: Add support for a default imagePullSecret

mgruener:
**Description:**
Add environment variables to the vault-secrets-webhook that allow the specification of a default imagePullSecret that will be used when the pod that should get secrets injected does not specify an imagePullSecret (or the specified imagePullSecret does not contain matching credentials).

**Use case:**
We use a centralized docker registry to host our images and to proxy requests to upstream registries. There is only one set of credentials that is configured by default on all our kubernetes nodes so that we do not have to hand out the credentials to our users.

By being able to specify a default secret for the vault-secret-webhook, we can keep a setup where our users do not need to know about the credentials of the central docker registry. But the implementation still retains the option to explicitly specify a secret on the pod-to-be-injected if necessary.

### DIFF
--- a/cmd/vault-secrets-webhook/registry/registry.go
+++ b/cmd/vault-secrets-webhook/registry/registry.go
@@ -18,7 +18,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
-	"os"
 	"strings"
 	"sync"
 
@@ -107,12 +106,12 @@ func GetImageConfig(
 func getImageBlob(container ContainerInfo) (*imagev1.ImageConfig, error) {
 	imageName, tag := parseContainerImage(container.Image)
 
-	registrySkipVerify := os.Getenv("REGISTRY_SKIP_VERIFY")
+	registrySkipVerify := viper.GetBool("registry_skip_verify")
 
 	var hub *registry.Registry
 	var err error
 
-	if registrySkipVerify == "true" {
+	if registrySkipVerify {
 		hub, err = registry.NewInsecure(container.RegistryAddress, container.RegistryUsername, container.RegistryPassword)
 	} else {
 		hub, err = registry.New(container.RegistryAddress, container.RegistryUsername, container.RegistryPassword)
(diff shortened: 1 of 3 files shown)


cmd/vault-secrets-webhook/registry/registry.go
```diff
@@ -230,29 +247,33 @@ func (k *ContainerInfo) Collect(container *corev1.Container, podSpec *corev1.Pod
...
 		if found {
 			break
 		}
 	}
 
+	// if we did not find a matching imagePullSecret on the pod itself,
+	// try to find matching registry credentials in the default imagePullSecret
+	// if one was provided
+	if !found {
+		defaultImagePullSecret := os.Getenv("DEFAULT_IMAGE_PULL_SECRET")
```
bonifaido: Could you please use the viper library for settings as in https://github.com/banzaicloud/bank-vaults/blob/4c71e4697b5c3a20c36e02e7dec9298802ad2fd5/cmd/vault-secrets-webhook/main.go#L819?

bonifaido: You can set empty values as defaults.

mgruener: Sure!
What about https://github.com/banzaicloud/bank-vaults/blob/bb13557a1f4d78c9b92bd42116e84400945f0e8e/cmd/vault-secrets-webhook/registry/registry.go#L109

Mind if I change this too, while I am at it? It is not related to my actual code but I was using this line as a reference on how to retrieve values from the environment. Or should I create a new PR for this?

The webhook does not specify any EnvPrefix, right?

bonifaido: No, please go ahead in the same PR. 👍 

Yes, the EnvPrefix is not used currently.

mgruener: Done

bonifaido: Excellent.

cmd/vault-secrets-webhook/registry/registry.go
```diff
@@ -230,29 +247,33 @@ func (k *ContainerInfo) Collect(container *corev1.Container, podSpec *corev1.Pod
...
 			break
 		}
 	}
 
+	// if we did not find a matching imagePullSecret on the pod itself,
+	// try to find matching registry credentials in the default imagePullSecret
+	// if one was provided
+	if !found {
+		defaultImagePullSecret := viper.GetString("default_image_pull_secret")
+		defaultImagePullSecretNamespace := viper.GetString("default)image_pull_secret_namespace")
```
mgruener: I will have to fix this before the PR gets merged.

PR #592, opened 2019-08-08
feat: add servicemonitor config to the webhook chart

jurgenweber:
and we have metrics in prom;

https://take.ms/TjvS5

### DIFF
--- a/charts/vault-secrets-webhook/values.yaml
+++ b/charts/vault-secrets-webhook/values.yaml
@@ -28,10 +28,10 @@ env:
   # DEFAULT_IMAGE_PULL_SECRET_NAMESPACE:
 
 metrics:
-  enabled: true
+  enabled: false
   port: 8443
   serviceMonitor:
-    enabled: true
+    enabled: false
     scheme: https
     tlsConfig:
       insecureSkipVerify: true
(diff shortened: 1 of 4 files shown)


charts/vault-secrets-webhook/values.yaml
```diff
@@ -27,6 +27,14 @@ env:
   # DEFAULT_IMAGE_PULL_SECRET:
   # DEFAULT_IMAGE_PULL_SECRET_NAMESPACE:
 
+metrics:
+  enabled: true
```
bonifaido: I would make this `false` by default since I'm not sure all the people have the ServiceMonitor CRD type in their clusters.

jurgenweber: yeah, ok

jurgenweber: those people be crazy. :)

jurgenweber:
I found this generic dashboard; https://grafana.com/grafana/dashboards/240 worth putting the link somewhere?

bonifaido:
Maybe to here https://github.com/banzaicloud/bank-vaults/tree/master/docs/monitoring ?

PR #594, opened 2019-08-08
[webhook,operator] Webhook docker credentials auth + documentation update

mgruener:
**Add support for the `auth` field in docker credentials**
Instead of solely relying on the `username` / `password` fields in the `imagePullSecrets` for a given registry, fall back to the `auth` field if necessary.

I did some research and found no conclusive information if any of the fields are "required". https://github.com/moby/moby/blob/master/api/types/auth.go lists all fields as "omitempty". Doing a `docker login` with a recent version of docker creates a config.json only containing the `auth` field. Using `kubectl create secret docker-registry` creates a secret containing all 3 fields.

**Update webhook documentation**
Add a reference to the default imagePullSecrets to the webhook documentation

**Update operator transit unseal example**
From what I see in other examples e.g. https://github.com/banzaicloud/bank-vaults/blob/73c7c5bff7ebcdefec89e640eeb725d99015ce87/operator/deploy/cr.yaml#L123-L126 and my own experiments the ServiceAccount used for the vault-secrets-webhook also needs access to the unseal vault to at least retrieve the `VAULT_TOKEN: "vault:login"`

### DIFF
--- a/cmd/vault-secrets-webhook/registry/registry.go
+++ b/cmd/vault-secrets-webhook/registry/registry.go
@@ -15,6 +15,7 @@
 package registry
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


bonifaido:
@mgruener thanks for the extensions, have you considered this commented out section (I started to implement it, but haven't finished it):

https://github.com/banzaicloud/bank-vaults/blob/master/cmd/vault-secrets-webhook/registry/registry.go#L247

mgruener:
@bonifaido Yes, I have seen the commented out code and the TODO but that is more related to #587 than to this PR, right?

From how I understand it, this would be a third method to retrieve the registry credentials, the other two being the pod imagePullSecrets and the webhook default secret (implemented in the aforementioned PR).

It shouldn't be so hard to implement this and I can give it a shot if you want but it would be nice if this was not tied to acceptance of this current PR :smile:

PR #595, opened 2019-08-08
webhook: expose mutation metrics in prometheus format

bonifaido:
Fixes: #593 

The Dashboard is not ready yet! ;)


cmd/vault-secrets-webhook/main.go
```diff
@@ -863,6 +865,8 @@ func init() {
 	viper.SetDefault("vault_ignore_missing_secrets", "false")
 	viper.SetDefault("vault_env_passthrough", "")
 	viper.SetDefault("mutate_configmap", "false")
+	viper.SetDefault("tls_cert_file", "")
+	viper.SetDefault("tls_private_key_file", "")
```
jurgenweber: should these be added as env vars?

jurgenweber: ignore that. I am confused. :)

PR #597, opened 2019-08-08
[webhook] Small logging / comment additions

mgruener:
**Description**
Add a little bit of logging to the imagePullSecret secret selection and remove the TODO about pulling imagePullSecrets from ServiceAccounts (see explanation in the comments to this PR).

### DIFF
--- a/cmd/vault-secrets-webhook/registry/registry.go
+++ b/cmd/vault-secrets-webhook/registry/registry.go
@@ -276,11 +276,12 @@ func (k *ContainerInfo) Collect(container *corev1.Container, podSpec *corev1.Pod
 
 	k.Image = k.fixDockerHubImage(container.Image)
 
-	// k.clientset.Core().ServiceAccounts(k.Namespace).Get(podSpec.ServiceAccountName)
-
-	// TODO read ServiceAccount's imagePullSecrets as well
 	var err error
 	found := false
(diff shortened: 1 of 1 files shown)


mgruener:
@bonifaido I just tested this and I think this is unnecessary. A pod that uses a ServiceAccount that has  an imagePullSecret attached, will have the imagePullSecret injected automatically. So from the point of view of the vault-secrets-webhook, the imagePullSecret is always attached to the pod, no need to handle this specific situation in the vault-secrets-webhook code.

I just checked the official documentation for this: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account

>Now, any new pods created in the current namespace will have this added to their spec:
> ```
> spec:
>  imagePullSecrets:
>  - name: myregistrykey
>```

Did I misinterpret your intentions in regard to this feature?

bonifaido:
In that case, we are fine, which is great! TBH I didn't know that this happens automatically. Sorry for that!

mgruener:
@bonifaido Neither I nor my colleagues did know that it works like this :smile: I have removed the logic dealing with the service accounts and just left the small additions to the logging in. I also clarified the situation with regard to the ServiceAccount in the comments.

I just want to test the changes in our setup and if everything works, I will mark this ready for review.

Issue #521, opened 2019-06-17
consul-template sidecar: missing client token


devlounge:
**Describe the bug**

Unable to make the consul-template sidecar work, following the example
The main container can read data populated by vault-env and stored as env vars, but the consult-template container complains about "missing client token"

versions: 0.4.16 & 0.4.17-rc.4

I keep getting:

```
kubectl -n banzaicloud-poc  log -f --all-containers pod/hello-secrets-55c7c65d48-zvs58
==> Vault server started! Log data will stream in below:

==> Vault agent configuration:

                     Cgo: disabled
               Log Level: info
                 Version: Vault v1.1.3
             Version Sha: 9bc820f700f83a7c4bcab54c5323735a581b34eb

2019-06-17T16:16:33.719Z [INFO]  sink.file: creating file sink
2019-06-17T16:16:33.719Z [INFO]  sink.file: file sink configured: path=/vault/.vault-token
2019-06-17T16:16:33.720Z [INFO]  auth.handler: starting auth handler
2019-06-17T16:16:33.720Z [INFO]  auth.handler: authenticating
2019-06-17T16:16:33.720Z [INFO]  sink.server: starting sink server
2019-06-17T16:16:33.782Z [INFO]  auth.handler: authentication successful, sending token to sinks
2019-06-17T16:16:33.782Z [INFO]  auth.handler: starting renewal process
2019-06-17T16:16:33.782Z [INFO]  sink.file: token written: path=/vault/.vault-token
2019-06-17T16:16:33.782Z [INFO]  sink.server: sink server stopped
2019-06-17T16:16:33.782Z [INFO]  sinks finished, exiting
2019/06/17 16:16:35.660281 [WARN] (clients) disabling vault SSL verification
2019/06/17 16:16:35.669231 [WARN] (view) vault.read(default/data/workloads/hello-secrets-dynamic): vault.read(default/data/workloads/hello-secrets-dynamic): Error making API request.

URL: GET https://cluster-vault.vault:8200/v1/default/data/workloads/hello-secrets-dynamic
Code: 400. Errors:

* missing client token (retry attempt 1 after "250ms")
2019/06/17 16:16:35.921565 [WARN] (view) vault.read(default/data/workloads/hello-secrets-dynamic): vault.read(default/data/workloads/hello-secrets-dynamic): Error making API request.

URL: GET https://cluster-vault.vault:8200/v1/default/data/workloads/hello-secrets-dynamic
Code: 400. Errors:

[...]

PR #600, opened 2019-08-09
webhook: runAsUser not needed

### LINKED ISSUE #599: [vault-secret-webhook] vault-agent container with runAsUser 100 fails to spawn on OpenShift
**Describe the bug**:
In OpenShift the vault-secret-webhook injects the init-container vault-agent with the securityContext "runAsUser: 100".

This is by default not allowed and you have to set the scc (security context contraint) anyuid to the service account which runs the pod.

**Expected behaviour**:
The webhook should respect the uid range of the namespace or the security context "runAsUser" should be removed.

- Install the Webhook
- Deploy the sample application from documentation

**Additional context**:
The runAsUser part is hard-coded in file "cmd/vault-secrets-webhook/main.go" at line 120.

**Environment details**:
- OpenShift 3.11
- helm template ... | oc apply -f -

/kind bug

aidun:
This PR reference an Issue occurs on OpenShift, fixes #599


aidun:
> Can we in case make this optional rather than just remove it ?

Of course. I already viewed at it. I would implement it, as the other options: With Viper and an annotation.

Can you explain me why the runasuser part is important? I like to understand the advantages in this case.

primeroz:
Replying from my phone so can't really look around that much but if I am not mistaken this defines the container settings for vault agent , in a cluster / namespace that does not allow running as user 0 we need to set this.

aidun:
I thought about it a bit and looked in the code again. The other containers which are injected, does not bekomm runAsUser: 100. Maybe I should add it to all injected containers?

cmd/vault-secrets-webhook/main.go
```diff
@@ -99,6 +101,18 @@ func hasTLSVolume(volumes []corev1.Volume) bool {
 func getInitContainers(originalContainers []corev1.Container, vaultConfig vaultConfig, initContainersMutated bool, containersMutated bool, containerEnvVars []corev1.EnvVar, containerVolMounts []corev1.VolumeMount) []corev1.Container {
 	var containers = []corev1.Container{}
 
+	var securityContext *corev1.SecurityContext
+	if vaultConfig.securityContextRunAsUser {
```
primeroz: do we need 2 options here ? One to enable the `runAsUser` and one to set it to a value ? 

I think the default should respect what was done so far, so set it to `100`, if set to `0` it could revert to your _else_ case where no `runAsUser` is set

I can see you are already doing this ( Setting default to _true_ and value to _100_ ) so maybe we can shorten the amount of "options / annotations" required ?

primeroz:
> I thought about it a bit and looked in the code again. The other containers which are injected, does not bekomm runAsUser: 100. Maybe I should add it to all injected containers?

Good point.
Which injected container are you talking about ? I guess "vault-env" and "consul-template" 

* vault-env builds on top of the vault image ( since i think it will use vault-agent to grab the actual token  ? )  , by default it runs as root, does some permissions changes,  and drop privileges to vault user through docker-entrypoint  - We need to make sure it keeps working when changed to user 100 from this change ( i think it should with no issues )

* consult-template - will try to change directory permissions on startup to 100:1000 before dropping privileges to the user, so this as well need to make sure that it keeps working as before. 
There were some discussions about uid for consul-template but can't remember exactly what the point was, @jurgenweber might know more about it.
see [this](https://github.com/banzaicloud/bank-vaults/issues/521#issuecomment-503337936) 


@bonifaido do we have enough checks in the current acceptance-test.sh to ensure that all the webhook different options ( vault-env, consul-template) to make sure it will keep working seamlessly with such a change ?

cmd/vault-secrets-webhook/main.go
aidun: Okay, i will give it a try.

bonifaido:
Unfortunately, the consul-template code paths are not tested in `acceptance-test.sh`, but it is a very nice (and not extremely hard) challenge to add those :)

aidun:
I talk about vault-env and the copy container. The consul-template container was not in my scope.

primeroz:
> I talk about vault-env and the copy container. The consul-template container was not in my scope.

my bad, i can see now that the consul-template security context is not affected by the change so that should be ok

cmd/vault-secrets-webhook/main.go
primeroz: looks good

primeroz:
@aidun looks good and CI is passing , using default values. Would you be able to extend the acceptance test to check the feature is working as expected ?

aidun:
@primeroz I added some test, with the new annotation. Is that how you imagined it?

hack/acceptance-test.sh
```diff
@@ -80,9 +83,24 @@ helm install ./charts/vault-secrets-webhook \
...
+kubectl delete -f deploy/test-secret.yaml
+kubectl delete -f deploy/test-deployment.yaml
+
+
+
+kubectl apply -f deploy/test-secret-runasuser.yaml
+test `kubectl get secrets sample-secret -o jsonpath='{.data.\.dockerconfigjson}' | base64 --decode | jq -r '.auths[].username'` = "dockerrepouser" 
+test `kubectl get secrets sample-secret -o jsonpath='{.data.\.dockerconfigjson}' | base64 --decode | jq -r '.auths[].password'` = "dockerrepopassword"
+
+kubectl apply -f deploy/test-deployment.yaml
```
primeroz: typo : _deploy/test-deployment-runasuser_

hack/acceptance-test.sh
```diff
@@ -80,9 +83,24 @@ helm install ./charts/vault-secrets-webhook \
...
+
+# default
 kubectl apply -f deploy/test-secret.yaml
 test `kubectl get secrets sample-secret -o jsonpath='{.data.\.dockerconfigjson}' | base64 --decode | jq -r '.auths[].username'` = "dockerrepouser"
 test `kubectl get secrets sample-secret -o jsonpath='{.data.\.dockerconfigjson}' | base64 --decode | jq -r '.auths[].password'` = "dockerrepopassword"
 
 kubectl apply -f deploy/test-deployment.yaml
 kubectl wait --for=condition=available deployment/hello-secrets --timeout=120s
+
+# cleanup
```
primeroz: Unfortunately because of the way the verification is done ( Through the _finish_ [function trapped _on exit_ ](https://github.com/banzaicloud/bank-vaults/blob/4ffe5ed082610cde8433784a239a2397349381f6/hack/acceptance-test.sh#L30) ) this won't work 

Can you please , instead of using the same name for deployment and secret and having to delete the resources, use different names for your resources and the one without run-as-user and add a check in the _finish_ function  ? 

Would be great to verify that the UID was actually set properly on the pod but i don't know if that information is exposed on the pod , in case just pull that field out of the deployment and test is set as expected.

deploy/test-secret-runasuser.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: sample-secret
+  annotations:
+    vault.security.banzaicloud.io/vault-addr: "https://vault.default.svc.cluster.local:8200"
+    vault.security.banzaicloud.io/vault-role: "default"
+    vault.security.banzaicloud.io/vault-skip-verify: "true"
+    vault.security.banzaicloud.io/vault-path: "kubernetes"
+    vault.security.banzaicloud.io/security-context-run-as-user-id: "0"
```
primeroz: Don't think we need it on the secret since there is no injection involved here and the webhook itself will mutate the secret , right @bonifaido

bonifaido: Yes, that is true, no runtime behavior is associated with Secret/ConfigMap mutation outside of the webhook.

jurgenweber:
before this is merged I will need to test this, that runAsUser is there for a very specific reason and there is a high probability you will just fix it for you and break it for everyone else.

deploy/test-secret-runasuser.yaml
aidun: I will remove it from the tests.

hack/acceptance-test.sh
aidun: I will implement it this afternoon. Should be no Problem.

aidun:
@jurgenweber I think I haven't changed the current behavior and only made it possible to deploy the webhook on OpenShift. This use case is very specific to OpenShift, but on OpenShift the default :/

primeroz:
> before this is merged I will need to test this, that runAsUser is there for a very specific reason and there is a high probability you will just fix it for you and break it for everyone else.
> 
> So please ensure you maintain current behaviour and manipulate it for your edge case.

Please test it if you get a chance. I am not 100% familiar with the webhook code since i am not running it yet. 

I do believe that this change only affect the `vault agent init container` and `vault-env` keeping the default behavior as it is now. 
`Consul-template` container should not be affected since the security context for that container is defined in [its own function](https://github.com/banzaicloud/bank-vaults/blob/4ffe5ed082610cde8433784a239a2397349381f6/cmd/vault-secrets-webhook/main.go#L178)

hack/acceptance-test.sh
aidun: Everything except the uid verification is in the PR.

cmd/vault-secrets-webhook/main.go
```diff
@@ -862,6 +874,8 @@ func init() {
 	viper.SetDefault("vault_agent", "false")
 	viper.SetDefault("vault_ct_share_process_namespace", "")
 	viper.SetDefault("psp_allow_privilege_escalation", "false")
+	viper.SetDefault("security_context_run_as_user", "true")
```
primeroz: just noticed we still have this, we don't need it anymore

bonifaido:
CircleCI is back, finally the test can excercise this!

primeroz:
The CI now failed because of "resource requests" and amount of cpu on the node used for testing

@bonifaido I would dare say that starting all workloads and checking in the `on_finish` is not scalable and might need some re-thought going forward

@aidun For the time being rather than check in the `on_finish` it might be better to start the workloads, check, cleanup the workload

bonifaido:
Hi, yes it is not scalable, it is mostly for debugging purposes. What we could do here is to remove the deployments after they went successful to spare some resources, so, for example, we can remove the first like this:

```bash
kubectl apply -f deploy/test-deployment.yaml
kubectl wait --for=condition=available deployment/hello-secrets --timeout=120s
+ kubectl delete -f deploy/test-deployment.yaml
```

Anyhow the test needs to be reworked, but of course not in this PR.

primeroz:
yeah, lets test what we want before the delete for now instead of using the `on_finish`

hack/acceptance-test.sh
```diff
@@ -80,9 +89,17 @@ helm install ./charts/vault-secrets-webhook \
...
+
+# default
 kubectl apply -f deploy/test-secret.yaml
 test `kubectl get secrets sample-secret -o jsonpath='{.data.\.dockerconfigjson}' | base64 --decode | jq -r '.auths[].username'` = "dockerrepouser"
 test `kubectl get secrets sample-secret -o jsonpath='{.data.\.dockerconfigjson}' | base64 --decode | jq -r '.auths[].password'` = "dockerrepopassword"
 
 kubectl apply -f deploy/test-deployment.yaml
 kubectl wait --for=condition=available deployment/hello-secrets --timeout=120s
+
+kubectl apply -f deploy/test-deployment-runasuser-0.yaml
```
bonifaido: Could you please delete the deployment after it is checked, so we don't run out of resources.

hack/acceptance-test.sh
```diff
@@ -80,9 +89,17 @@ helm install ./charts/vault-secrets-webhook \
...
+kubectl wait --for=condition=available deployment/hello-secrets-runasuser-0 --timeout=120s
+
+kubectl apply -f deploy/test-deployment-runasuser-101.yaml
+kubectl wait --for=condition=available deployment/hello-secrets-runasuser-101 --timeout=120s
```
bonifaido: And here as well delete please.

aidun:
Sorry I haven't contacted you. the last few days have been a little stressful. 

i've been working with Kubernetes / OpenShift, RBAC, PSP and so on. Some questions arose concerning the implementation and I'm not sure anymore if these are cleanly implementable under OpenShift. For me there were two problems:

The way the containers / init containers are added always requires that the service account under which the init container is started must have extended privileges. Unfortunately I haven't found a reason why this causes problems under OpenShift or why it is necessary for Kubernetes.

Another point is, the required security constraints under OpenShift  require that the executing service account runs with more rights than necessary. Unfortunately, it cannot be restricted to the injected containers. For the Init containers I see no reason to set special SecurityContexts that were not set by the Pod.

At first I thought it could be simplified and reduced to a minimum. I'd like to get more familiar with the topic and analyze it before I jump to conclusions and waste your time with reviews.

For this I have to build a suitable test setup to understand the exact differences between OpenShift SCC and Kubernetes PSP. I'm also looking for a better way to debug the webhook to get to the created but not logged manifest of the init-container.
Until then I would like to withdraw the PR, because it does not lead to the desired goal, the OpenShift support.

But there is nothing to prevent a merge, if it helps the project.

matyix:
Hello @aidun, sorry to hear about your struggle with OpenShift - unluckily, though they claim to be compatible with upstream Kubernetes by far it is not the case. Its is really a shame, but the main reason they do thinks differently are: 

1. lots of legacy code/architecture back from the OpenStack days
2. **vendor lock-in**

@pbalogh-sa will deep dive into the PR in more details and check what and how would make sense to lift from your PR, or propose some alternative solutions. 

As always, we appreciate the time you put into the PR and willingness to contribute back, sorry to learn about the OpenShift issues (lucky us, we interact with OpenShift only when we migrate customers over to [PKE](https://github.com/banzaicloud/pke) and [Pipeline](https://github.com/banzaicloud/pipeline)).

hack/acceptance-test.sh
```diff
@@ -80,9 +79,38 @@ helm install ./charts/vault-secrets-webhook \
...
 
 kubectl apply -f deploy/test-deployment.yaml
 kubectl wait --for=condition=available deployment/hello-secrets --timeout=120s
+
+kubectl describe deployment/hello-secrets
+kubectl describe rs hello-secrets
+kubectl describe pod hello-secrets
+kubectl logs deployment/hello-secrets --all-containers
+kubectl get secret -n vswh -o yaml
+kubectl delete -f deploy/test-deployment.yaml
```
pbalogh-sa: These lines are found in finish function, so acceptance test fail due to deleting hello-secrets deployment in this section.

pbalogh-sa:
@aidun do you have any plan with this?

pbalogh-sa:
fixed by #698

PR #603, opened 2019-08-13
webhook/configmap: mutate BinaryData as well

rrondeau:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

This PR adds the possibility to mutate the binary data of configmaps.
Binary data must be base64 encoded inside vault (https://github.com/hashicorp/vault/issues/1423)

### Why?

We need to mutate configMaps with binary data

### Additional context

This only work with base64 encoded data inside vault since vault cannot store binary data without encoding

### DIFF
--- a/cmd/vault-secrets-webhook/configmap.go
+++ b/cmd/vault-secrets-webhook/configmap.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"encoding/base64"
 	"fmt"
 	"strings"
 
@@ -25,6 +26,11 @@ import (
 func configMapNeedsMutation(configMap *corev1.ConfigMap) bool {
 	for _, value := range configMap.Data {
 		if strings.HasPrefix(value, "vault:") {
+			return true
+		}
+	}
+	for _, value := range configMap.BinaryData {
+		if strings.HasPrefix(string(value), "vault:") {
 			return true
 		}
 	}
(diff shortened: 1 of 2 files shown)


rrondeau:
Hi
I put an example into the deploy directory like you ask.
If you want some other docs / example, i will be happy to oblige.

bonifaido:
Thanks @rrondeau , we are having some Circle CI seats issue right now, under investigation, thanks!

deploy/test-configmap.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: configMap
```
pepov:
```suggestion
kind: ConfigMap
```

cmd/vault-secrets-webhook/configmap.go
```diff
@@ -28,6 +29,11 @@ func configMapNeedsMutation(configMap *corev1.ConfigMap) bool {
 			return true
 		}
 	}
+	for _, value := range configMap.BinaryData {
+		if strings.HasPrefix(string(value), "vault:") {
```
pepov: binaryData is automatically base64 decoded here?

deploy/test-configmap.yaml
rrondeau: Yep sorry for the typo

cmd/vault-secrets-webhook/configmap.go
rrondeau: yes like secrets data

bonifaido:
CircleCI has deployed a fix to resolve our issues, thanks for your patience, and for the contribution of course!

PR #604, opened 2019-08-14
[config] Add support for auth mount config

mgruener:
| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?
Add support for the mount options described at https://www.vaultproject.io/api/system/auth.html#config

This adds a new "options" field to the auth method configuration:

```
auth:
  - type: ""
    path: ""
    options:
      listing_visibility: "unauth"
      ...
    config: {}
    roles: []
```

### Why?
The main reason to implement this was the ability to set the listing_visibility option for authentication methods to configure which auth methods will be listed in the web ui.

### Checklist

I am not sure about the error handling guideline, I followed the style I found in the file I modified itself.

### DIFF
--- a/pkg/vault/operator_client.go
+++ b/pkg/vault/operator_client.go
@@ -479,19 +479,47 @@ func (v *vault) configureAuthMethods(config *viper.Viper) error {
 			}
 		}
 
+		// get auth mount options
+		// https://www.vaultproject.io/api/system/auth.html#config
+		var authConfigInput api.AuthConfigInput
+		var hasMountOptions bool
+		if _, hasMountOptions = authMethod["options"]; hasMountOptions {
+			err = mapstructure.Decode(authMethod["options"], &authConfigInput)
+			if err != nil {
+				return fmt.Errorf("error parsing auth method options: %s", err.Error())
+			}
+		}
+
 		if !exists {
 			logrus.Debugf("enabling %s auth backend in vault...", authMethodType)
 
 			// https://www.vaultproject.io/api/system/auth.html
-			options := api.EnableAuthOptions{
-				Type:        authMethodType,
-				Description: description,
+			var options api.EnableAuthOptions
+			if hasMountOptions {
+				options = api.EnableAuthOptions{
+					Type:        authMethodType,
+					Description: description,
+					Config:      authConfigInput,
+				}
+			} else {
+				options = api.EnableAuthOptions{
+					Type:        authMethodType,
+					Description: description,
+				}
 			}
 
 			err := v.cl.Sys().EnableAuthWithOptions(path, &options)
 
 			if err != nil {
 				return fmt.Errorf("error enabling %s auth method for vault: %s", authMethodType, err.Error())
+			}
+		} else if hasMountOptions {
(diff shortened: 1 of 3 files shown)


bonifaido:
Thanks @mgruener, we are having some troubles with our CircleCI builds, please say tuned 🤞

pkg/vault/operator_client.go
```diff
@@ -479,20 +479,48 @@ func (v *vault) configureAuthMethods(config *viper.Viper) error {
...
+
 		if !exists {
 			logrus.Debugf("enabling %s auth backend in vault...", authMethodType)
 
 			// https://www.vaultproject.io/api/system/auth.html
-			options := api.EnableAuthOptions{
-				Type:        authMethodType,
-				Description: description,
+			var options api.EnableAuthOptions
+			if hasMountOptions {
```
primeroz: should not this `if !hasMountOptions` to keep the old behavior if options are not defined _else_ set the _Config_ key to the _authConfigInput_ extracted at [line 487](https://github.com/banzaicloud/bank-vaults/pull/604/files#diff-c97c1ac6b5509cf75c39cdc58bcb9dfaR487) ?

primeroz: should not this be `if !hasMountOptions` to keep the old behavior if options are not defined _else_ set the `Config` key to the `authConfigInput` extracted at [line 487](https://github.com/banzaicloud/bank-vaults/pull/604/files#diff-c97c1ac6b5509cf75c39cdc58bcb9dfaR487) ?

mgruener: You are absolutely right, thanks.

primeroz:
Great, last thing we need for this is an example / documentation . 

@bonifaido we should really get a section in the https://github.com/banzaicloud/bank-vaults/docs/ to show all options for secrets and backend so that we don't need to keep adding examples in the deploy directory for every single change

for now should we just get something in https://github.com/banzaicloud/bank-vaults/tree/master/operator/deploy ?

bonifaido:
Yes please, and I absolutely agree (however maintaining a set of examples and a reference documentation increases the amount of effort, but we can't spare that). We should find the time to extract all those examples to reference documentation in a set of different PRs.

bonifaido:
We still struggle with CircleCI support to get the builds back running, but I will test this change manually.

bonifaido:
@mgruener would you mind adding an example for this to vault-config.yaml?

mgruener:
@bonifaido If have added my usecase as an example to the vault-config.yaml and the cli-tool documentation where I think it is most applicable.

bonifaido:
Great job @mgruener Thanks!

primeroz:
LGTM . just waiting on the CI fix to make sure it is all clean before approving.

matyix:
@mgruener @primeroz sorry about this CircleCI inconvenience - apparently (based on the conversation we had with them) you need to modify the last commit, otherwise will fail - thus rerunning the build will not help. They need a new commit ID. I suggest to amend the latest commit, change the commit message, and force push. 

Sorry again for the inconvenience.

PR #606, opened 2019-08-14
[helm vault-secrets-webhook] Allow setting pod annotations

chrisob:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Allow setting pod annotations for the vault-secrets-webhook Helm chart from values.yaml.

### Why?
Currently there is no way to set pod annotations for the webhook deployment. 

This is helpful to have in many setups, for example: In a cluster where Istio mTLS is required, the readiness probes for the webhook's deployment will fail as the containerPort is intercepted by the Envoy sidecar. Adding the pod annotation `sidecar.istio.io/rewriteAppHTTPProbers: "true"` fixes the probe.

### Additional context

Tested deployment with both default values, as well as values override for `podAnnotations`.

I bumped the minor chart version as IMO this introduces a new feature as described by SemVer, but if desired I can also just bump the patch version.

### Checklist

### To Do

### DIFF
--- a/charts/vault-operator/Chart.yaml
+++ b/charts/vault-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 0.5.0
 description: A Helm chart for banzaicloud/bank-vaults operator
 name: vault-operator
-version: 0.3.0
+version: 0.4.0
 home: https://www.vaultproject.io/
 icon: https://www.vaultproject.io/assets/images/mega-nav/logo-vault-0f83e3d2.svg
 sources:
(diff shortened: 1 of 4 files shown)


bonifaido:
I changed the chart version of the webhooks and the operator back to the original, just releasing it :)

chrisob:
@bonifaido wow I'm a total idiot, I just realized I incremented the wrong chart version :man_facepalming:

chrisob:
Welp, that's one reason to not have auto-builds/publishing of Helm charts!

bonifaido:
This is not publishing the chart until we assign a tag specific chart so we captured it in the right phase :)

bonifaido:
Released in 0.4.4.

PR #633, opened 2019-08-23
feat: use recommended k8s labels

### LINKED ISSUE #631: Use recommended K8s labels rather than custom
Kubernetes has a recommended set of labels for applications. At minimum, the vault webhook helm cart is not using them.

See https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Use the recommended labels as is being adopted by many helm charts.

nesl247:
This changes to the recommended k8s labels as documented in #631. This PR was based upon #632 as to not cause conflicts. I will rebase this PR once that is merged in, or vice versa depending on what you ask for.

Closes #631

### DIFF
--- a/charts/vault-operator/templates/crd.yaml
+++ b/charts/vault-operator/templates/crd.yaml
@@ -3,10 +3,10 @@ kind: CustomResourceDefinition
 metadata:
   name: vaults.vault.banzaicloud.com
   labels:
-    app: {{ include "vault-operator.name" . }}
-    chart: {{ include "vault-operator.chart" . }}
-    release: {{ .Release.Name }}
-    heritage: {{ .Release.Service }}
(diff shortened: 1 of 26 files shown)


bonifaido:
We have one conflict, but looks easy.

chrisob:
@nesl247 @bonifaido Why no version bump in Chart.yaml? This overrides the 0.5.1 chart version, and actually breaks `helm install`s which _should_ be no-ops with the following error:

```Error: Deployment.apps "vault-secrets-webhook" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"vault-secrets-webhook", "app.kubernetes.io/name":"vault-secrets-webhook"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable

PR #638, opened 2019-08-26
[webhook] Serving telemetry on another address avoid TLS problems

leominov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | yes

### What's in this PR?

Serving vault-secrets-webhook telemetry on another address avoid TLS problems.

### Why?

Without changes we can't collect metrics via Prometheus. It's my fault, sorry.

### Checklist

### To Do


bonifaido:
What kind of TLS problems do you have with the current setup?

leominov:
Prometheus haven't some sort of insecure options per target, so vault-secrets-webhook is always down for him. We can only change `__address__ `, `__scheme__ ` and `__metrics_path__ ` via Kubernetes annotations.

leominov:
I saw sources of `kubernetes/autoscaler` and `istio/gallery`, looks like it's common practice to separate this things.

bonifaido:
I understand your issue, TLS can be configured on the Prometheus scrape job level only. Can you try that?

Right now we use the Prometheus Operator ServiceMonitor configuration to get our metric targets as in here: https://github.com/banzaicloud/bank-vaults/blob/master/charts/vault-secrets-webhook/templates/prometheus-monitorservice.yaml which use TLS by default (which can be defined on the Service level here). So this change would break the current behavior. Can we make this configurable so start on HTTP (instead of HTTPS) only when defined so?

leominov:
You absolutely right, but making separated jobs for every service that we use it's not an option. For now we don't use any operator for Prometheus, so we configure it with `kubernetes_sd_config` for all pods in cluster.

So yes, I can make it configurable and we discuss about it later.

bonifaido:
So I guess you don't use the Helm chart? :)

leominov:
Nope :) Is it problem?

bonifaido:
No, absolutely not, was just a guess!

leominov:
Thanks, I think my next step will be adding new metrics :D

PR #641, opened 2019-08-27
Add postgresql as an allowed HA backend

### LINKED ISSUE #639: Postgres isn't supported as an HA backend
If I add a storage stanza for `postgres` and include `ha_enabled: "true"`, the operator complains when the stateful set is > 1.

The current list of HA backends does not include `postgres`.

Since I wasn't sure how the list is managed, I wanted to verify that a PR to include `postgres` is acceptable.

My current workaround is to also provide `ha_storage` which gets around the issue right now.

/kind bug

fraenkel:
Allow PostgreSQL to be a valid HA backend

Fixes #639


bonifaido:
Thanks @fraenkel

PR #642, opened 2019-08-28
[charts/webhook] Make listen address configurable

leominov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Webhook internal port did not pass as environment variable into container.

### Why?

Webhook always using default port instead port from configuration.

### Checklist

### DIFF
--- a/charts/vault-secrets-webhook/templates/webhook-deployment.yaml
+++ b/charts/vault-secrets-webhook/templates/webhook-deployment.yaml
@@ -50,7 +50,7 @@ spec:
             - name: TLS_PRIVATE_KEY_FILE
               value: /var/serving-cert/tls.key
             - name: LISTEN_ADDRESS
-              value: {{ .Values.service.internalPort | quote }}
+              value: ":{{ .Values.service.internalPort}}"
             - name: DEBUG
               value: {{ .Values.debug | quote }}
             {{- range $key, $value := .Values.env }}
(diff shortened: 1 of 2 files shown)


charts/vault-secrets-webhook/templates/webhook-deployment.yaml
```diff
@@ -49,6 +49,8 @@ spec:
               value: /var/serving-cert/tls.crt
             - name: TLS_PRIVATE_KEY_FILE
               value: /var/serving-cert/tls.key
+            - name: LISTEN_ADDRESS
+              value: {{ .Values.service.internalPort }}
```
bonifaido: The port should be quoted with `| quote }}`. Please bump the chart version as well.

leominov: Ok

charts/vault-secrets-webhook/templates/webhook-deployment.yaml
```diff
@@ -49,6 +49,8 @@ spec:
               value: /var/serving-cert/tls.crt
             - name: TLS_PRIVATE_KEY_FILE
               value: /var/serving-cert/tls.key
+            - name: LISTEN_ADDRESS
+              value: {{ .Values.service.internalPort | quote }}
```
bonifaido: I just checked CI and it fails with the webhook not coming up, since listen-address is not just a port, I guess `":"` has to be added as well: https://github.com/banzaicloud/bank-vaults/blob/master/cmd/vault-secrets-webhook/main.go#L875

leominov: Yes, I forgot about that.

bonifaido:
Just released the new chart version @leominov

leominov:
We don't use Helm, it's for community :)

bonifaido:
Thanks, you're giving back! :)

PR #644, opened 2019-08-29
[charts/webhook] Avoid insecure options in ServiceMonitor

leominov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

For now we can use separated port with telemetry avoid using insecure options. Also I've add default Prometheus annotations.

### Why?

Insecure – is not cool.

### Checklist

### Todo

What type of versioning you are using?


bonifaido:
The PR contains a lot of nice changes, but one thing I can't agree with is removing the TLS termination entirely, and changing it to plain HTTP. Why is this better? :)

leominov:
Heh, I don't understand why it's good – using insecure options in any part of software that created for increase your infrastructure security :)

More easily and universal configuration `kubernetes_sd_config` in case when you are not using CoreOS Prometheus Operator.

bonifaido:
I totally agree with the second part which makes the chart monitorable without the Prometheus operator, but forcing users to transform something from https to http (event if the ca validation is turned off) can't be more secure than it was :)

sagikazarmark:
@leominov insecure here (and in general related to TLS) means skipping CA verification. Without CA verification you can't be sure that a certificate was issued by a trusted issuer, but the connection will still be encrypted. Since in this case you are both the server and the client, you can trust that the self-signed certificate on the server side is valid (eg. you are not a victim of a MITM attack).

If you can't have that guarantee on your private infrastructure, you have much worse issues than skipping CA verification. Also, this is kind of an industry standard practice for internal services where you really don't even have the ability to use services issued by trusted third parties.

leominov:
I think its my mistake that I've add metrics on TLS. So, what you think if I revert ServiceMonitor changes and metrics port for backward compatibilities and only if user set port not the same as internal port I've create non-TLS configuration with `TELEMETRY_LISTER_ADDRESS`?

pbalogh-sa:
@leominov 
> So, what you think if I revert ServiceMonitor changes and metrics port for backward compatibilities and only if user set port not the same as internal port I've create non-TLS configuration with TELEMETRY_LISTER_ADDRESS?

This idea is the right way, please implement it.

leominov:
ok, set wip for now

charts/vault-secrets-webhook/templates/webhook-service.yaml
```diff
@@ -11,10 +11,36 @@ metadata:
...
   selector:
     app.kubernetes.io/name: {{ template "vault-secrets-webhook.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
+{{- if .Values.metrics.enabled }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+  name: {{ template "vault-secrets-webhook.fullname" . }}-metrics
```
primeroz: Do we need this `extra` service for the metrics ? I think we can just add an extra port to the service above `if .Values.metrics.enabled` 

If we instead prefer the extra service i think we need to fix the `serviceMonitor selector` [here](https://github.com/banzaicloud/bank-vaults/blob/4f50e9c74be7d76db3b8772fae13fffab85fa8ba/charts/vault-secrets-webhook/templates/prometheus-monitorservice.yaml#L35) to be more specific and use a set of labels that **only** match the metrics service.

I am not super familiar with helm ( i don't use it ) but maybe should the `app.kubernetes.io/name: {{ template "vault-secrets-webhook.name" . }}` label in this service be `{{ template "vault-secrets-webhook.fullname" . }}-metrics` so that the `serviceMonitor selector` can use this ?

bonifaido: I also vote for using the same existing Service.

leominov:
Back on a week :)

PR #664, opened 2019-09-13
Delete securityContext for copy-vault-env initContainer

### LINKED ISSUE #663: copy-vault-env initContainer do not overwrite securityContext
**Describe the bug**:
copy-vault-env initContainer always try to run as root as its securityContext gets set by the mutating webhook.

**Expected behaviour**:
It should run as the same user than the non-mutated pod main container.

**Additional context**:
With a podSecurityPolicy setting
```
runAsUser:
    rule: MustRunAsNonRoot

```
the pod cannot be scheduled.

**Environment details**:
- Kubernetes version 1.13.5:
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc): On premises
- bank-vaults version (e.g. 0.4.17): 0.5.3

/kind bug

devlounge:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #663 

### What's in this PR?
see #663

### Why?
see #663

### DIFF
--- a/cmd/vault-secrets-webhook/main.go
+++ b/cmd/vault-secrets-webhook/main.go
@@ -153,9 +153,6 @@ func getInitContainers(originalContainers []corev1.Container, vaultConfig vaultC
 					MountPath: "/vault/",
 				},
 			},
-			SecurityContext: &corev1.SecurityContext{
-				AllowPrivilegeEscalation: &vaultConfig.pspAllowPrivilegeEscalation,
-			},
 			Resources: corev1.ResourceRequirements{
 				Limits: corev1.ResourceList{
 					corev1.ResourceCPU:    resource.MustParse("50m"),


cmd/vault-secrets-webhook/main.go
```diff
@@ -153,9 +153,6 @@ func getInitContainers(originalContainers []corev1.Container, vaultConfig vaultC
 					MountPath: "/vault/",
 				},
 			},
-			SecurityContext: &corev1.SecurityContext{
```
pbalogh-sa: I think, deleting `securityContext` from container level will not solve the issue.

pbalogh-sa: And it will break the privileged feature.

pbalogh-sa:
@devlounge do you have any further plans with this?

devlounge:
I've built our own webhook, which does some extra things based on how we implement vault security internally.
But the solution was to clone, from spe.containers[0], the securityContext.capabilities.drop into the copy-vault-env initContainer and the vault-agent initContainer. and then it works.

pbalogh-sa:
@devlounge Would you plan to contribute back to Bank-Vaults the solution?

pbalogh-sa:
fixed by #698

PR #678, opened 2019-09-26
Vault unseal conf

srleyva:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | Slack

### What's in this PR?

This PR is meant to allow use of vault as a KV store for unseal config. The use case being if another vault wanted to be utilized as a KV Store for the unseal-keys and root token. While there is similar functionality with vault transit unseal, this requires extra configuration on that vault side. This solution purely uses vault as a secure backend KV Store. One strange thing to call out is the introduction of a `VAULT_CREDENTIAL_PATH` to the unseal flow for Vault. This was done to maintain consistency with the other stores, and avoid any potential conflicts with the cluster vault instances themselves.

### Why?

Another feature for a use case where one doesn't have access to KMS with public cloud but maintains an on-prem Vault solution.

### Checklist

### To Do

- [x] - This is still a WIP. Waiting to adjust documentation/cli help until feedback has been given on the proposal.


bonifaido:
Any update on this @srleyva? I think this path is quite good, only the operator unsealConfig part has to be handled as well.

srleyva:
Oh this is correct. Apologies, didn't see your original message. I'll update by EOD. @bonifaido

srleyva:
Actually, I assume because they are implementing the same interface, this should just to be able to be used as another unsealConfig backend because a `KVService` interface is returned and called as such? @bonifaido

srleyva:
This is definitely out of the scope of this PR, but wanted to see if there was an agreement or something I am missing. My personal use case is satisfied with a Kubernetes `hostPath` mount but the `credentialsConfig` supports Kubernetes `secrets`. Would something like AppRole be better as you can restrict by `secret_id_bound_cidrs` and `secret_id_num_uses`? The Operator could generate a `secret_id`, good for one time use, with a special token and then use an `app_role_id` passed in through the `credentialsConfig` to generate a short lived token for unseal. In theory, the unseal token could be very short lived as its only needed for the duration of the unseal and every unseal event could generate a new `secret_id`. This way the token that is passed to the operator (and presumably in kubernetes secrets) can only generate a `secret_id` and this needs 1. to be coupled with the `app_role_id`, 2. to be used from a specific CIDR block. Additionally, this produces a short lived token. Admittedly, this would require some behavior change on the kv vault client creation...just wanted to see if this was a better solution than passing a long lived token into kubernetes secrets.

bonifaido:
If you would use the Vault Client included in Bank-Vaults that would allow you to use Kubernetes SA based authentication (and token renewal)  and also it does the same file reading if the vault-token file is available. Please check: https://github.com/banzaicloud/bank-vaults/blob/master/pkg/sdk/vault/client.go#L118

srleyva:
Ohhh yah! I considered using the SDK vault client but went this route instead. That is definitely a better approach. I'll make the change.

srleyva:
I made the changes to use SDK @bonifaido. The supported flow is what comes out of the box with the SDK. 

Order of priority:
1. Vault Token set directly in Vault config
2. Vault Token file
3. Vault Kubernetes Auth

bonifaido:
@srleyva I have extended the implementation to the operator and added some changes there: https://github.com/banzaicloud/bank-vaults/pull/687

srleyva:
Apologies, I was conflating bank-vault with the kubernetes operator. Those changes are awesome! Thanks for the help on this feature! @bonifaido

PR #679, opened 2019-09-29
Enable possibility for JSON logging

tommy-dk:
| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | yes
| API breaks?     | no|yes
| Deprecations?   | no|yes
| Related tickets | none

### What's in this PR?
Added the possibility of logging in JSON format in both `vault-secrets-webhook` as well as in `vault-env`. Furthermore I've added a global hook for adding additional field to all log entries. I added `app` as a first field, for easier debugging.

Enabling JSON logging can be configured through the environment variable `enable_json_log` and can be overridden individually through the annotation: `vault.security.banzaicloud.io/enable-json-log` to control the output of `vault-env`.

Cleanup and parity of log methods were also done. So now logrus is used everywhere.

### Why?
JSON logging is very useful for structured logging, and is easily parseable for tools such as Datadog, 
Papertrail, etc. Furthermore JSON logging makes it easy to add new fields.

### Checklist

### To Do
I would like if someone could help me thoroughly test this feature for any regression.

### DIFF
--- a/cmd/vault-env/main.go
+++ b/cmd/vault-env/main.go
@@ -54,6 +54,21 @@ var sanitizeEnvmap = map[string]bool{
 	"VAULT_PATH":                   true,
 	"VAULT_IGNORE_MISSING_SECRETS": true,
 	"VAULT_ENV_PASSTHROUGH":        true,
+    "VAULT_JSON_LOG":               true,
+}
+
+var logger *log.Logger
+
+type GlobalHook struct {
+}
+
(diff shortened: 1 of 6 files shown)


bonifaido:
Hi @tommy-dk, thanks for the PR, tbh looks good to me, the acceptance test would have been failed if this caused any regression. Nice job!

tommy-dk:
Awesome, thank you! 👍

PR #696, opened 2019-10-09
Slim vault-env docker image size from 127Mb to 24Mb

lhotrifork:
Hi,

I couldn't find out any use of vault inside `cmd/vault-env/main.go` and experimented with removing vault. Sure enough it worked.

The resulting image is now ~24Mb instead of 127Mb.

Hope you find this useful!

### What's in this PR?
vault-env no longer build from `vault:1.2.3` but from `alpine:3.10` to remove the unneeded dependency on vault.

### Why?
I was wondering why the docker-env image was so big, and found that it included the entire 104Mb vault binary.

### Checklist

### DIFF
--- a/Dockerfile.vault-env
+++ b/Dockerfile.vault-env
@@ -17,9 +17,9 @@ COPY . /build
 RUN go mod download
 
 COPY . /go/src/${PACKAGE}
-RUN CGO_ENABLED=0 go install ./cmd/vault-env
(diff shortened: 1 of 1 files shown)


bonifaido:
Hi @lhotrifork great improvement! I'm going to test this out as well. Could you please sign the CLA in the meantime?

PR #701, opened 2019-10-12
Add first unit tests to vault-secrets-webhook

lhotrifork:
| Q               | A
| --------------- | ---
| Bug fix?       | yes 

### What's in this PR?
I've added two tests, during which I found one bug.
The tests can serve as a starting point for developing a full unit test suite.
Before I spend too much time on writing more tests, I would like some feedback on the style.

I've restructured the code a little bit to make it easier to write tests:
- Use `Kubernetes.Interface` instead of `Kubernetes.ClientSet` in `main.go` to allow mocking
- Moved `vaultConfig` to `internal` package to allow mocking it
- Extracted logic for determining if a env-var has `vault:` or `>>vault:` prefix into a function.

I also found that containers with a `>>vault:` prefix would not get mutated - which seemed strange to me, so I changed it. Let me know if this was intended behaviour.

Most of all I'm just looking for feedback on the PR, as I'm still quite new to writing Go :)

### DIFF
--- a/cmd/vault-secrets-webhook/configmap.go
+++ b/cmd/vault-secrets-webhook/configmap.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"strings"
 
+	internal "github.com/banzaicloud/bank-vaults/internal/configuration"
 	"github.com/banzaicloud/bank-vaults/pkg/sdk/vault"
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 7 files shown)


cmd/vault-secrets-webhook/configmap.go
```diff
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"strings"
 
+	i "github.com/banzaicloud/bank-vaults/internal/configuration"
```
bonifaido: One comment (nit): why is this called `i`? :)

lhotrifork: No particular reason. I've renamed it to internal if that fits better?

bonifaido: Great thanks!!

Issue #705, opened 2019-10-16
[Operator] [Power] : Unable to deploy created operator docker image on Openshift for Power


akhileshbhople:
I have made necessary changes to the dockerfile for operator to work with power(ppc64le) architecture. However, while deploying the operator using the steps mentioned here (https://github.com/banzaicloud/bank-vaults/tree/master/docs/operator#k8s-deployment) the pods show following errors in the logs:

{"level":"error","ts":1571124854.9311707,"logger":"cmd","msg":"","error":"failed to create or get service for metrics: services \"vault-operator-metrics\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: no RBAC policy matched, <nil>","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\nmain.main\n\t/build/operator/cmd/manager/main.go:160\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:200"}

I looked through the repository but could not find any references to the vault-operator-metrics.
Is there anything I am missing in the deployment steps?

Please help fix the issue.

bonifaido:
It sounds like an RBAC issue for me, for which we had a PR already but was closed then. Let me try to get it into master again.

May I ask in what kind of environment do you use it on Openshift Power?
Also, are you willing to contribute the Power Dockerfile to the project?

bonifaido:
A possible fix has been merged @akhileshbhople , can you please try again?

akhileshbhople:
> It sounds like an RBAC issue for me, for which we had a PR already but was closed then. Let me try to get it into master again.
> 
> May I ask in what kind of environment do you use it on Openshift Power?
> Also, are you willing to contribute the Power Dockerfile to the project?

If all the tests go through as expected for ppc64le, will be very interested to contribute Dockerfile for ppc64le.

bonifaido:
I think now you have to create a Vault instance with it, so please follow the above-linked deployment steps and deploy the rbac.yaml and the cr.yaml, if that works properly we are more or less okay!

bonifaido:
Any update on this @akhileshbhople ?

akhileshbhople:
[...]

PR #707, opened 2019-10-16
operator: fix rbac for openshift

bonifaido:

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes (partially) #705 

### What's in this PR?

Setting finalizers on deployments has to be enabled for creating the `vault-monitor-service` by the operator (its owner is the vault-operator deployment itself.

### DIFF
--- a/charts/vault-operator/Chart.yaml
+++ b/charts/vault-operator/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: 0.5.0
+appVersion: 0.5.3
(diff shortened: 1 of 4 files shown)


satoshinakamotoblockchain1:
So because I own so much we are going to make 5 sets and one master set that say where every thing is at every set has is back up on the blockchain as well privately as we coin stock stakes tokens companies owne AI U [ GO ] MAKE IT SO TRIPPLE BIOMETRIC LOCKS

PR #711, opened 2019-10-21
[webhook/vault-env] Support Vault Transit Secret Engine for PODs

leominov:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Support Transit Secret Engine in environment variables.

P.S.: Looks like whole app must be rewritten.

### Why?

Our developers haven't rich access to Vault, with this changes they can able to encrypt data (we give write access to `encrypt` endpoint of transit engine) and Push it into specs without us. Role used in cluster should have write access to `decrypt` endpoint.

### Checklist


bonifaido:
Great feature @leominov, I'm okay with it! Could you please provide a working example app that uses this feature? (Also some operator CR which configures the backend accordingly?)

leominov:
Sure

leominov:
@bonifaido I'll also tweak webhook :)

cmd/vault-env/main.go
```diff
@@ -107,6 +111,7 @@ func main() {
 	}
 
 	ignoreMissingSecrets := os.Getenv("VAULT_IGNORE_MISSING_SECRETS") == "true"
+	ignoreTransitDecryptError := os.Getenv("VAULT_IGNORE_TRANSIT_DECRYPT_ERROR") == "true"
```
bonifaido: I'm not sure, maybe we can just reuse `VAULT_IGNORE_MISSING_SECRETS` here?

leominov: Good compromise I think

leominov:
Wow, great!

PR #719, opened 2019-10-25
Add support for watching external secrets by annotation as well as labels

### LINKED ISSUE #720: watchedSecretsLabels does not support cert-manager created secrets from 0.11
`cert-manager` generated secrets starting with version 0.11 do not label secrets anymore , they instead moved to annotations. see https://github.com/jetstack/cert-manager/issues/2265

This means that `watchedSecretsLabels` can't be used anymore to monitor cert-manager secrets and roll restart vault pods when they are renewed

MR https://github.com/banzaicloud/bank-vaults/pull/719

Unfortunately Secrets can not be pulled with an `Annotations Selector` from kubernetes directly like labels can, so i ended up rewriting the whole bit with backward compatibility.

Could not think of any other than extending the operator to be able to watch secrets based on annotations as well as labels

primeroz:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #720

### What's in this PR?
This PR extend the already existing support for watching external secrets by labels to be able to also watch them by annotation.

Unfortunately Secrets can not be pulled with an Annotations Selector from kubernetes directly like labels can, so i ended up rewriting the whole bit with backward compatibility.

Now the list of Secrets for the `v.Namespace` is pulled down and the selection / comparison of secrets metadata to selectors is performed by the operator itself rather than asking kubernetes for the list of secrets matching the selector.

I could have left the labels one to be done the old way and just do the annotations in the new way but i thought consistency would be good. 

It surely is less performant than asking kubernetes apiserver for a list of secrets matching a selector but in my local tests the performances of doing it this way is still good.

### Why?
This is required for secrets created by cert-manager 0.11+ which is probably one the most likely use cases for this feature

### Additional context
I rerun the `operator-sdk generate k8s` with operator-sdk v0.7.0 and got more changes then expected, not sure what is going on there. Do i need a new version of the sdk ? 

If this is merged *Ideally* i would like to do an hotfix to 0.5 as well since i kind of need this in production where i currently had to lock my vault clusters to cert-manager 0.10.1 to not loose this functionality

### Checklist

### To Do


operator/pkg/controller/vault/vault_controller.go
```diff
@@ -191,6 +191,71 @@ func createOrUpdateObjectWithClient(c client.Client, o runtime.Object) error {
 	return err
 }
 
+// Check if passed object labels/annotation match the passed selector
+// all K/V in the selector must be equal in the object to return true
+func secretMatchLabelsOrAnnotationsCheck(object map[string]string, selector map[string]string) bool {
```
bonifaido: Maybe we could just replace this with `labels.SelectorFromSet(selector).Matches(object)` where `labels` is from `"k8s.io/apimachinery/pkg/labels"` and it is already imported and used here.

primeroz: Sure. Is there an equivalent one for annotations that you know of ?

I changed the previous label selection so that it would work the same for both labels and annotations

Sorry for brevity, writing from phone

bonifaido: No problem!

This function accepts type aliases for `map[string]string` so I think it should work for both labels and annotations as well.

primeroz: This is done and now using `labels.SelectorFromSet(selector).Matches(object)` for matching selectors for both annotations and labels

primeroz:
I also added checking this works as expected to `acceptance-test.sh` 

I added it as *second test* though so that all the rest will use the same vault deployment as before which now is the *third test*

primeroz:
@bonifaido all changes done and after 1001 runs now CI is passing as well

operator/pkg/controller/vault/vault_controller.go
bonifaido: Really great!

bonifaido:
Thankfully this is an open source project so all 1001 runs was on the House!

PR #723, opened 2019-10-29
[charts/vault-secrets-webhook] Add option to use static certicates for vault-secrets-webhook helmchart.

lhotrifork:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Allow inserting static 64encoded CA and TLS certificates for the mutating webhook - off by default.
```
certificate:
    generate: true
    server:
        tls:
            crt:
            key:
    ca:
        crt:
```
I've also changed the helm chart so the top-level `List` object is no longer needed.

### Why?
The current implementation is causing us some troubles because we use git-ops via helmsman. 
Every time helmsman checks if there’s any changes to deploy (with helm-diff), helm will generate a new certificate and that will cause a redeploy.

### DIFF
--- a/charts/vault-secrets-webhook/Chart.yaml
+++ b/charts/vault-secrets-webhook/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "0.5.1"
 description: A Helm chart that deploys a mutating admission webhook that configures applications to request env vars from Vault Secrets
 name: vault-secrets-webhook
-version: 0.5.3
+version: 0.5.4
 maintainers:
 - name: Banzai Cloud
   email: info@banzaicloud.com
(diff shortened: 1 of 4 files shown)


lhotrifork:
@bonifaido 
I was unsure if it was better keep the values in plaintext and let the helm chart base64 encode them. 
In the end I chose keep the values base64 encoded, because then I could simply copy/paste the values from the last helm deploy into my values-file.

bonifaido:
At is absolutely okay, now I'm trying to resurrect the build.

bonifaido:
Sorry, I can't get the build back, still trying (reopened it, but CircleCI is having really hard time in the past few months).

bonifaido:
I submitted https://github.com/banzaicloud/bank-vaults/pull/725 to circumvent the test run, once it completes I will close that one :)

PR #726, opened 2019-10-31
WIP: [webhook] Allow disable cache or cache only images with digest

leominov:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Now cache type are configurable, cache options contains option to disable caching tagged Docker-images, some tests for it.

### Why?

- In some cases no need to cache information about Docker-image (currently storages works without TTL)
- In some cases images with same tag contains different entrypoints

### Checklist

### DIFF
--- a/cmd/vault-secrets-webhook/main.go
+++ b/cmd/vault-secrets-webhook/main.go
@@ -915,6 +915,8 @@ func init() {
 	viper.SetDefault("registry_skip_verify", "false")
 	viper.SetDefault("debug", "false")
 	viper.SetDefault("enable_json_log", "false")
+	viper.SetDefault("registry_cache_storage", "inmemory")
+	viper.SetDefault("registry_cache_opts_digest_only", "false")
 	viper.AutomaticEnv()
 
 	logger = log.New()
--- a/cmd/vault-secrets-webhook/main.go
+++ b/cmd/vault-secrets-webhook/main.go
@@ -916,7 +916,7 @@ func init() {
 	viper.SetDefault("debug", "false")
 	viper.SetDefault("enable_json_log", "false")
 	viper.SetDefault("registry_cache_storage", "inmemory")
-	viper.SetDefault("registry_cache_opts_digest_only", "false")
+	viper.SetDefault("registry_cache_digest_only", "false")
 	viper.AutomaticEnv()
 
 	logger = log.New()
--- a/cmd/vault-secrets-webhook/registry/image_cache.go
+++ b/cmd/vault-secrets-webhook/registry/image_cache.go
@@ -15,7 +15,7 @@
 package registry
 
 import (
-	"sync"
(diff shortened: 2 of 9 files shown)


cmd/vault-secrets-webhook/main.go
```diff
@@ -915,6 +915,8 @@ func init() {
 	viper.SetDefault("registry_skip_verify", "false")
 	viper.SetDefault("debug", "false")
 	viper.SetDefault("enable_json_log", "false")
+	viper.SetDefault("registry_cache_storage", "inmemory")
+	viper.SetDefault("registry_cache_opts_digest_only", "false")
```
bonifaido: I think we can leave out the `opts_` part from there.

leominov: Sure, we can :)

cmd/vault-secrets-webhook/registry/image_cache.go
```diff
@@ -0,0 +1,50 @@
...
+type ImageCacheOptions struct {
+	// DigestOnly used for more sensitive caching, when you are using same
+	// tags for Docker-images that can change as can change the entrypoint.
+	//
+	// Setting as `true` may increase network traffic on Docker Registry.
+	DigestOnly bool
+}
+
+// NewImageCache returns cache by storage name
+func NewImageCache(storage string, opts *ImageCacheOptions) ImageCache {
```
bonifaido: I'm also not sure if we pass in an `ImageCacheOptions` can't we just put the storage type to there as well? ;-)

leominov: hm, I didn't think about it :) Ok, I'll fix it little bit later.

leominov: I think storage type in only thing that can be passed to NewImageCache for now. Options from `ImageCacheOptions` should be named as something else and checked before calling cache methods, cause another storage type (such as redis or file) must contains another check of `DigestOnly` (for example) and bunch of tests for it. Also I think in future `ImageCacheOptions` sould contains more specific storage options like storage URI, seperator or prefix (for kv storages), db index (for redis, why not?), credentials, TTL (for some of kv storages) or data directory (for file storage).

What you think about it, @bonifaido?

bonifaido: Yes, since this just a constructor the only thing it matters now that it returns an `ImageCache` interface, and that should remain. I'm okay with it :)

leominov:
:ohno:

bonifaido:
No problem, we can pin this PR. Do you plan to add some more commits to it (I gues)?

leominov:
sure :)

PR #730, opened 2019-11-06
add the ability to apply additional labels to the vault statefulset 

### LINKED ISSUE #729: Allow for custom labels on the vault statefulset and pods when using the vault helm chart
As it stands the vault helm chart does not allow for the application of custom labels to the vault statefulset or pods. The use of custom labels is many fold from tracking, monitoring or even something such as Azure AAD Pod Identity for authentication. 

I would like the ability to define and apply custom labels to the vault statefulset and the resulting pods.

Manually patching deployed resources would be a work around but this is not a viable solution

KealanM:
add the ability to apply additional labels to the vault statefulset  and pods addresses #729

Signed-off-by: Murphy, Kealan <kealan.murphy@aspect.com>

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #729 

### What's in this PR?
This PR gives users the ability to define and apply additional labels to the Vault Statefulset and Pods, this has many use cases. The one use case in particular that would be of use is when using running Vault in Azure and you want to use the aad-pod-identity plugin to allow the vault pod to assume a Managed Identity.

### Checklist

### DIFF
--- a/charts/vault/values.yaml
+++ b/charts/vault/values.yaml
@@ -54,6 +54,10 @@ podAnnotations:
   prometheus.io/scrape: "true"
   prometheus.io/path: "/metrics"
   prometheus.io/port: "9102"
+
+additionalLabels: 
+  #  team: banzai
+
 resources: {}
   # We usually recommend not to specify default resources and to leave this as a conscious
   # choice for the user. This also increases chances charts run on environments with little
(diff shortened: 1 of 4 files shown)


charts/vault/values.yaml
```diff
@@ -54,6 +54,10 @@ podAnnotations:
   prometheus.io/scrape: "true"
   prometheus.io/path: "/metrics"
   prometheus.io/port: "9102"
+
+additionalLabels:
```
bonifaido: Please call it just `labels` otherwise LGTM!

PR #731, opened 2019-11-07
[webhook] Disable image caching for Policy=PullAlways and ref=latest

leominov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Disabling cache for Docker images that tagged as `latest` and for images from containers with ImagePullPolicy `PullAlways`.

### Why?

With "latest" tag user says, that image can be changed and must be pulled every time; with "Pull Always" option user says, that image has static tag, but might be changed in any time.

### To Do

First step to done with #726.


leominov:
I don't think that this would be a part of #726, but default caching policy should works as this. I think :)

PR #780, opened 2019-12-05
adding new parameters to control the sideEffect

nullck:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
I'm using an old version of Kubernetes (1.10.11)  here in the company I'm working (shame on me) , and we're trying to update the vault-secrets-webhook and we got this issue:

```
Comparing vault-secrets-webhook banzaicloud-stable/vault-secrets-webhook
error validating "": error validating data: ValidationError(MutatingWebhookConfiguration.webhooks[1]): unknown field "sideEffects" in io.k8s.api.admissionregistration.v1beta1.Webhook
```
The reason behind this error is explained here:  https://github.com/linkerd/linkerd2/issues/2850

### Why?
This PR allows the user to choose between to active or not the sideEffects in the MutatingWebhookConfiguration.

### Checklist

### To Do

### DIFF
--- a/charts/vault-secrets-webhook/templates/apiservice-webhook.yaml
+++ b/charts/vault-secrets-webhook/templates/apiservice-webhook.yaml
@@ -139,6 +139,6 @@ webhooks:
       values:
       - {{ .Release.Namespace }}
 {{- end }}
-{{- if and (eq (int $major) 1) (ge (int $minor) 12) }}
+{{- if and (ge (int $major) 1) (ge (int $minor) 12) }}
   sideEffects: {{ .Values.apiSideEffectValue }}
 {{- end }}
--- a/charts/vault-secrets-webhook/templates/apiservice-webhook.yaml
+++ b/charts/vault-secrets-webhook/templates/apiservice-webhook.yaml
@@ -139,6 +139,6 @@ webhooks:
       values:
       - {{ .Release.Namespace }}
 {{- end }}
-{{- if .Values.apiSideEffect }}
+{{- if and (eq (int $major) 1) (ge (int $minor) 12) }}
   sideEffects: {{ .Values.apiSideEffectValue }}
 {{- end }}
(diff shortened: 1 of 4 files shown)


charts/vault-secrets-webhook/templates/apiservice-webhook.yaml
```diff
@@ -139,4 +139,6 @@ webhooks:
       values:
       - {{ .Release.Namespace }}
 {{- end }}
-  sideEffects: NoneOnDryRun
+{{- if .Values.apiSideEffect }}
```
Kasama: What do you think of using the `$major` and `$minor` variables to detect if this flag is supported automatically, instead of having this be user-configurable.
Something like
```yaml
{{- if and (eq (int $major) 1) (ge (int $minor) 12) }}
  sideEffects: NoneOnDryRun
{{- end }}
```

bonifaido: I agree with @Kasama on this, this would be more convenient.

nullck: @Kasama  ok looks good, I'm going to check how to implement

charts/vault-secrets-webhook/templates/apiservice-webhook.yaml
```diff
@@ -139,4 +139,6 @@ webhooks:
       values:
       - {{ .Release.Namespace }}
 {{- end }}
-  sideEffects: NoneOnDryRun
+{{- if and (eq (int $major) 1) (ge (int $minor) 12) }}
```
Kasama: it's probably better to use `(ge (int $major))` just in case kubernetes rolls out a version 2. What do you think?

nullck:
my pleasure @bonifaido  @pbalogh-sa @Kasama , thanks for accept this change :D

Issue #420, opened 2019-04-11
vault-env should renew leases of dynamic secrets


msvticket:
**Is your feature request related to a problem? Please describe.**
When using the Mutating Webhook feature to access dynamic secrets (like database passwords) the default_ttl now needs to be set to the anticipated execution time of the pod instead of being able to set default_ttl to a lower value and only set max_ttl to the anticipated execution time.

**Describe the solution you'd like**
vault-env renews the lease at the appropriate times.

**Describe alternatives you've considered**
Maybe this could be solved by integrating with Vault Agent.

bonifaido:
Hi @msvticket! The vault-env binary replaces itself with the actual program it is requesting secrets for, so it is best used for very long living secrets like certificates/ssh keys or KV secrets.

For dynamic secrets there is a PR in the making, actually mutating the pods to inject a consul-template sidecar into them, please follow up on https://github.com/banzaicloud/bank-vaults/pull/410 for more details.

msvticket:
Yeah, I know the current design of vault-env would need to be change to fulfil my suggestion. Thanks for the pointer to the pull request. I'm not surprised that the problem is being worked on.

PR #788, opened 2019-12-10
Terminate on lease time

boekkooi-fresh:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | alternative to https://github.com/banzaicloud/bank-vaults/issues/662, related to https://github.com/banzaicloud/bank-vaults/issues/420

### What's in this PR?
This PR adds a feature that will terminate the vault-env entrypoint process once the lowest secret lease time is reached, allowing k8s to reschedule the pod and get fresh secrets.
This feature can be enabled and configured by using the follow annotations:
- vault.security.banzaicloud.io/ttl-terminate: Enable the feature by setting this to `true`
- vault.security.banzaicloud.io/ttl-terminate-before: Time to start termination before the lease expires
- vault.security.banzaicloud.io/ttl-terminate-signal: The termination signal (default = SIGTERM)
- vault.security.banzaicloud.io/ttl-terminate-grace-period: Time to wait after the Signal was send before a SIGQUIT is send

When the feature is enabled it changes the behavior of `vault-env` because the entry points process is now started using `os.StartProcess` (which uses `syscall.StartProcess`). 
This means that vault-env process will no longer be replaced but can keep running in order to terminate the entry points process when the secret lease expires.

### Why?
Currently getting the TTL for a secret to work properly requires the use of consul-template which in turn means the application or the bootstrap of the application needs to read the file. 
In my case this means changing a bunch of code that I would rather not touch. In order to avoid this in our case terminating the processes and existing will cause k8s to spin up a new pod.
This new pod will then get fresh secrets and will allow our systems to keep running.

### Additional context
This is still a WIP but I would really appreciate some comments around this idea and other suggestions on how to improve the code and add proper tests.

### Checklist

### To Do

- [ ] Implement Tests (I really could use some help with this)
- [ ] Update documentation


bonifaido:
Hi @boekkooi-fresh, thank you for your contribution! Is this still a WIP am I right?

startnow65:
@boekkooi-fresh Thank you for this, this would take us closer to achieve consuming environment variable based dynamic secrets from vault.

I would like to suggest adding a random time interval (+/-) to the value of `ttl-terminate-before`.

Something like `ttl-terminate-vary` which would get a random int `x` such that `(-1 * ttl-terminate-vary) < x < ttl-terminate-vary` the value of `x` would be added to `ttl-terminate-before` to get the actual time to terminate for a pod.
Or have `ttl-terminate-before-min` and `ttl-terminate-before-max` to provide the range to pick the actual `ttl-terminate-before` for a pod.

This would spread out the termination time and prevent bringing down a bunch of pods from the same deployment that came up at about the same time

boekkooi-fresh:
> Hi @boekkooi-fresh, thank you for your contribution! Is this still a WIP am I right?

Hi @bonifaido 

Yes this is still WIP since It's missing tests and docs but please give it a review and provide your thoughts. 
I'm now running some tests with this code on one of our staging environment everything looks good except for the issue that all pods restart at the same time if they where created at the same time so implementing some sort of random factor there maybe a good idea as @startnow65 suggested,

Please do give this a review and tell if you think the solution is viable :smile: 
Thanks in advance!

boekkooi-fresh:
The more I'm looking at this solution and testing it I'm not sure if this is the right way forward. Since a Lease can be renewed which would allow the pod to keep running. Combining this with a lack of support for K/V secret changes and it feels a little to much like a hack. 
Having a operator or sidecar to handle lease renewals/revokes etc. would be a lot nicer in my opinion.

boekkooi-fresh:
After doing more tests with this PR we have to conclude that in order to fix the issues in regards to secret leasing we need an extra controller. Since there is no way to ensure that the pod can restart if a containers fails it can happen that all pods go into a bad state making it unreachable causing failed.

PR #795, opened 2019-12-11
Startupsecret from k8ssecret

### LINKED ISSUE #774: Feature request: startupSecret from configmap/secrets
When are You bootstraping cluster ( ready to use ), there is usually need to have secrets available before bank-vaults is even deployed. 
- no ingress -> You can not connect to vault
- no certificate -> if You are using something like cert-manager, You have to solve insecure skip verify  
- no dns record -> if You are using something like external-dns, You can not even connect 

Would be great if startupSecrets can be set from k8s secrets/configmaps ( https://github.com/banzaicloud/bank-vaults/blob/master/operator/deploy/cr.yaml#L138 )

Something like:
```
    startupSecrets:
      - type: kv
        path: secret/data/accounts/aws
        data:
           configMapKeyRef
             - name: AWS_ACCESS_KEY_ID
               key: awsid
           secretKeyRef:
             - name: AWS_SECRET_ACCESS_KEY
               key: awskey
```

Is possible to run batch/job to import initial secrets from k8s to vault with some scripting.
- curl
- vault write 

**Additional context**
none

pbalogh-sa:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #774 

### What's in this PR?
With this PR the vault-configurer can use Kubernetes secrets as startup secrets for Vault

### Why?
In some cases, it will be useful to store initial secrets in Kubernetes, and synchronizing to the Vault in configuration time.


pkg/sdk/vault/operator_client.go
```diff
@@ -1555,3 +1566,40 @@ func isConfigNoNeedName(secretEngineType string, configOption string) bool {
...
+		return map[string]interface{}{}, err
+	}
+
+	k8sCfg := crconfig.GetConfigOrDie()
+	c, err := crclient.New(k8sCfg, crclient.Options{})
+	if err != nil {
+		return map[string]interface{}{}, err
+	}
+
+	vaultNamespace := strings.Split(strings.Split(os.Getenv("VAULT_ADDR"), ".")[1], ":")[0]
```
primeroz: quick question, is there no better way to grab the namespace at this point ? 

since we are already exposing the VUALT_ADDR environment variable [here](https://github.com/banzaicloud/bank-vaults/blob/d677f137e9211d941d7c0de3ae0d0dedae7edab4/operator/pkg/controller/vault/vault_controller.go#L982) would it make sense to also expose an ENV for the currently running Namespace ? 

Or probably does not matter ... just wanted to ask, i am not a big fan of string manipulation but in the end is just the same

pbalogh-sa: I agree with you. I will expose namespace to bank-vaults container trough env.

PR #796, opened 2019-12-12
Add Extra init containers

### LINKED ISSUE #777: `fsGroup` Modifies Permission Bits Preventing Vault Backend PostgreSQL with Certificate-based Authentication
**Describe the bug**:
Vault refuses to read the private key file due to permissions not being 0600 or less.

**Expected behaviour**:
Expect Vault to be able to read its private key file set with permissions as described in the `cr.yaml`.

1. Define a secret containing a certificate and private key for PostgreSQL.
2. Mount this secret with permission bits explicitly set.

Something like:

```
...
    volumeMounts:
    - mountPath: /certs/client/
      name: client-cert
      readOnly: true
...

  volumes:
  - name: client-cert
    projected:
      defaultMode: 420
      sources:
      - secret:
          items:
          - key: tls.crt
            mode: 256
            path: tls.crt
          - key: tls.key
            mode: 256
            path: tls.key
          name: vault-client
...
```

3. When running, the Vault container complains that:

> Error initializing storage of type postgresql: failed to check for native upsert: pq: Private key file has group or world access. Permissions should be u=rw (0600) or less

**Additional context**:
I modified operator-generated YAML so that I could get a shell inside the Vault container.

Here's what I found:

- Permissions of `/certs/client/..data/tls.key` are set to `0440` and the ownership is `root:vault`
- I tracked this down to **precisely one setting in the initContainer YAML**:

```
   securityContext:
     fsGroup: 1000
```

Commenting that out lets the user specified configuration of mode bits (and potentially ownership) take precedence.

**Environment details**:
- Kubernetes version (e.g. v1.10.2): v1.16.2
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc): Rancher
- bank-vaults version (e.g. 0.4.17): `helm install --namespace automation vault-operator banzaicloud-stable/vault-operator --version 0.5.6 --set image.tag=0.6.3`
- Install method (e.g. helm or static manifests): `helm install --namespace automation vault-operator banzaicloud-stable/vault-operator --version 0.5.6 --set image.tag=0.6.3`
- Logs from the misbehaving component (and any other relevant logs): Reproduced above
- Resource definition (possibly in YAML format) that caused the issue, without sensitive data: As defined above (put important parts above, it is difficult to redact enough without rewriting a new CR).

/kind bug

pbalogh-sa:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #777 

### What's in this PR?
This PR addresses the capability of defining extra init containers in the vault pod.

### Why?
If some initializations are required before the vault container is started, you'll be able to do it with this feature.

### Checklist

### To Do


operator/pkg/apis/vault/v1alpha1/vault_types.go
```diff
@@ -316,6 +316,9 @@ type VaultSpec struct {
 	// IstioEnabled describes if the cluster has a Istio running and enabled.
 	// default: false
 	IstioEnabled bool `json:"istioEnabled,omitempty"`
+
+	// InitContainers add extra initContainers
+	InitContainers []v1.Container `json:"initContainers,omitempty"`
```
primeroz: should we call it `vaultInitContainers` since they really are going to be attached to the vault stateful set ?

operator/deploy/cr-init-containers.yaml
```diff
@@ -0,0 +1,205 @@
...
+    # Specify Ingress object annotations here, if TLS is enabled (which is by default)
+    # the operator will add NGINX, Traefik and HAProxy Ingress compatible annotations
+    # to support TLS backends
+    annotations:
+    # Override the default Ingress specification here
+    # This follows the same format as the standard Kubernetes Ingress
+    # See: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#ingressspec-v1beta1-extensions
+    spec: {}
+
+  initContainers:
```
primeroz: this could definetely do with an acceptance test :)

PR #797, opened 2019-12-12
vault-env: add version to secret cache key

### LINKED ISSUE #794: vault-env: the secret cache only caches 1 version of the secret
**Describe the bug**:
If one "wanted" to use different versions of the same secret in different env variables, the values would all end up beeing the ones of the first cached version, as the cache key access is based on path and not path+version.

**Expected behaviour**:
make the cache key out of path+version

set 2 env variables:
* one with vault:/foo#bar#1
* one with vault:/foo#bar#2

Othe the env variables with end up with the value of the first fetched version.

bonifaido:

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #794 

### What's in this PR?

The version was missing from the secret cache key.

### DIFF
--- a/cmd/vault-env/main.go
+++ b/cmd/vault-env/main.go
@@ -216,16 +216,18 @@ func main() {
 			version = split[2]
 		}
 
+		secretCacheKey := valuePath + "#" + version
+
 		var secret *vaultapi.Secret
 		var err error
 
-		if secret = secretCache[valuePath]; secret == nil {
+		if secret = secretCache[secretCacheKey]; secret == nil {
 			if update {
 				secret, err = client.RawClient().Logical().Write(valuePath, map[string]interface{}{})
 				if err != nil {
 					logger.Fatalln("failed to write secret to path:", valuePath, err.Error())
 				}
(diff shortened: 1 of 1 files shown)


cmd/vault-env/main.go
```diff
@@ -216,16 +216,18 @@ func main() {
...
+		if secret = secretCache[secretCacheKey]; secret == nil {
 			if update {
 				secret, err = client.RawClient().Logical().Write(valuePath, map[string]interface{}{})
 				if err != nil {
 					logger.Fatalln("failed to write secret to path:", valuePath, err.Error())
 				}
-				secretCache[valuePath] = secret
+				secretCache[secretCacheKey] = secret
 			} else {
 				secret, err = client.RawClient().Logical().ReadWithData(valuePath, map[string][]string{"version": {version}})
```
primeroz: Does this mean we can only read from kv2 ?

bonifaido: No, this is just the cache key in the map.

PR #799, opened 2019-12-13
fix nodeSelector conditionCheck in vault-operator chart deployment file

raoofm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
fix nodeSelector conditionCheck in the deployment

### Why?
in deployment the condition being checked to populate nodeSelector is for tolerations, may be a typo.


raoofm:
update @bonifaido. I didn't update earlier as I thought that will be updated as part of release. Probably charts are not released but just updated in master. My bad sorry :)

PR #814, opened 2019-12-30
Adding new kv type in which encryption/decryption will be handled using rest api call to third party encryption/decryption service

adityagu0910:
| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | yes
| API breaks?     | no|yes
| Deprecations?   | no|yes

### What's in this PR?
This change is to add capability to have ecnryption/decryption using rest api call to third party HSM device like IBM Data Power.

### Additional context
Files were updated and tested locally in my environment by using below variables in helm values.yaml for unsealer:args

args: [
"--mode",
"k8s-with-restapi",
"--k8s-with-restapi-secret-namespace",
"vault",
"--k8s-with-restapi-secret-name",
"bank-vaults-with-restapi",
"--k8s-with-restapi-keyidname",
"testapp",
"--k8s-with-restapi-encryption-url",
"https://test-vault-seal-key:8447/test/api/vault/v1/crypto/encrypt",
"--k8s-with-restapi-decryption-url",
"https://test-vault-seal-key:8447/test/api/vault/v1/crypto/decrypt"
]

### Checklist
- [x] Error handling code meets the [guideline]

### To Do
- Error handling were handled to some extent. Please review and suggest if it is missed at some places.
- Please suggest if any user guide that need to be updated for this small change.


pkg/kv/k8srestapi/k8srestapi.go
```diff
@@ -0,0 +1,195 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package k8srestapi 
+
+import (
+        "encoding/json"
+	"encoding/base64"
```
bonifaido: Can you please format the whole PR code with `go fmt`?

adityagu0910:
@bonifaido  We created encryption/decryption methods in Data power but it can be generic as we can have encryption/decryption REST APIs anywhere third party hosted and we just need to call those REST API URLs. (for Ex.. apiconnect)

pkg/kv/k8srestapi/k8srestapi.go
adityagu0910: Done

bonifaido:
I see @adityagu0910. My thinking here is that the current implementation works with Datapower, and holds some Datapower specific parts we should name it as a Datapower. Do you use a proxy application or you call the Datapower endpoints directly? (As you can see I'm not that familiar with the Datapower API :)  But if this code is totally Datapower unaware I'm happy to leave it as it is and call it REST, but I wonder how can we format the request properly for other APIs we are not aware of currently.

pkg/kv/k8srestapi/k8srestapi.go
bonifaido: Thanks!

pkg/kv/k8srestapi/k8srestapi.go
```diff
@@ -0,0 +1,193 @@
...
+	request.Header.Set("Content-Type", "application/json")
+	client := &http.Client{}
+	response, err := client.Do(request)
+	if err != nil {
+		fmt.Printf("The http request failed %s\n", err)
+	} else {
+		plainTextTemp, _ := ioutil.ReadAll(response.Body)
+		err := json.Unmarshal(plainTextTemp, &pp)
+		if err != nil {
+			panic(err)
```
bonifaido: Could you please return this `error` instead of panic-ing and pass it upstream?

pkg/kv/k8srestapi/k8srestapi.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+	}
+	decoded, err := base64.StdEncoding.DecodeString(pp.PlaintextB64)
+	if err != nil {
+		panic(err)
```
bonifaido: Return the error here rather.

pkg/kv/k8srestapi/k8srestapi.go
```diff
@@ -0,0 +1,193 @@
...
+	request.Header.Set("Content-Type", "application/json")
+	client := &http.Client{}
+	response, err := client.Do(request)
+	if err != nil {
+		fmt.Printf("The http request failed %s\n", err)
+	} else {
+		cipherTextTemp, _ := ioutil.ReadAll(response.Body)
+		err := json.Unmarshal(cipherTextTemp, &ep)
+		if err != nil {
+			panic(err)
```
bonifaido: Return the error here rather.

pkg/kv/k8srestapi/k8srestapi.go
```diff
@@ -0,0 +1,193 @@
...
+			Data: map[string][]byte{key: cipherText},
+		}
+		if k.ownerReference != nil {
+			secret.ObjectMeta.SetOwnerReferences([]metav1.OwnerReference{*k.ownerReference})
+		}
+		secret, err = k.cl.CoreV1().Secrets(k.namespace).Create(secret)
+
+	} else if err == nil {
+		secret.Data[key] = cipherText
+		secret, err = k.cl.CoreV1().Secrets(k.namespace).Update(secret)
```
bonifaido: Please handle this `err`.

adityagu0910:
HI @bonifaido 
There is no datapower specific part in the code as we are just calling REST API URLs. I am calling REST  URL directly as they are accessible directly in my case from container host network. 
I do not think it is possible to format request according to diff-diff API but we can document how we can create REST APIs. For Example below is the swagger file - 

[swagger.txt](https://github.com/banzaicloud/bank-vaults/files/4032364/swagger.txt)

pkg/kv/k8srestapi/k8srestapi.go
adityagu0910: This function is copied as it is from k8s.go :) 
below will work for this ?
```suggestion
		       } else if err == nil {
                secret.Data[key] = cipherText
                secret, err = k.cl.CoreV1().Secrets(k.namespace).Update(secret)
                //reflect.DeepEqual()
                if err != nil {
                        return fmt.Errorf("error updating secret key '%s' into secret '%s': '%s'", key, k.secret, err.Error())
                }
```

pkg/kv/k8srestapi/k8srestapi.go
adityagu0910: printing the error is fine or should I return this as part of decrypt/encrypt function ?

adityagu0910: Added like this 
```suggestion
			else {
                cipherTextTemp, _ := ioutil.ReadAll(response.Body)
                err := json.Unmarshal(cipherTextTemp, &ep)
                if err != nil {
                        return nil, err
                }
        }
        return ([]byte(ep.EncryptedB64)), nil
}
```

bonifaido: Please don't print, rather retrun ;)

pkg/kv/k8srestapi/k8srestapi.go
bonifaido: Ah sorry, missed the context, looks great!

bonifaido:
Are you using https://hub.docker.com/r/ibmcom/datapower/ ? Where can we find some docs about the API?

adityagu0910:
HI @bonifaido 

swagger file itself a documentation for API. 
https://github.com/adityagu0910/API-Swagger-File/blob/master/swagger.yaml

I will add link for how to code for API in Datapower/Apiconnect. Should I add all these details in below doc or should I create another ?
https://github.com/banzaicloud/bank-vaults/blob/master/docs/unseal-keys/README.md

pkg/kv/k8srestapi/k8srestapi.go
adityagu0910: Thanks

pkg/kv/k8srestapi/k8srestapi.go
adityagu0910: return statement was added in decrypt and encrypt functions for errors
Thanks!

adityagu0910:
Here are some URL for creating restful APIs 
https://medium.com/@rohitgoyal/restful-api-with-ibm-datapower-part-1-ab3645a93d0a

https://www.ibm.com/support/knowledgecenter/SS9H2Y_7.7.0/com.ibm.dp.doc/encrypt-string_cryptographicfunction.html

adityagu0910:
HI @bonifaido 

Could you please check once you get time if error handling is okay now and if it can  be merged.

bonifaido:
Hi @adityagu0910, could you please check the Banzai Community Slack? I have asked some questions there, but I see that you haven't logged in for a while, and maybe we could discuss there much quicker ;-)

Issue #822, opened 2020-01-08
Injecting consul-template via webhook stalls on initcontainer when Istio mTLS is employed


funkypenguin:
**Describe the bug**:

When using Istio mTLS, and following https://github.com/banzaicloud/bank-vaults/blob/master/docs/mutating-webhook/consul-template.md, by using the `vault.security.banzaicloud.io/vault-ct-configmap` annotation, a pod can't run beyond the InitContainer stage.

Why? Because Istio's CNI causes all outbound traffic to be redirected to the port of the istio-proxy sidecar, even if that sidecar is not running _yet_. Effectively, unless worked around with excludeOutboundIPs or excludeOutboundPorts, an initcontainer will have no egress network connectivity when mTLS is enforced.

Note that this is similar to the issue raised in #773 

**Expected behaviour**:

InitContainers should be avoided due to the complications with Istio mTLS.

**Steps to reproduce the bug**:

1. Create Istio-enabled namespace, deploy vault with Istio mTLS enabled
2. Deploy pod with `vault.security.banzaicloud.io/vault-ct-configmap` set to valid configmap, per 
https://github.com/banzaicloud/bank-vaults/blob/master/docs/mutating-webhook/consul-template.md
3. Pod will fail at the vault-agent init container, logging "connection refused" when attempting to connect to Vault
4. Optionally, add `traffic.sidecar.istio.io/excludeOutboundPorts: "8200"` annotation to pod to permit egress connection to Vault, to validate (this is my current workaround, thanks @waynz0r )

**Additional context**:

A solution was discussed in https://community-banzaicloud.slack.com/archives/CFJJW9L94/p1578513527045200

**Environment details**:
- Kubernetes version (e.g. v1.10.2): 1.16.2
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc): Bare metal
- bank-vaults version (e.g. 0.4.17): latest
- Install method (e.g. helm or static manifests): helm

/kind bug

bonifaido:
Fixed in #823 with vault-agent support.

PR #823, opened 2020-01-09
Implement vault-agent templating support

### LINKED ISSUE #302: vault-secrets-webhook: vault agent based authentication
In the vault-secrets-webhook source code, I can multiple lines triggering more or less actions if "useAgent" is set in the config.

However, these specific lines bellow confuse me a lot because "banzaicloud/vault-secrets-init" image does not exist on dockerhub.

```
if vaultConfig.useAgent {
		containers = append(containers, corev1.Container{
			Name:            "vault-agent",
			Image:           "banzaicloud/vault-secrets-init:latest",


```

Moreover, the tag is hardcoded to latest, which is a bad practice in general.

### LINKED ISSUE #464: vault-secrets-webhook: Support re-authentication with Consul-Template
Right now the vault-token is set via vault-agent during init phase. Consul-Template will use the token and renew it if the token is near the expiration time.
AFAIU, two different use-cases are not handled right now:

1. kubernetes auth role is configured with a max_ttl of N hours. When max_ttl is reached, the token is not renewable anymore but CT is not able to retrieve a new token and will eventually fail. After restart, CT will use the same expired token, thus not being able to self recover unless a new deployment is triggered.
2. The token is manually revoked via Vault. The result is similar to the previous use case.

Somehow the system should be able to self-recover. We may use vault-agent as a sidecar in order to have it handle the token re-creation (this way, we may be also able to use consul-template 0.20.0 since the renewal of the token would be handled by vault-agent).

Create a custom CT image that includes vault-agent as a daemon. This may be lighter, but is against the one-process-per-container rule.

### LINKED ISSUE #804: [webhook] OwnerReferences for vault-agent ConfigMap missing if pod has a name
**Describe the bug**:
I'm creating a standalone pod with a CtConfigMap annotation. After the pod deletion, the "vault-agent-config" configmap does not get deleted.

**Expected behaviour**:
The ConfigMap should be garbage collected.

1. Create a pod with a Consul-Template configMap configuration
2. Delete the pod

**Additional context**:
It seems that in the vault-agent-config configmap created by the webhook for the pod the `ownerReferences` object is missing, therefore there's no link for the K8s garbage collector.

According to https://github.com/banzaicloud/bank-vaults/blob/master/cmd/vault-secrets-webhook/main.go#L479 the OwnerReferences object is retrieved from the pod only if the name is not set (I assume this is for when the pod is part of a [replica|daemon]Set). In my case, the pod name is set, hence the vault-agent-config map name is taken from there, but the OwnerReferences object is missing.

I think the handling of OwnerReferences object should be separate from the name handling. If pod does not have an OwnerReferences object, one should be created and it should reference the pod, so that the configmap will be removed once the pod is deleted.

/kind bug

marcoreni:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #804, fixes #464, fixes #302

### What's in this PR?
This PR is a WIP/POC to support vault-agent templating capabilities.

I have some open point I'd like to discuss, and something that still needs to be done to make this review-ready.

### Why?
Starting with Vault 1.3.0 Templating capabilities have been ported from consul-template to vault-agent. Therefore, we can replace consul-template sidecar with vault-agent.
There are several advantages to this:
- the init container (and its configmap) is no longer required, since vault-agent has auto_auth capabilities.
- https://github.com/banzaicloud/bank-vaults/issues/804 would be solved since the vault-agent-config configmap is not needed anymore 
- https://github.com/banzaicloud/bank-vaults/issues/464 should be solved since the token would be handled by vault-agent directly (I'm still testing this).
- https://github.com/banzaicloud/bank-vaults/issues/302 would be solved since the full configuration will have to be provided inside the configmap, therefore user will be able to use all auth methods provided by vault-agent to perform authentication.

#### Open Points
The approach I used is to declare an additional boolean parameter called "vault-agent-templating". While this is the easiest way to maintain old codebase, it leaves everything named as "ct"/"consul-template", and this may be misleading since consul-template is not involved anymore. Moreover, the config.hcl file the user provides inside the ConfigMap needs to be extended to contain auto_auth and sink stanzas (because the init container is and the `ct-image` must be different (vault-agent instead of consul-template). Therefore, one cannot simply add the `vault-agent-templating` annotation and be done with it.

One solution may be to define a new `vault-agent-configmap` that triggers the new handling so that it's clearer that "something is different" from an end-user point of view, but with this approach we would recycle the other `ct-...` parameters for a different use and this may be confusing.

A better solution would be to define a whole new set of parameters (eg. prefixed with `va-` instead of `ct-`) but this would yield to a lot of code duplication.

Some proably-non-backwards-compatible solutions are available as well, for example changing the prefix to a common name (`sidecar-...` or similar) and leaving only the specific ones with the appropriate prefixes.

I also stumbled upon vault-agent parameter that may be used for this ( https://github.com/banzaicloud/bank-vaults/issues/302 ).

Discussion and suggestions are highly welcome.

### Checklist

### To Do
- [x] Tests
- [x] Documentation

### DIFF
--- a/cmd/vault-secrets-webhook/main.go
+++ b/cmd/vault-secrets-webhook/main.go
@@ -214,6 +214,77 @@ func getContainers(vaultConfig internal.VaultConfig, containerEnvVars []corev1.E
 	return containers
 }
 
+func getVaContainers(originalContainers []corev1.Container, vaultConfig internal.VaultConfig, containerEnvVars []corev1.EnvVar, containerVolMounts []corev1.VolumeMount) []corev1.Container {
+	var containers = []corev1.Container{}
+	securityContext := &corev1.SecurityContext{
+		AllowPrivilegeEscalation: &vaultConfig.PspAllowPrivilegeEscalation,
+		Capabilities: &corev1.Capabilities{
+			Add: []corev1.Capability{
+				"IPC_LOCK",
+			},
+		},
+	}
+
+	if vaultConfig.VaShareProcess {
+		securityContext = &corev1.SecurityContext{
+			AllowPrivilegeEscalation: &vaultConfig.PspAllowPrivilegeEscalation,
+			Capabilities: &corev1.Capabilities{
+				Add: []corev1.Capability{
+					"SYS_PTRACE",
+					"IPC_LOCK",
+				},
+			},
+		}
+	}
+
+	var serviceAccountMount corev1.VolumeMount
+mountSearch:
+	for _, container := range originalContainers {
+		for _, mount := range container.VolumeMounts {
+			if mount.MountPath == "/var/run/secrets/kubernetes.io/serviceaccount" {
+				serviceAccountMount = mount
+				break mountSearch
+			}
+		}
+	}
+
+	containerVolMounts = append(containerVolMounts, serviceAccountMount, corev1.VolumeMount{
+		Name:      "va-secrets",
+		MountPath: "/vault/secrets",
+	}, corev1.VolumeMount{
+		Name:      "va-configmap",
+		MountPath: "/vault/va-config/config.hcl",
+		ReadOnly:  true,
+		SubPath:   "config.hcl",
+	},
+	)
+
+	var vaCommandString []string
+	if vaultConfig.VaOnce {
+		vaCommandString = []string{"agent", "-config", "/vault/va-config/config.hcl", "-once"}
+	} else {
+		vaCommandString = []string{"agent", "-config", "/vault/va-config/config.hcl"}
+	}
+
+	containers = append(containers, corev1.Container{
+		Name:            "vault-agent",
+		Image:           vaultConfig.VaImage,
+		Args:            vaCommandString,
+		ImagePullPolicy: vaultConfig.VaImagePullPolicy,
+		SecurityContext: securityContext,
+		Env:             containerEnvVars,
+		VolumeMounts:    containerVolMounts,
+		Resources: corev1.ResourceRequirements{
+			Limits: corev1.ResourceList{
+				corev1.ResourceCPU:    vaultConfig.VaCPU,
+				corev1.ResourceMemory: vaultConfig.VaMemory,
+			},
+		},
+	})
+
+	return containers
+}
+
 func getSecurityContext(podSecurityContext *corev1.PodSecurityContext, vaultConfig internal.VaultConfig) *corev1.SecurityContext {
 	if hasPodSecurityContextRunAsUser(podSecurityContext) == true {
 		return &corev1.SecurityContext{
@@ -291,6 +362,38 @@ func getVolumes(existingVolumes []corev1.Volume, agentConfigMapName string, vaul
 					ConfigMap: &corev1.ConfigMapVolumeSource{
 						LocalObjectReference: corev1.LocalObjectReference{
 							Name: vaultConfig.CtConfigMap,
+						},
+						DefaultMode: &defaultMode,
+						Items: []corev1.KeyToPath{
+							{
+								Key:  "config.hcl",
+								Path: "config.hcl",
+							},
+						},
+					},
+				},
+			})
+	}
+
+	if vaultConfig.VaConfigMap != "" {
+		logger.Debugf("Add vault-agent volumes to podspec")
+
+		defaultMode := int32(420)
+		volumes = append(volumes,
+			corev1.Volume{
+				Name: "va-secrets",
+				VolumeSource: corev1.VolumeSource{
+					EmptyDir: &corev1.EmptyDirVolumeSource{
+						Medium: corev1.StorageMediumMemory,
+					},
+				},
+			},
+			corev1.Volume{
+				Name: "va-configmap",
+				VolumeSource: corev1.VolumeSource{
+					ConfigMap: &corev1.ConfigMapVolumeSource{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: vaultConfig.VaConfigMap,
 						},
 						DefaultMode: &defaultMode,
 						Items: []corev1.KeyToPath{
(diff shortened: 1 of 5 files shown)


bonifaido:
Really nice, thank you @marcoreni !

My question is that is vault-agent supporting some kind of signaling that consult-template has so that applications can become aware that the files holding the secrets have changed?

bonifaido:
It would also solve: https://github.com/banzaicloud/bank-vaults/issues/822

marcoreni:
@bonifaido Yes, the template structure is the same as consul-template (actually, it seems that the code is the same). 

Ref: https://www.vaultproject.io/docs/agent/template/index.html
Ref: https://github.com/hashicorp/vault/blob/aaabeff47951374df35c83fe13e0c8d180a1db1e/command/agent/template/template.go#L13

bonifaido:
It would be nice if the two solutions could live together here (the consul-template and the agent as well). Do you think that is possible?

Two reasons for this:
- compatibility/upgrade-path
- some people rely on the signaling capability of consul-template

If yes, I'm happy to merge! 🙂

cmd/vault-secrets-webhook/main.go
```diff
@@ -195,6 +195,24 @@ func getContainers(vaultConfig internal.VaultConfig, containerEnvVars []corev1.E
 		ctCommandString = []string{"-config", "/vault/ct-config/config.hcl"}
 	}
 
+	if vaultConfig.UseAgentTemplating {
+		var serviceAccountMount corev1.VolumeMount
+	mountSearch:
```
bonifaido: This could be moved to a function, since we use the same thing for consul-template.

marcoreni:
@bonifaido yes, keeping both implementations is one of the options I discussed. My only doubt is related to configuration values. If we want to keep backwards compatibility, we need to define a new set of parameters for the vaultAgent option.

bonifaido:
That should be the way forward, so we can't just drop consul-template, and we don't want to introduce breaking changes, so I vote for `va-` prefixes.

bonifaido:
Also, do you have an example where this can be tried out? 🙂

marcoreni:
Hi @bonifaido ,

I refactored the implementation to use `va-..` params as discussed. I also added some tests and documentation.
Here is a sample:
```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    app.kubernetes.io/name: my-app
    my-app.kubernetes.io/name: my-app-vault-agent
    branches: "true"
  name: my-app-vault-agent
data:
  config.hcl: |
    vault {
      retry {
        backoff = "1s"
      }
      auto_auth {
        method "kubernetes" {
          mount_path = "auth/kubernetes"
          config = {
            role = "my-role"
          }
        }
        sink "file" {
          config = {
            path = "/vault/.vault-token"
          }
        }
      }
    }
    template {
      contents = <<EOH
        {{- with secret "database/creds/readonly" }}
        username: {{ .Data.username }}
        password: {{ .Data.password }}
        {{ end }}
      EOH
      destination = "/etc/secrets/config"
      command     = "/bin/sh -c \"kill -HUP $(pidof vault-demo-app) || true\""
    }
---
kind: Pod
apiVersion: v1
metadata:
  name: my-pod
  annotations:
    vault.security.banzaicloud.io/vault-va-configmap: my-app-vault-agent
spec:
  containers:
  - name: main-container
    image: alpine
    command:
    - cat
    tty: true
  serviceAccount: my-service-account
```

cmd/vault-secrets-webhook/main.go
```diff
@@ -470,6 +567,57 @@ func parseVaultConfig(obj metav1.Object) internal.VaultConfig {
 		vaultConfig.TransitPath = val
 	}
 
+	if val, ok := annotations["vault.security.banzaicloud.io/vault-va-configmap"]; ok {
+		vaultConfig.VaConfigMap = val
+	} else {
+		vaultConfig.VaConfigMap = ""
+	}
+
+	if val, ok := annotations["vault.security.banzaicloud.io/vault-va-image"]; ok {
```
bonifaido: I think we can reuse the `"vault_image"` property here, and remove this one.

cmd/vault-secrets-webhook/main.go
```diff
@@ -470,6 +567,57 @@ func parseVaultConfig(obj metav1.Object) internal.VaultConfig {
...
+		vaultConfig.VaImage = val
+	} else {
+		vaultConfig.VaImage = viper.GetString("vault_va_image")
+	}
+
+	if val, ok := annotations["vault.security.banzaicloud.io/vault-va-pull-policy"]; ok {
```
bonifaido: Applies to this one as well ("vault_image_pull_policy").

bonifaido:
Two minor comments, otherwise I'm ready to merge @marcoreni ! 🎉

cmd/vault-secrets-webhook/main.go
```diff
@@ -943,6 +1134,8 @@ func init() {
 	viper.SetDefault("registry_skip_verify", "false")
 	viper.SetDefault("debug", "false")
 	viper.SetDefault("enable_json_log", "false")
+	viper.SetDefault("vault_va_image", "hashicorp/vault:latest")
```
bonifaido: The value should be just `vault:latest`, but when reusing `"vault_image"` this will be solved automatically.

cmd/vault-secrets-webhook/main.go
```diff
@@ -214,6 +217,68 @@ func getContainers(vaultConfig internal.VaultConfig, containerEnvVars []corev1.E
...
+	}
+
+	serviceAccountMount := getServiceAccountMount(originalContainers)
+
+	containerVolMounts = append(containerVolMounts, serviceAccountMount, corev1.VolumeMount{
+		Name:      "va-secrets",
+		MountPath: "/vault/secrets",
+	}, corev1.VolumeMount{
+		Name:      "va-configmap",
+		MountPath: "/vault/va-config/config.hcl",
```
bonifaido: Can we mount this to /vault/config/ instead?

cmd/vault-secrets-webhook/main.go
```diff
@@ -214,6 +217,68 @@ func getContainers(vaultConfig internal.VaultConfig, containerEnvVars []corev1.E
...
+		ReadOnly:  true,
+		SubPath:   "config.hcl",
+	},
+	)
+
+	var vaCommandString []string
+	if vaultConfig.VaOnce {
+		vaCommandString = []string{"agent", "-config", "/vault/va-config/config.hcl", "-once"}
```
bonifaido: And then change the path here as well.

bonifaido:
In the meantime, I managed to modify the example to work with the current examples in the repo:

```
apiVersion: v1
kind: ConfigMap
metadata:
    labels:
        app.kubernetes.io/name: my-app
        my-app.kubernetes.io/name: my-app-vault-agent
        branches: "true"
    name: my-app-vault-agent
data:
    config.hcl: |
        vault {
            retry {
                backoff = "1s"
            }
        }
        auto_auth {
            method "kubernetes" {
                mount_path = "auth/kubernetes"
                config = {
                    role = "default"
                }
            }
            sink "file" {
                config = {
                    path = "/vault/.vault-token"
                }
            }
        }
        template {
            contents = <<EOH
            {{- with secret "secret/mysql" }}
            username: {{ .Data.MYSQL_PASSWORD }}
            password: {{ .Data.MYSQL_ROOT_PASSWORD }}
            {{ end }}
            EOH
            destination = "/etc/secrets/config"
            command     = "/bin/sh -c \"kill -HUP $(pidof vault-demo-app) || true\""
        }
---
kind: Pod
apiVersion: v1
metadata:
    name: my-pod
    annotations:
        vault.security.banzaicloud.io/vault-addr: https://vault:8200
        vault.security.banzaicloud.io/vault-tls-secret: "vault-tls"
        vault.security.banzaicloud.io/vault-va-configmap: my-app-vault-agent
spec:
    containers:
    - name: main-container
      image: alpine
      command:
      - cat
      tty: true
    serviceAccount: default
```

Although, there is an ongoing issue with the vault-agent TLS settings towards vault: https://github.com/hashicorp/vault/issues/7889 which still occurred during my testings when using a custom CA.

marcoreni:
Hey @bonifaido , I refactored to use `agent-` prefix and I changed the references from `Va` to `Agent` inside the code.

Here is the updated example:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
    labels:
        app.kubernetes.io/name: my-app
        my-app.kubernetes.io/name: my-app-vault-agent
        branches: "true"
    name: my-app-vault-agent
data:
    config.hcl: |
        vault {
            retry {
                backoff = "1s"
            }
        }
        auto_auth {
            method "kubernetes" {
                mount_path = "auth/kubernetes"
                config = {
                    role = "default"
                }
            }
            sink "file" {
                config = {
                    path = "/vault/.vault-token"
                }
            }
        }
        template {
            contents = <<EOH
            {{- with secret "secret/mysql" }}
            username: {{ .Data.MYSQL_PASSWORD }}
            password: {{ .Data.MYSQL_ROOT_PASSWORD }}
            {{ end }}
            EOH
            destination = "/etc/secrets/config"
            command     = "/bin/sh -c \"kill -HUP $(pidof vault-demo-app) || true\""
        }
---
kind: Pod
apiVersion: v1
metadata:
    name: my-pod
    annotations:
        vault.security.banzaicloud.io/vault-addr: https://vault:8200
        vault.security.banzaicloud.io/vault-tls-secret: "vault-tls"
        vault.security.banzaicloud.io/vault-agent-configmap: my-app-vault-agent
spec:
    containers:
    - name: main-container
      image: alpine
      command:
      - cat
      tty: true
    serviceAccount: default
```

Let me know if there's anything else that can be improved! :)

PR #825, opened 2020-01-09
Add registry address to cache key

Kasama:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | none

### What's in this PR?

Change the cache key of the ECR token to use ECR registry URL.

### Why?

When using ECR images from multiple regions, the ECR key would be the same, causing an authentication problem, because the previously generated token was created for only one region

### DIFF
--- a/cmd/vault-secrets-webhook/registry/registry.go
+++ b/cmd/vault-secrets-webhook/registry/registry.go
@@ -369,7 +369,7 @@ func (k *ContainerInfo) Collect(container *corev1.Container, podSpec *corev1.Pod
 			logger.Infof("trying to request AWS credentials for ECR registry %s", k.RegistryAddress)
 
 			var data string
-			cachedToken, usingCache := credentialsCache.Get(ecrCredentialsKey)
+			cachedToken, usingCache := credentialsCache.Get(ecrCredentialsKey + k.RegistryAddress)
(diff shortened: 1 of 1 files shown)


cmd/vault-secrets-webhook/registry/registry.go
```diff
@@ -369,7 +369,7 @@ func (k *ContainerInfo) Collect(container *corev1.Container, podSpec *corev1.Pod
 			logger.Infof("trying to request AWS credentials for ECR registry %s", k.RegistryAddress)
 
 			var data string
-			cachedToken, usingCache := credentialsCache.Get(ecrCredentialsKey)
+			cachedToken, usingCache := credentialsCache.Get(ecrCredentialsKey + k.RegistryAddress)
```
ramonberrutti: Create a var: 
```go
tokenKey := ecrCredentialsKey + k.RegistryAddress
cachedToken, usingCache := credentialsCache.Get(tokenKey)

// ... 

credentialsCache.Set(tokenKey, data, expiration)
```

bonifaido: 👍 for that!

Issue #836, opened 2020-01-16
Renew dynamic secrets (env vars)


dklesev:
**Is your feature request related to a problem? Please describe.**
Yes.
As of now to be able to renew the lease of dynamic secrets one needs to use consul templates. However, it seems not to be possible to renew a lease of an environment variable.

**Describe the solution you'd like**
1. Add ability to vault-env to retrieve the `lease_id`, `lease_duration`, `lease_renewable`.
2. Save it to a shared directory (empty-dir).
3. Start sidecar container watching the shared directory and renew all leases within it.
4. If no dynamic secrets are used, steps above are not needed.

**Describe alternatives you've considered**
Parts of this:
https://medium.com/@gmaliar/dynamic-secrets-on-kubernetes-pods-using-vault-35d9094d169

**Additional context**
RabbitMQ:
```
$ vault read rabbitmq/creds/my-role
Key                Value
---                -----
lease_id           rabbitmq/creds/my-role/XEl5Chkthg2xiGgMgfs7CFbT
lease_duration     336h
lease_renewable    true
password           e98ea98d-c6fe-7b1a-fe75-d1f166542494
username           root-cc0f6899-8c68-9173-904b-c4d479cd1c68
```

bonifaido:
Hi @dklesev, this is something that we are currently working on actively, it will come in the following weeks, stay tuned! 🤓

bonifaido:
FYI: https://github.com/banzaicloud/bank-vaults/pull/856

cablespaghetti:
@bonifaido Good work! However can you explain how this would work, i can't quite work it out from the code? I don't see any code to restart the application to pick up the updated environment variable.

bonifaido:
Hi @cablespaghetti, I think we don't need to restart the application in this case, since the environment variables won't change, for example, you have a dynamic database credentials engine enabled, with credential lease time of 10minutes, the job of vault env would be to call vault within 10m to renew the lease and the credential will not change in this case.

bonifaido:
Yeah I guess I know what you mean, so what happens after max_ttl expires. :)

bonifaido:
That should be added to the implementation as well! :)

cablespaghetti:
I'm very new to Vault and haven't really explored dynamic secrets so I didn't know about renewing the secret without actually changing the value. Glad I helped though! 😁

rgaoistheone:
Sorry , just bump into this similar requests .
[...]

PR #856, opened 2020-01-30
vault-env: add daemon mode with dynamic secret renewal

bonifaido:

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/banzaicloud/bank-vaults/issues/836

### What's in this PR?

Add the functionality to start `vault-env` in a so-called daemon mode, so it doesn't replace itself with the user process after setting the environment variables, but executes it as a child-process and remains in memory, so it can renew the leases on Vault tokens and on dynamic secrets.

### Why?

To be able to use dynamic secrets in long-running processes with `vault-env`.

### Additional context

The functionality can be enabled with the `vault.security.banzaicloud.io/vault-env-daemon: "true"` annotation, the old functionality remains the default option.

**The current limitation is if the max_ttl of the Secret from Vault expires the renewal stops as well, so the Secrets get revoked.**

### Checklist

### DIFF
--- a/cmd/vault-env/main.go
+++ b/cmd/vault-env/main.go
@@ -21,12 +21,13 @@ import (
 	"strings"
 	"syscall"
 
-	"github.com/banzaicloud/bank-vaults/internal/configuration"
-	"github.com/banzaicloud/bank-vaults/pkg/sdk/vault"
-
+	"emperror.dev/errors"
 	vaultapi "github.com/hashicorp/vault/api"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cast"
+
+	"github.com/banzaicloud/bank-vaults/internal/configuration"
+	"github.com/banzaicloud/bank-vaults/pkg/sdk/vault"
 )
 
 // The special value for VAULT_ENV which marks that the login token needs to be passed through to the application
@@ -61,6 +62,7 @@ var (
 		"VAULT_ENV_PASSTHROUGH":        true,
 		"VAULT_JSON_LOG":               true,
 		"VAULT_REVOKE_TOKEN":           true,
+		"VAULT_ENV_DAEMON":             true,
 	}
 
 	logger *log.Logger
@@ -89,8 +91,22 @@ func (environ *sanitizedEnviron) append(name string, value string) {
 	}
 }
 
+func startSecretRenewal(client *vault.Client, secret *vaultapi.Secret) error {
+	renewerInput := vaultapi.RenewerInput{Secret: secret}
+	secretRenewer, err := client.RawClient().NewRenewer(&renewerInput)
+	if err != nil {
+		return errors.Wrap(err, "failed to create secret renewer")
+	}
+
+	go secretRenewer.Renew()
+
+	return nil
+}
+
 func main() {
 	enableJSONLog := os.Getenv("VAULT_JSON_LOG")
+
+	daemonMode := cast.ToBool(os.Getenv("VAULT_ENV_DAEMON"))
 
 	logger = log.New()
 	// Add additional fields to all log messages
@@ -241,6 +257,14 @@ func main() {
 					secretCache[secretCacheKey] = secret
 				}
 			}
+
+			if daemonMode && secret.Renewable && secret.LeaseDuration > 0 {
+				log.Infof("secret %q has a lease duration of %ds, starting renewal", valuePath, secret.LeaseDuration)
+				err = startSecretRenewal(client, secret)
+				if err != nil {
+					logger.Fatalln("secret renewal can't be established", err)
+				}
+			}
 		}
 
 		if secret == nil {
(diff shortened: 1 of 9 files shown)


cmd/vault-env/main.go
```diff
@@ -89,9 +92,35 @@ func (environ *sanitizedEnviron) append(name string, value string) {
 	}
 }
 
+func startSecretRenewal(client *vault.Client, path string, secret *vaultapi.Secret) error {
+	renewerInput := vaultapi.RenewerInput{Secret: secret}
+	secretRenewer, err := client.RawClient().NewRenewer(&renewerInput)
```
primeroz: is "NewRenewer" deprecated ? 
https://godoc.org/github.com/hashicorp/vault/api#Client.NewRenewer

cmd/vault-env/main.go
```diff
@@ -89,9 +92,35 @@ func (environ *sanitizedEnviron) append(name string, value string) {
...
+	go secretRenewer.Renew()
+
+	go func() {
+		for {
+			select {
+			case renewOutput := <-secretRenewer.RenewCh():
+				log.Infof("secret %s renewed for %ds", path, renewOutput.Secret.LeaseDuration)
+			case doneError := <-secretRenewer.DoneCh():
+				log.WithField("error", doneError).Infof("secret renewal for %s has finished", path)
+				return
```
primeroz: Will this cause the pod to fail if the renewal fail for any reason ? ( network error for example ? ) 

Ideally we don't want a pod keep running if vault-env can't renew the lease for any reason.

cmd/vault-env/main.go
bonifaido: Yes, it is, but this version of the api package is not released yet by Hashicorop, the latest release is from last summer (github.com/hashicorp/vault/api 1.0.4). If you think we can pull in the master version and pin it, I have no issues against it.

cmd/vault-env/main.go
bonifaido: Yes, possibly! Good point, I will fix this.

cmd/vault-env/main.go
primeroz: ah i did not see that.

No I think we should stick to the current taggeed version and change it later on

cmd/vault-env/main.go
bonifaido: Fixed.

cmd/vault-env/main.go
```diff
@@ -89,9 +93,43 @@ func (environ *sanitizedEnviron) append(name string, value string) {
...
+			case doneError := <-secretRenewer.DoneCh():
+				log.WithField("error", doneError).Infof("secret renewal for %s has stopped, sending SIGTERM to process", path)
+
+				sigs <- syscall.SIGTERM
+
+				timeout := <-time.After(10 * time.Second)
+				log.Infoln("killing process due to SIGTERM timeout =", timeout)
+				sigs <- syscall.SIGKILL
+
+				return
```
primeroz: Going to ask a question just for safeness sake.

I understand this will pipe SIGTERM and then after a timeout SIGKILL into the sigs channel and will signal the child process, so it will shutdown either cleanly or hardly.

What will kill the `vault-env` wrapper though ? 
Do we need to return `errors.Wrap ("something something something dark")` here ?

moskitone:
When there is non existing secret (typo in path) we get this error 

`time="2020-02-05T12:28:32Z" level=info msg="received new Vault token"
time="2020-02-05T12:28:32Z" level=info msg="initial Vault token arrived"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x113579a]
goroutine 1 [running]:
main.main()
	/build/cmd/vault-env/main.go:283 +0x22ba`

cmd/vault-env/main.go
bonifaido: 👍

bonifaido:
Thanks @moskitone, just pushed a fix for that. Nice catch!

cmd/vault-env/main.go
bonifaido: Vault-env will be exiting if the child process exits right after (or itself receives a SIGKILL from Kubernetes). Maybe we should add a handler for SIGTERM as well to vault-env, so it can close the token and secret renewal cleanly?

primeroz: I see. 

Well, i am always up for a cleaner shutdown

bonifaido: We can't return an error from here, this is a background thread, which will shut down after sending the SIGKILL.

Vault-env will quit because of the sub-process exited after the SIGKILL, here: https://github.com/banzaicloud/bank-vaults/blob/84387878c43d6cc52319c141a35dcbf8a3cf4a22/cmd/vault-env/main.go#L372

primeroz: sounds good

bonifaido: I'm planning to merge this if there are no more comments.

moskitone:
Thanks

PR #859, opened 2020-02-05
Ability to use custom TLS secret

### LINKED ISSUE #781: Override Vault Default TLS Parameters (CA, Cert, and Private Key)
I have tried changing the default certificate and key to custom ones.  The Vault instances fail to initialize, I believe because the Vault commands no longer know the proper CA Root Certificate to use.

I would like to be able to override the environment variable `VAULT_CACERT`, or have some other method of specifying the secret to use for certificates in the system.

Perhaps being able to replace the `vault-tls` secret with a `cert-manager` generated secret would work much better.

Since `vault-tls` secrets and use of `VAULT_CACERT` appear hardcoded, I don't see an easy alternative.

**Additional context**
N/A

pbalogh-sa:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #781 

### What's in this PR?
- We can define existing secret contains TLS certificate (accepted secret type: kubernetes.io/tls)
- If it is set, generating certificate will be disabled


docs/tls/README.md
```diff
@@ -69,3 +69,41 @@ helm upgrade --install vault ../charts/vault --set tls.secretName=vault-tls
 ```bash
 kubectl apply -f vault-cr.yaml
 ```
+
+## Generating custom certificates with cert-manager for Bank-Vaults
```
bonifaido: I would probably move this to a subsection of the one underneath (`## Using the generated custom TLS certificate with vault-operator:`).

pbalogh-sa: OK

gnadaban:
@bonifaido : Just a suggestion: you should consider [squashing commits](https://blog.carbonfive.com/2017/08/28/always-squash-and-rebase-your-git-commits/) in PRs before merging if you want to keep your Git history clean.

Issue #863, opened 2020-02-07
bank-vaults support for existing HashiCorp Vault users


Mel-Github:
**Is your feature request related to a problem? Please describe.**
Some of the existing users have already an existing HashiCorp Vault instance running in their environment. However, currently to leverage the bank-vaults solutions we need to create a new vault instance via the custom crd. 

**Describe the solution you'd like**
I would like to still leverage the vault bank solution but running on top of my existing Hashicorp vault instance as there are hundreds of secrets already in my existing vault system.

mazzy89:
I've explored this solution right today and here my findings that can be obvious but useful to report:

- the bank-vaults codebase is pretty easy to follow and at high-level expose two high-level commands that can be consumed:
 
  * `init`
  * `unseal`

These two commands make the Vault operations easy and straightforward. 

- bank-vaults is distributed as a Docker image so easy to run in any environment 

- the `unseal` command refers to a variable `POD_NAME` fetched by the environment. this makes the tool not really platform-agnostic. Although IMHO is not a big deal because thanks to the interface a small change here can be applied and make the command platform-agnostic easily.

Am I missing anything here? I would be keen to make the adjustments where necessary so this nice and handy tool can be used also in a non-kubernetes environment or in existing Vault clusters non-running on Kubernetes.

mazzy89:
Would make it make sense here to change the line https://github.com/banzaicloud/bank-vaults/blob/a1c1d5f4e76bd73eb96188eabdeedb6905b6d8bf/cmd/bank-vaults/unseal.go#L106

in

```go
if !v.isInit() && !unsealConfig.raftSecondary {
...
```

in this way instead, to check the `POD_NAME` we would check the state using the following API request https://www.vaultproject.io/api/system/init.html

@bonifaido Do you see anything wrong in this?

bonifaido:
Hi @mazzy89,

yes, this is right, you got the point.

For the second question, I'm not sure if just calling the API alone would be enough here, how would you decide in case of the first instance that it should be initialized or joining the cluster?

mazzy89:
I was thinking that since there is a built-in back-off mechanism, the init would work in a way that the first instance that is up, initializes the cluster and all the other instances just follow joining the Raft cluster.

bonifaido:
And how do you decide in the case of the second instance that it should join instead of initing?

mazzy89:
[...]

PR #868, opened 2020-02-10
Make unseal cmd platform-agnostic

mazzy89:
Signed-off-by: Salvatore Mazzarino <dev@mazzarino.cz>

| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | #863 

### What's in this PR?

`bank-vaults` expects as env variable the `POD_NAME` and in case this is the first pod of a StatefulSet (in kubernetes this is suffixed with `-0`), it attempts to initialize.

`bank-vaults` could be used not only in a Kubernetes environment but also in the existing Vault cluster where is very useful to automate the Raft join of nodes.

In this scenario, as discussed with @bonifaido, we can verify if Vault is initialized by checking whether the keys stored in the configured Cloud KMS are in place. 

### Additional context
Does this PR break any existing cluster? I'd say no.

### Checklist

### DIFF
--- a/pkg/sdk/vault/operator_client.go
+++ b/pkg/sdk/vault/operator_client.go
@@ -358,6 +358,10 @@ func (v *vault) Init() error {
 func (v *vault) RaftInitialized() (bool, error) {
 	rootToken, err := v.keyStore.Get(v.rootTokenKey())
 	if err != nil {
+		if e, ok := err.(notFoundError); ok && e.NotFound() {
+			return false, nil
+		}
+
 		return false, fmt.Errorf("unable to get key '%s': %s", v.rootTokenKey(), err.Error())
 	}
 
--- a/pkg/sdk/vault/operator_client.go
+++ b/pkg/sdk/vault/operator_client.go
@@ -358,7 +358,7 @@ func (v *vault) Init() error {
 func (v *vault) RaftInitialized() (bool, error) {
 	rootToken, err := v.keyStore.Get(v.rootTokenKey())
 	if err != nil {
-		return fmt.Errorf("unable to get key '%s': %s", v.rootTokenKey(), err.Error())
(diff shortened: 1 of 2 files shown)


pkg/sdk/vault/operator_client.go
```diff
@@ -353,6 +354,20 @@ func (v *vault) Init() error {
 	return nil
 }
 
+// in our case Vault is initialized when root key is stored in the Cloud KMS
+func (v *vault) isInit() (bool, error) {
+	rootToken, err := v.keyStore.Get(v.rootTokenKey())
```
mazzy89: Should I check only the root key existence?

bonifaido:
In a few seconds the raft-test branch will be part of master, please rebase once I get to merge it :)

pkg/sdk/vault/operator_client.go
bonifaido: yes, it should be enough.

pkg/sdk/vault/operator_client.go
```diff
@@ -84,6 +84,7 @@ var _ Vault = &vault{}
 // a Vault server.
 type Vault interface {
 	Init() error
+	isInit() (bool, error)
```
bonifaido: I would call this method `RaftInitialized()` to be in pair with RaftJoin()

mazzy89:
@bonifaido tests green here. waiting for your tests to be merged, rebease and fingers crossed that everything is green after.

bonifaido:
I just merged master to your branch, where the raft test is in 🤞

mazzy89:
@bonifaido tests all green whoa 😇 squash and merge, right?

mazzy89:
@bonifaido thanks for your tests ;-)

Issue #871, opened 2020-02-11
Automate the pulling of Raft leader address


mazzy89:
**Is your feature request related to a problem? Please describe.**

Automate the process and avoid to pass as input the Raft leader address.

**Describe the solution you'd like**

Other nodes joining the cluster would pull the Raft Leader cluster before joining.

**Describe alternatives you've considered**
Having the logic of retrieving Raft leader address external to `bank-vaults`. 

**Additional context**

I've spoken already with @bonifaido and we agreed that this feature is nice to have.

The idea is to retrieve the leader address using `/sys/leader` and then store it in the unsealed KV backend.

Question: what should happen when the leader address change?

bonifaido:
> Question: what should happen when the leader address change?

Basically I think we don't need to care about it since Vault just needs to use it once when joining only.

mazzy89:
It makes sense. nice

mazzy89:
@bonifaido so to be clear about the flow:

- if `raft-leader-address` flag is not passed in then retrieve the raft leader automatically

- call `/sys/leader` and get the leader address

talking about store the value. why do we want to store it? do we want to use it for later usage? can't we consume it immediately bypassing to the `RaftJoin()` function?

bonifaido:
I think that unseal KV backend and Cloud KMS are the same here in this context.

mazzy89:
Right because all are built as an abstraction of KV service. thank you. all clear now.

bonifaido:
Closing this for now, as the original PR because closed as well, and the problem got resolved on the user side.

PR #875, opened 2020-02-12
bank-vaults: retrieve automatically Raft leader adress

mazzy89:
Signed-off-by: Salvatore Mazzarino <dev@mazzarino.cz>

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | https://github.com/banzaicloud/bank-vaults/issues/871

### What's in this PR?

During the `init` process store the Raft leader API address in the Cloud KMS and retrieve it before joining the cluster

### Why?

This would increase the automation level and makes the tool avoid asking for the raft leader address as input.

### Checklist

- [ ] Error handling code meets the [guideline](https://github.com/banzaicloud/pipeline/blob/master/docs/error-handling-guide.md)

### To Do

### DIFF
--- a/cmd/bank-vaults/unseal.go
+++ b/cmd/bank-vaults/unseal.go
@@ -112,6 +112,14 @@ from one of the followings:
 					logrus.Fatalf("error initializing vault: %s", err.Error())
 				}
 			} else {
+				if unsealConfig.raftLeaderAddress == "" {
+					logrus.Info("retrieving raft leader IP address...")
+					unsealConfig.raftLeaderAddress, err = v.RaftLeaderAPIAddr()
+					if err != nil {
+						logrus.Fatalf("error retrieving raf leader IP address: %s", err.Error())
+					}
+				}
+
 				logrus.Info("joining raft cluster...")
 				if err := v.RaftJoin(unsealConfig.raftLeaderAddress); err != nil {
 					logrus.Fatalf("error joining leader vault: %s", err.Error())
--- a/cmd/bank-vaults/unseal.go
+++ b/cmd/bank-vaults/unseal.go
@@ -112,16 +112,18 @@ from one of the followings:
 					logrus.Fatalf("error initializing vault: %s", err.Error())
 				}
 			} else {
+				raftLeaderAddress := unsealConfig.raftLeaderAddress
 				if unsealConfig.raftLeaderAddress == "" {
 					logrus.Info("retrieving raft leader IP address...")
-					unsealConfig.raftLeaderAddress, err = v.RaftLeaderAPIAddr()
+					raftLeaderAddress, err = v.RaftLeaderAPIAddr()
+
 					if err != nil {
 						logrus.Fatalf("error retrieving raf leader IP address: %s", err.Error())
 					}
 				}
 
 				logrus.Info("joining raft cluster...")
-				if err := v.RaftJoin(unsealConfig.raftLeaderAddress); err != nil {
+				if err := v.RaftJoin(raftLeaderAddress); err != nil {
 					logrus.Fatalf("error joining leader vault: %s", err.Error())
 				}
 			}
--- a/cmd/bank-vaults/unseal.go
+++ b/cmd/bank-vaults/unseal.go
@@ -105,6 +105,8 @@ from one of the followings:
 				logrus.Fatalf("error checking if vault is initialized: %s", err.Error())
 			}
 
+			logrus.Info("Vault initialized", initialized)
+
 			// If this is the first instance we have to init it, this happens once in the clusters lifetime
 			if !initialized && !unsealConfig.raftSecondary {
 				logrus.Info("initializing vault...")
--- a/cmd/bank-vaults/unseal.go
+++ b/cmd/bank-vaults/unseal.go
@@ -105,7 +105,7 @@ from one of the followings:
 				logrus.Fatalf("error checking if vault is initialized: %s", err.Error())
 			}
 
-			logrus.Info("Vault initialized", initialized)
+			logrus.Info("Vault initialized: ", initialized)
 
 			// If this is the first instance we have to init it, this happens once in the clusters lifetime
 			if !initialized && !unsealConfig.raftSecondary {
--- a/cmd/bank-vaults/unseal.go
+++ b/cmd/bank-vaults/unseal.go
@@ -113,7 +113,7 @@ from one of the followings:
 				}
 			} else {
 				raftLeaderAddress := unsealConfig.raftLeaderAddress
-				if unsealConfig.raftLeaderAddress == "" {
+				if raftLeaderAddress == "" {
 					logrus.Info("retrieving raft leader IP address...")
 					raftLeaderAddress, err = v.RaftLeaderAPIAddr()
 
--- a/pkg/sdk/vault/operator_client.go
+++ b/pkg/sdk/vault/operator_client.go
@@ -66,6 +66,9 @@ type Config struct {
 
 	// should the KV backend be tested first to validate access rights
 	PreFlightChecks bool
+
+	// if vault runs as storage backend Raft
+	Raft bool
 }
 
 // vault is an implementation of the Vault interface that will perform actions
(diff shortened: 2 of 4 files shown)


cmd/bank-vaults/unseal.go
```diff
@@ -108,18 +108,26 @@ from one of the followings:
...
+
 				logrus.Info("joining raft cluster...")
 				if err := v.RaftJoin(unsealConfig.raftLeaderAddress); err != nil {
 					logrus.Fatalf("error joining leader vault: %s", err.Error())
 				}
 			}
 		} else if unsealConfig.proceedInit {
 			logrus.Info("initializing vault...")
-			if err := v.Init(); err != nil {
+			if err := v.Init(unsealConfig.raft); err != nil {
```
mazzy89: I'm not sure if this is the best thing here. this would require to change the interface of `Init()` everywhere also in the `init.go` a command where it is not needed.

pkg/sdk/vault/operator_client.go
```diff
@@ -297,6 +298,21 @@ func (v *vault) Init() error {
 
 	rootToken := resp.RootToken
 
+	if isRaft {
+		resp, err := v.cl.Sys().Leader()
+		if err != nil {
+			return false, fmt.Errorf("error checking leader: %s", err.Error())
+		}
+
+		leaderAPIAddr := resp.LeaderAddress
```
mazzy89: `resp` returns two properties

```json
{
  "ha_enabled": true,
  "is_self": false,
  "leader_address": "https://127.0.0.1:8200/",
  "leader_cluster_address": "https://127.0.0.1:8201/",
  "performance_standby": false,
  "performance_standby_last_remote_wal": 0
}
```

In this case I assume we would need the `leader_address`. right?

cmd/bank-vaults/unseal.go
mazzy89: @bonifaido what do you think here?

pkg/sdk/vault/operator_client.go
mazzy89: I found confirmation here https://learn.hashicorp.com/vault/operations/raft-storage#join-nodes-to-the-cluster. to join the cluster is needed the `leader_address`.

pkg/sdk/vault/operator_client.go
```diff
@@ -297,6 +298,19 @@ func (v *vault) Init() error {
 
 	rootToken := resp.RootToken
 
+	leaderResp, err := v.cl.Sys().Leader()
```
mazzy89: in case that we don't use Raft, calling /sys/leader would not return a proper address?

@bonifaido what do you think here?

mazzy89:
I'm not sure whether during the acceptance tests this case is run. I'd like to see if not passing raft leader address as input, things work as expected.

bonifaido:
You can run the tests manually, by building the required docker images and running `./hack/acceptance-test.sh`. If you check the CI config you can see how it goes.

mazzy89:
I've noticed that at the moment the operator generates the command with always the `--raft-leader-address` flag as below

```shell
bank-vaults unseal --init --raft --raft-leader-address https://vault:8200
```

In order to test this, it would require to modify the behavior of the operator here https://github.com/banzaicloud/bank-vaults/blob/master/operator/pkg/controller/vault/vault_controller.go#L1156

Which is your opinion here @bonifaido?

bonifaido:
You mean you would pass only this value if it is defined in the CR right @mazzy89 ?

mazzy89:
Yeah exactly

bonifaido:
I'm happy with that, please go ahead if you need to change that.

bonifaido:
Any roadblocks on this @mazzy89 I can help to resolve?

mazzy89:
Hey @bonifaido thank you to follow back on this. currently, I don't have much capacity to finish this feature. I'll get back on it on Monday. 

I tried last week to take a look at it and I got stuck because I could not find the specific reason of why acceptance tests at this stage are failing. 

Need to modify the operator so the flag `raft-leader-address` is not always loaded and then I'll give it another run.

pkg/sdk/vault/operator_client.go
bonifaido: I think a leader address is returned in any case if Vault is running in HA mode (Raft or not). But we can determine if we are running in Raft mode via: `unsealConfig.raft` and pass that information down here.

bonifaido:
No problem @mazzy89, take your time!

pkg/sdk/vault/operator_client.go
mazzy89: I was thinking to pass `unsealconfig.raft` down inside the `Init()` function. however, this would lead us to change the contract of the `Init()`. `Init()` is used also in place whee we don't pass `raft` flag. how would we deal with that?

bonifaido: I would pass this information to the Vault client instead and read it in `Init()`.

mazzy89:
@bonifaido the build fails

```sh
make license-check
bin/licensei check
github.com/opentracing/opentracing-go: no license file found
github.com/coreos/etcd-operator: no license file found
github.com/coreos/prometheus-operator: no license file found
github.com/mailru/easyjson: no license file found
github.com/go-openapi/swag: no license file found
github.com/go-openapi/jsonpointer: no license file found
github.com/PuerkitoBio/urlesc: no license file found
github.com/PuerkitoBio/purell: no license file found
github.com/go-openapi/jsonreference: no license file found
github.com/go-openapi/spec: no license file found
github.com/emicklei/go-restful: no license file found
k8s.io/kube-openapi: no license file found
github.com/banzaicloud/k8s-objectmatcher: no license file found
github.com/coreos/etcd: no license file found
go.uber.org/zap: no license file found
github.com/coreos/go-systemd: no license file found
github.com/coreos/pkg: no license file found
github.com/google/go-cmp: no license file found
github.com/golang/groupcache: no license file found
gomodules.xyz/jsonpatch/v2: no license file found
github.com/go-logr/zapr: no license file found
make: *** [main-targets.mk:63: license-check] Error 2

Exited with code exit status 2
```

pkg/sdk/vault/operator_client.go
mazzy89: Moved `Raft` flag into Vaul `Config`.

bonifaido:
Please ignore the Circle check, only the GitHub check works currently, which overcomes this issue, but fails in the acceptance test phase: https://github.com/banzaicloud/bank-vaults/pull/875/checks?check_run_id=464523365

mazzy89:
NIce. then I will check why acceptance test still doesn't pass

mazzy89:
Seems here that the issue is due to the fact that the pod `vault-1` gets into `CrashLoopBackOff`

```
NAME                                READY   STATUS             RESTARTS   AGE
vault-0                             3/3     Running            1          44s
vault-1                             1/3     CrashLoopBackOff   1          23s
vault-configurer-77d46f9f77-cfcwf   1/1     Running            0          43s
vault-operator-664779c59b-zrp8p     1/1     Running            0          3m31s
```

In particular, this is due to the error:

```
+ kubectl logs --all-containers pod/vault-1
time="2020-03-03T16:56:28Z" level=fatal msg="error initializing vault: error before init: value for key 'vault-unseal-0' already exists"
```

mazzy89:
It seems that the changes in this PR introduced somehow a regression. In fact, the pod `vault-1` tries to initialize again when actually Vault has been already initialized with the pod `vault-0`. The pod `vault-0` is the one that writes the root token key in the KMS, in this case, it should be the file storage if I'm not wrong.

In my previous PR, I've introduced that concept of `isInitialized` when the root token key is written in the Cloud KMS. 

The acceptance tests pass and the master is not in a broken state AFAIK so I'm missing the reason why this behavior.

The only thing that comes in my mind is some race condition. When the pod `vault-1` checks if it is initialized the key is still not created by the pod `vault-0`. But this hypothesis does not explain why tests in master passes.

mazzy89:
A bit of debug here.

both the pods `vault-0` and `vault-1` reports this

```
time="2020-03-04T09:32:22Z" level=info msg="Root token: []"
time="2020-03-04T09:32:22Z" level=info msg="NotFound: true"
time="2020-03-04T09:32:22Z" level=info msg="Vault initialized: false"
```
I would expect this from the `vault-0` according the timestamp but not from `vault-1`.

mazzy89:
So the issue here is that k8s secret `vault-root` is not created so `RaftInitialized()` returns always `false`.

There is something that I really don't get. From the logs https://github.com/banzaicloud/bank-vaults/pull/875#issuecomment-594068916 it seems that the pod `vault-0` receives from Vault that it is already initialized. While the pod `vault-1` by querying Vault receives that Vault is not initialized. 

I really don't how this would be possible when clearly the `vault-0` runs before the pod `vault-1`. Parking this here because I've run out of ideas. Looking for feedbacks.

mazzy89:
Closing this. Solved this in a different way.

PR #895, opened 2020-02-26
Don't add IPC_LOCK capability when disable_mlock is true in Vault config

### LINKED ISSUE #889: Run Vault in unprivileged SCC group
We are unable start Vault in unprivileged group.

If we  start Vault with disable_mlock and SKIP_SETCAP=true, we will face to the necessarity of  running Vault under SCC with "allowedCapabilities: - IPC_LOCK" becaue it capability is included into securitycontext of statefulset. 
I found that it is hardcoded into code of opearotor and it will be added in any case:
https://github.com/banzaicloud/bank-vaults/blob/3908ad73b09670131f921cc580b8fe743299ecf4/operator/pkg/controller/vault/vault_controller.go#L1218-L1221

Could you please add the opportunity to create vault statesfulset without IPC_LOCK in cases if we disable it?

pbalogh-sa:
Signed-off-by: Peter Balogh <p.balogh.sa@gmail.com>

| Q               | A
| --------------- | ---
| Bug fix?        |no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #889 

### What's in this PR?
If in the vault config section `disable_mlock` is set to true, the operator doesn't add the IPC_lock capability to `ContainerSecurityContext` of the Vault container.

### DIFF
--- a/operator/pkg/controller/vault/vault_controller.go
+++ b/operator/pkg/controller/vault/vault_controller.go
@@ -981,7 +981,7 @@ func deploymentForConfigurer(v *vaultv1alpha1.Vault, configmaps corev1.ConfigMap
 			},
 		},
 		Volumes:         withTLSVolume(v, withCredentialsVolume(v, volumes)),
-		SecurityContext: withSecurityContext(v),
+		SecurityContext: withPodSecurityContext(v),
 		NodeSelector:    v.Spec.NodeSelector,
 		Tolerations:     v.Spec.Tolerations,
 	}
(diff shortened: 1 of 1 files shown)


operator/pkg/controller/vault/vault_controller.go
```diff
@@ -1809,7 +1805,18 @@ func withNamespaceEnv(v *vaultv1alpha1.Vault, envs []corev1.EnvVar) []corev1.Env
 	}...)
 }
 
-func withSecurityContext(v *vaultv1alpha1.Vault) *corev1.PodSecurityContext {
+func withContainerSecurityContext(v *vaultv1alpha1.Vault) *corev1.SecurityContext {
+	if disableMlock, ok := v.Spec.Config["disable_mlock"]; ok && disableMlock.(bool) {
```
bonifaido: You can use `cast.ToBool(v.Spec.Config["disable_mlock"])` as seen in `vault_type.go` to make this a bit more readable.

pbalogh-sa: Thanks

bonifaido: 👍

PR #925, opened 2020-03-16
Make client timeout configurable

busser:
| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | yes
| API breaks?     | no|yes
| Deprecations?   | no|yes

### What's in this PR?

This PR adds a new environment variable `VAULT_CLIENT_TIMEOUT` that allows users to configure how long `vault-env` and `vault-secrets-webhook` keep trying to fetch secrets before failing. The current hard-coded behaviour is to wait for 10 seconds, so this is the default value set for this new configuration parameter.

### Why?

We are currently having pods injected with `vault-env` that are failing because the pod's Istio proxy sidecar sometimes takes more than 10 seconds to be ready. When that happens, `vault-env` fails to connect to Vault and the pod fails.

This PR, if merged, would allow us to make `vault-env` wait for longer.

### Additional context

To test this implementation, I built custom `vault-secrets-webhook` and `vault-env` container images and deployed them to a Kubernetes cluster to test the new behaviour. Pods injected with `vault-env` worked as expected according to the `VAULT_CLIENT_TIMEOUT` variable.

### Checklist

### DIFF
--- a/charts/vault-secrets-webhook/values.yaml
+++ b/charts/vault-secrets-webhook/values.yaml
@@ -35,6 +35,7 @@ env:
   # an imagePullSecret
   # DEFAULT_IMAGE_PULL_SECRET:
   # DEFAULT_IMAGE_PULL_SECRET_NAMESPACE:
+  # VAULT_CLIENT_TIMEOUT: 10s
 
 metrics:
   enabled: false
--- a/cmd/vault-secrets-webhook/main.go
+++ b/cmd/vault-secrets-webhook/main.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"time"
 
 	vaultapi "github.com/hashicorp/vault/api"
 	"github.com/prometheus/client_golang/prometheus"
@@ -48,6 +49,7 @@ type VaultConfig struct {
 	Path                        string
 	SkipVerify                  string
 	TLSSecret                   string
+	ClientTimeout               time.Duration
 	UseAgent                    bool
 	VaultEnvDaemon              bool
 	TransitKeyID                string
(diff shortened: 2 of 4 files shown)


primeroz:
Not that i am against this but just wanted to ask something.

it is my understanding that the Vault client does , bu default, 2 retry with exponential backoff before returning an error on every http request ... is that not enough to deal with the issue ?

busser:
> it is my understanding that the Vault client does , bu default, 2 retry with exponential backoff before returning an error on every http request ... is that not enough to deal with the issue ?

Sadly no, because it stops trying all together after 10 seconds and closes the client. We would like for it to keep trying a little bit longer, is all :)

busser:
@bonifaido Any idea when these changes will make into an official release of the bank-vaults container images?

primeroz:
> > it is my understanding that the Vault client does , bu default, 2 retry with exponential backoff before returning an error on every http request ... is that not enough to deal with the issue ?
> 
> Sadly no, because it stops trying all together after 10 seconds and closes the client. At least, that is my understanding of the code. We would like for it to keep trying a little bit longer, is all :)

Cool, that's what i was after.

Then i guess for some reason `our vault client` is not working properly with the backoff and retry

bonifaido:
Yeah our client is the official one from Hashicorp with extra glue added, maybe we should upgrade it?

bonifaido:
The issue is that they haven't released SDK and API since last summer.

bonifaido:
@busser since we have at least two fixes probably this week, but let's wait for some time, possibly some other fixes are in the pipe for this week as well.

busser:
@bonifaido Hello! Any updates on the next release date? 🙂

bonifaido:
Just preparing it right now!

bonifaido:
https://github.com/banzaicloud/bank-vaults/pull/925

Charts and images will be published in half an hour

PR #933, opened 2020-04-03
Implement the ability for a CA to persist when the Server certificate needs to be regenerated

jamie-34254:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no|
| Deprecations?   | no

### What's in this PR?
This PR implements the ability for Bank Vault's operator to use an existing automatically generated Certificate Authority when generating a new Server Certificate in the event of a new one being required.

It seems like etcd's CA was always replaced regardless of one already existing so I haven't implemented support for that as we don't have access to the secret at the point where we would need the CA/key to update the Server cert instead of generating a full chain

### Why?
We were having issues with our cross cluster CA certificate expiring when a LoadBalancer IP was changing and causing the Server Certificate to be renewed

### Additional context
https://community-banzaicloud.slack.com/archives/CFJJW9L94/p1585579074011400

### Checklist

### DIFF
--- a/operator/pkg/controller/vault/vault_controller.go
+++ b/operator/pkg/controller/vault/vault_controller.go
@@ -355,7 +355,7 @@ func (r *ReconcileVault) Reconcile(request reconcile.Request) (reconcile.Result,
 		} else if v.Spec.ExistingTLSSecretName == "" && len(sec.Data) > 0 {
 			// If tls secret exists check expiration date
 			certPEM := string(sec.Data["server.crt"])
-			tlsExpiration, err = getCertExpirationDate(certPEM)
+			tlsExpiration, err = bvtls.GetCertExpirationDate([]byte(certPEM))
 			if err != nil {
 				return reconcile.Result{}, fmt.Errorf("failed to get certificate expiration: %v", err)
 			}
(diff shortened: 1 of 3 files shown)


pkg/sdk/tls/tls.go
```diff
@@ -99,38 +147,147 @@ func (sh *separatedCertHosts) validate() error {
...
+	}
+
+	notAfter = notBefore.Add(validityDuration)
+	return
+}
+
+// NewChain generates ca, server, client and peer TLS certificates.
 // hosts: Comma-separated hostnames and IPs to generate a certificate for
 // validity: Duration that certificate is valid for, in Go Duration format
-func GenerateTLS(hosts string, validity string) (*CertificateChain, error) {
```
bonifaido, line 105: This is a public API (sdk package) function, which we can't just remove, so I would suggest creating a shim for this one.

operator/pkg/controller/vault/vault_controller.go
```diff
@@ -1074,30 +1085,64 @@ func hostsAndIPsForVault(om *vaultv1alpha1.Vault, service *corev1.Service) []str
...
 	if err != nil {
-		return nil, time.Time{}, err
+		return time.Time{}, err
+	}
+
+	// If a secret already exists, load the ca.crt and ca.key from it
+	caCrt := ""
+	caKey := ""
+	if secret == nil {
+		return time.Time{}, errors.New("a nil secret was passed into populateTLSSecret, please instantiate the secret first")
```
jamie-34254, line 1101: One of our devs noticed this - you can't assign to a pointer passed in via a function parameter - since the address is passed in by value - but I don't think we were instantiating secret here anyway, so it's all good AFAIK.

jamie-34254:
Something of note is that the first time `populateTLSSecret` is called it'll regenerate the CA due to the missing ca.key, but it should only be called when the server certificate needs to be regenerated

PR #953, opened 2020-04-15
Add custom functions for awskms and file

### LINKED ISSUE #945: [Operator] Source bootstrap secrets from sources other than environment variables
I'm trying to bootstrap a database dynamic secrets configuration but I don't want to keep the secret in plain text in source control. Instead, I want to be able to store an encrypted string (in my case encrypted with AWS KMS but could be something else) and be able to reference it similarly to how environment variables are sourced [here](https://github.com/banzaicloud/bank-vaults/blob/master/operator/deploy/cr-mysql-ha.yaml#L67).

I can use something like `${ file "/path/to/file" }` to reference a volume mount that has the secret I injected securely, `${ awskms "SomeKMSEncryptedString" }` so I can keep the manifest in source control, including `"SomeKMSEncryptedString"`, or even maybe something like `${ keybase "username:SomeGPGEncryptedString" }`. Maybe add custom template functions in addition to the Sprig ones [here](https://github.com/banzaicloud/bank-vaults/blob/master/internal/configuration/template.go#L62)? Or maybe set additional fields in [this](https://github.com/banzaicloud/bank-vaults/blob/master/internal/configuration/template.go#L46) `env` struct that can be queried from the template using dot notation?

I tried generating the plain-text string in an init container, write it to a volume that is also mounted on the main `vault-configurer` container and read if from there, but the Sprig library doesn't provide access to the file system or to external binaries.

There could be potential workarounds with creating read-write configmaps, or maybe even a webhook that can modify things on the fly but I think those could add significant complexiy.

**Additional context**
Vault 1.3.2
Bank-vaults operator: 0.9.0
Tried on Kubernetes 1.17, but ideally it would also be supported at least on 1.12

patoarvizu:
Signed-off-by: patoarvizu <patoarvizu@gmail.com>

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #945 

### What's in this PR?
Extending the templating functionality by adding custom functions to enable getting strings from other sources like an AWS KMS-encrypted string or a file path.

### Why?
The only function available for templating that comes close to it is `env` which has limited usefulness for the use case.

### Checklist

- [X] Error handling code meets the [guideline](https://github.com/banzaicloud/pipeline/blob/master/docs/error-handling-guide.md)
- [X] Logging code meets the guideline (TODO)

### To Do

- Update documentation
- Add examples
- Add/update tests
- Maybe add other sources (e.g. PGP, Google Cloud KMS, a REST endpoint, another Vault, etc.)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -49,6 +49,7 @@ require (
 	github.com/ugorji/go v1.1.7 // indirect
 	golang.org/x/oauth2 v0.0.0-20190604053449-0f29369cfe45
 	google.golang.org/api v0.7.0
+	google.golang.org/genproto v0.0.0-20190716160619-c506a9f90610
 	k8s.io/api v0.18.0
 	k8s.io/apimachinery v0.18.1-beta.0
 	k8s.io/client-go v0.18.0
--- a/internal/configuration/template.go
+++ b/internal/configuration/template.go
@@ -16,7 +16,9 @@ package configuration
 
 import (
 	"bytes"
+	"context"
 	"encoding/base64"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"strings"
@@ -26,6 +28,9 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/kms"
 	"github.com/goph/emperror"
+
+	cloudkms "cloud.google.com/go/kms/apiv1"
+	kmspb "google.golang.org/genproto/googleapis/cloud/kms/v1"
 )
 
 const DefaultLeftDelimiter = "${"
@@ -90,6 +95,7 @@ func funcMap() map[string]interface{} {
 	return map[string]interface{}{
 		"awskms": awsKmsDecrypt,
 		"file":   fileContent,
+		"gcpkms": gcpKmsDecrypt,
 	}
 }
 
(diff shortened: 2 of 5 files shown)


patoarvizu:
@bonifaido here it is! 🙂

bonifaido:
First of all, it looks good to me @patoarvizu ! Don't worry about the CircleCI build, it became flappy because we are running low on resources in the requested VM, we have to increase it probably (the GitHub build which does the same still works).

patoarvizu:
Thanks @bonifaido!

I added a couple of examples (`operator/deploy/cr-file.yaml` and `operator/deploy/cr-awskms.yaml`), but where would be the best place to add actual documentation of the feature? More generally, I don't think that the fact that the Sprig library of functions are available to invoke is documented anywhere, so it might be good to add that as well (I can do it as part of this same PR if you would like me to).

Also, are there any other functions that you think might be good to add? I admit that I'm selfishly only adding `file` and `awskms` because those are the ones that fit my specific use case right now, but I'd be happy to look into other functions, or at least open issues to track them and implement them separately.

patoarvizu:
Ah, also: can you suggest other people I should add as reviewers?

bonifaido:
Hi @patoarvizu !

I just added @pbalogh-sa as well to the review.

The general documentation resides at https://github.com/banzaicloud/bank-vaults-docs

I think we should create a page with a link to http://masterminds.github.io/sprig/ and extend it with the new functions you added, and mention that these functions are available basically everywhere (in config as well).

I think Google KMS would be nice as well! (I think these two providers are used mostly) and maybe Azure later on.

Many thanks for the great work here!

patoarvizu:
Thanks!

Should I start the PR to bank-vaults-docs in parallel to this, or wait until this is merged before I start working on the documentation?

bonifaido:
Just fixed the acceptance test run (with a rerun)

patoarvizu:
Hey guys, I almost have something ready to also support Google Cloud KMS. I want to hold off on merging this and ping you again to review those additional changes when they're ready. Does that work for you?

patoarvizu:
I just added a `gcpkms` function and an example for it. I'm going to set this as 'ready for review' and let you guys take a look whenever you have time. I'm going to get a head start on the documentation on bank-vaults-docs but I can address any feedback you may have here before we merge. Thanks!

bonifaido:
Looks great, looking forward to the docs!🚀 Thanks again!

PR #959, opened 2020-04-24
add blob template function

bonifaido:

| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | no|yes
| API breaks?     | no|yes
| Deprecations?   | no|yes
| Related tickets | https://github.com/banzaicloud/bank-vaults/pull/953

### What's in this PR?

Support reading from blob storages (also from files) with the `blob` template function. https://gocloud.dev/howto/blob/

### Why?

To allow people reading from cloud storage instead of committing even encrypted text to the source code.

### Checklist


bonifaido:
@patoarvizu I just added a `blob` function as well, which can read from cloud storages and from file as well (gocloud.dev offered this as well) could you please check this one? :)

I'm just testing this on EKS with ServiceAccount based authentication.

operator/deploy/cr-awskms.yaml
```diff
@@ -90,6 +90,8 @@ spec:
             allowed_roles: ['*']
             username: root
             password: '${ awskms (env `ENCRYPTED_DB_CREDS`) }'
+            # You can use the blob function to read the encrypted value from S3 directly
+            # password: '${ awskms (blob `s3://bank-vaults/encrypted/db-creds?region=eu-west-1`) }'
```
patoarvizu, line 94: I think maybe you should add a comment here saying that since `awskms` expects a base64-encoded string, the user should make sure that what's stored on the `blob` target returns a base64 string (you can use Sprig's `b64enc` for that).

Or maybe make it more strict and enforce it by making `blob` *always* return a base64-encoded string? You'll just have to document that requirement.

bonifaido: Thanks you are right I just added the comment.

PR #978, opened 2020-05-13
Operator: Support JSON-formatted Vault policies

### LINKED ISSUE #979: Operator: Add support for JSON-formatted Vault policies
Vault supports policies in both HCL and JSON formats, but the operator currently only accepts HCL-formatted policies.

For context, our policies were becoming boilerplate-y and difficult to manage, so we have started templating them with Jsonnet, where generating JSON is far easier than HCL.

When we try providing policies as escaped JSON strings, we get the following error:

```
ERRO[0000] error configuring vault: error configuring policies for vault: error formatting example-policy policy rules: At 1:1: expected: IDENT | STRING got: LBRACE
```

This is because it's trying to use an HCL formatter to make it pretty, which won't work for JSON!

I've raised a pull request that attempts to parse the policy (with a Vault function which parses both JSON and HCL) and if it succeeds, applies it: https://github.com/banzaicloud/bank-vaults/pull/978

Thanks!

chrlwrd:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #979

### What's in this PR?
This adds support for Vault Policies to be provided as JSON strings, in addition to HCL-formatted strings. Current behaviour tries to use an HCL formatter to make it pretty, which won't work for JSON!

This change makes the operator attempt to parse the policy (with a Vault function which parses both JSON and HCL) and if it succeeds, applies it.

### Why?
Vault supports policies in both HCL and JSON format so the operator should too!

### Additional context
Our policies were becoming boilerplate-y and less maintainable so we decided to generate them via Jsonnet, where it's significantly easier to template JSON than HCL.

This has been tested with valid JSON and HCL policies and both are created successfully.

### Checklist

### DIFF
--- a/pkg/sdk/vault/operator_client.go
+++ b/pkg/sdk/vault/operator_client.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hashicorp/hcl"
 	hclPrinter "github.com/hashicorp/hcl/hcl/printer"
 	"github.com/hashicorp/vault/api"
 	"github.com/hashicorp/vault/sdk/helper/consts"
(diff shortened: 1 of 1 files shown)


primeroz:
Good Job @chrlwrd , i won't approve since i work with him but i am glad to bring him on board on this project!

bonifaido:
Ah geat, nice job! Welcome!

PR #999, opened 2020-05-27
fix(vault-secrets-webhook): typo in helm variable

nesl247:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
Fix the casing of `configMapPolicy` in the helm chart.

### Why?
Because it doesn't follow convention. Other variables in the file also use `configMap` instead.

### Checklist


pbalogh-sa:
Hi @nesl247 .  If you only change it in the `values.yaml`, the chart will be broken. You have to change the line as well, which refers to the value.
https://github.com/banzaicloud/bank-vaults/blob/d6c4b9bdff21c919973c68ef4fc54d9c5289651f/charts/vault-secrets-webhook/templates/apiservice-webhook.yaml#L135

bonifaido:
Also please use the Helm [`default`](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function) function to retain backward compatibility, and also please bump the chart version.

nesl247:
@bonifaido Sorry about that, I was doing too much at once yesterday. It should all be fixed up now.

Issue #1006, opened 2020-06-08
Inconsistent requirements for different resource types to enable processing


wnagele:
**Describe the bug**:
Currently secrets and unstructured objects are only processed if they are annotated with `vault.security.banzaicloud.io/vault-addr` (see [here](https://github.com/banzaicloud/bank-vaults/blob/1.3.0/cmd/vault-secrets-webhook/main.go#L319)).

For config maps the same is true for the annotation `vault.security.banzaicloud.io/mutate-configmap` (see [here](https://github.com/banzaicloud/bank-vaults/blob/1.3.0/cmd/vault-secrets-webhook/main.go#L324)).

This creates tons of duplication of information in the cluster and is especially problematic if these have already been provided in the global configuration.

**Expected behaviour**:
`Pod`, `Secret`, `ConfigMap` and unstructured types should all be considered equally as to when they receive processing. This could either be done through a on/off annotation that each resource has to receive or simply process them based on all relevant configuration being available (currently happens for Pods already).

**Steps to reproduce the bug**:
1. Create a global configuration where `vault-addr` and `vault-path` are set
2. Create a secret with a value requiring a vault lookup (starting with `vault:`) and annotate it with the relevant role (`vault.security.banzaicloud.io/vault-role`)
3. Check the secret - it will not have been processed
4. Modify the secret to include the `vault.security.banzaicloud.io/vault-addr` annotation with the same information already present in the global configuration
5. Check the secret - it has been processed

**Additional context**:
Instead of checking the annotation directly it could be checked using [vaultConfig](https://github.com/banzaicloud/bank-vaults/blob/1.3.0/cmd/vault-secrets-webhook/main.go#L314) to allow processing with the global configuration taking effect.

/kind bug
/kind enhancement

PR #1014, opened 2020-06-09
Remove explicit check for annotations

### LINKED ISSUE #938: [vault-secrets-webhook] Kube Secret injection should use the default Vault address when the vault-addr annotation is empty or not present
> Bugs should be filed for issues encountered whilst operating bank-vaults.
> You should first attempt to resolve your issues through the community support
> channels, e.g. Slack, in order to rule out individual configuration errors.
> Please provide as much detail as possible.

**Describe the bug**:

When deploying the webhook it's possible to alter the assumed defaults by providing environment variables to the Docker image "banzaicloud/bank-vaults".

E.g.: [Here](https://github.com/banzaicloud/bank-vaults/blob/d4b68c89ca3804483d7f842f5f725bc1b7d4e033/cmd/vault-secrets-webhook/main.go#L87) we can alter the default Vault address by using the env var VAULT_ADDR

When doing so, when doing pod injection we don't need to specify the vault-addr annotation every time because the webhook default will be assumed and the pod is always scanned (see [here](https://github.com/banzaicloud/bank-vaults/blob/d4b68c89ca3804483d7f842f5f725bc1b7d4e033/cmd/vault-secrets-webhook/main.go#L318))

However, when mapping Vault secrets to Kube Secrets we are forced to have the vault-addr annotation in the Kube Secret, or otherwise the Secret isn't scanned (see [here](https://github.com/banzaicloud/bank-vaults/blob/d4b68c89ca3804483d7f842f5f725bc1b7d4e033/cmd/vault-secrets-webhook/main.go#L320))

And you need to specify the true Vault address to make it work because, given that the vault-addr annotation is present, the default from env var VAULT_ADDR won't be applied. I tried to set vault-addr to  "" (empty) but the Vault API URL to be called in order to fetch the secret becomes invalid because it's not checked for "":

`Failed to request new Vault token Put "/v1/auth/kubernetes/login": unsupported protocol scheme ""` 

**Expected behaviour**:

Given that the vault-addr annotation is optional (as there is a default value, specified [here](https://banzaicloud.com/docs/bank-vaults/mutating-webhook/#podspec-secret-configmap-and-crd-annotations)), using it in Kube Secrets for injection shouldn't be mandatory.

Either:

1. (second best) The Kube Secrets should be always scanned, as it happens with the Pods (see [here](https://github.com/banzaicloud/bank-vaults/blob/d4b68c89ca3804483d7f842f5f725bc1b7d4e033/cmd/vault-secrets-webhook/main.go#L318)), i.e. no need to have the vault-addr annotation present.
1. (preferred) The Kube Secrets to be scanned should be marked with a dedicated annotation, similar to the "mutate-configmap" one for ConfigMap injection, therefore allowing the omission of vault-addr
1. (ugly) The vault-addr annotation in Kube Secrets should allow a value of "" (empty), which the webhook should replace by the default Vault address (env var VAULT_ADDR)

1. Follow the [standard installation tutorial](https://banzaicloud.com/blog/inject-secrets-into-pods-vault-revisited), but override the default Vault address with Helm: `helm ... --set env.VAULT_ADDR="${VAULT_ADDR}"`
1. Apply [the Secret injection example](https://banzaicloud.com/blog/inject-secrets-into-pods-vault-revisited/#getting-secrets-data-from-vault-and-transplanting-it-into-a-kubernetes-secret), but removing the vault-addr annotation
-- Consequence: The Secret is not detected and mutated by the webhook

**Additional context**:
I am using the webhook with a Vault service outside of Kubernetes, and I want to provide to the devs deploying apps in Kube with Vault secret injection in a transparent manner, so I want only the webhook to know the Vault address.

**Environment details**:

- Kubernetes version (e.g. v1.15.4):
- Cloud-provider/provisioner: On-prem K8s
- bank-vaults version: 1.0.1
- Install method (e.g. helm or static manifests): Helm

**Logs**

Even in debug mode, the webhook it's not reporting that the Secret is being skipped because of not having the annotation, maybe some "trace" level logs would be appropriate :
`
{"app":"vault-secrets-webhook","level":"debug","msg":"reviewing request xxxxx, named: my-namespace/","time":"2020-04-07T16:28:08Z"}

{"app":"vault-secrets-webhook","level":"debug","msg":"json patch for request xxxxx: []","time":"2020-04-07T16:28:08Z"}
`

/kind bug

flozzone:
Check for the addr annotation is duplicate here, because its already
covered by parseVaultConfig method.

Fixes #938

Signed-off-by: flozzone <flozzone@gmail.com>

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #938

### What's in this PR?
Remove checks for the address annotation, because it is already done by the parseVaultConfig method.

### Why?
This allows Secrets and Unstructured configurations to be mutated even there is no explicit `vault.security.banzaicloud.io/vault-addr` annotation. It will use the default configuration value instead.

### Additional context
Testings with this fix showed that Secrets get mutated as expected. See docker.hub repository
https://hub.docker.com/repository/docker/florinzz/vault-secret-webhook with the `PR-remove-annotation-check`. I assume this will applies also to Unstructured configurations.

### Checklist

### DIFF
--- a/cmd/vault-secrets-webhook/main.go
+++ b/cmd/vault-secrets-webhook/main.go
@@ -315,11 +315,10 @@ func (mw *mutatingWebhook) vaultSecretsMutator(ctx context.Context, obj metav1.O
 	switch v := obj.(type) {
 	case *corev1.Pod:
 		return false, mw.mutatePod(v, vaultConfig, whcontext.GetAdmissionRequest(ctx).Namespace, whcontext.IsAdmissionRequestDryRun(ctx))
+
 	case *corev1.Secret:
-		if _, ok := obj.GetAnnotations()["vault.security.banzaicloud.io/vault-addr"]; ok {
-			return false, mw.mutateSecret(v, vaultConfig)
-		}
-		return false, nil
+		return false, mw.mutateSecret(v, vaultConfig)
+
 	case *corev1.ConfigMap:
 		if _, ok := obj.GetAnnotations()["vault.security.banzaicloud.io/mutate-configmap"]; ok {
 			return false, mw.mutateConfigMap(v, vaultConfig)
@@ -327,10 +326,8 @@ func (mw *mutatingWebhook) vaultSecretsMutator(ctx context.Context, obj metav1.O
 		return false, nil
 
 	case *unstructured.Unstructured:
-		if _, ok := obj.GetAnnotations()["vault.security.banzaicloud.io/vault-addr"]; ok {
-			return false, mw.mutateObject(v, vaultConfig)
-		}
-		return false, nil
+		return false, mw.mutateObject(v, vaultConfig)
+
 	default:
 		return false, nil
 	}


bonifaido:
Thanks @flozzone, probably #1006 is also going to be easier a bit with this one!

PR #1034, opened 2020-06-25
fix helm test for https

jcooley:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | none

### What's in this PR?
Fix error an in helm test

### Why?
When vault is using https the "helm test" fails with "Client sent an HTTP request to an HTTPS server."

### Additional context
Added VAULT_SKIP_VERIFY=true as we are not testing if the cert is valid (e.g. self-signed).

### Checklist

### To Do

### DIFF
--- a/charts/vault/Chart.yaml
+++ b/charts/vault/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart for Vault, a tool for managing secrets
 name: vault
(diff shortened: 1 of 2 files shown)


bonifaido:
LGTM! Thanks!

PR #1054, opened 2020-07-21
Support k8s downward api in vault chart environment variables

jcooley:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature? |   yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | None

### What's in this PR?
The vault chart expects environmental variables in the form of name/value pairs and will
remove variables using the [k8s downward api](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#the-downward-api).

### Why?
At the moment you cannot use environment variable like 
```      
env:
  - name: MY_NODE_NAME
    valueFrom:
      fieldRef:
        fieldPath: spec.nodeName
```
as they are not in the form of simple name/value pairs.
### Additional context
This is useful so that vault can access the downward api. It could be used to set the api_addr and cluster_addr and we would use also use it to set the node_ip for Datadog integration. 

### Checklist

- [X] User guide and development docs updated (if needed)

### To Do
Completed.

### DIFF
--- a/charts/vault/Chart.yaml
+++ b/charts/vault/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: vault
-version: 1.3.6
+version: 1.3.7
 appVersion: 1.3.3
 description: A Helm chart for Vault, a tool for managing secrets
 home: https://www.vaultproject.io/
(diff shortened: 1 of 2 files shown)


jcooley:
Thanks @bonifaido, credit to @KealanM and @Bryji for figuring it out.

bonifaido:
Nice! Thanks to all!

PR #1055, opened 2020-07-21
Aws sse

### LINKED ISSUE #1053: Support Server Side Encryption (SSE) for AWS S3 buckets
We have a blanket policy of enforcing SSE for every bucket that we use in our infrastructure. This is a problem when working with the Vault Operator, because the policy will prevent anything to be written in the bucket unless it specifically requests that the objects are encrypted at rest. 

Add the ability to the Operator to request an object to be encrypted on the server side. There are two algorithms that can be used: `AES256` or `aws:kms`.
- In the case of `AES256` the user doesn't need to do anything, the S3 API will automatically use a key to encrypt the data at rest, transparently to the application. 
- In the case of `aws:kms`, the user must provide the ID of a Customer Managed Key (CMK) that will be used to encrypt the data at rest by the S3 API.
A potential way of implementing this would be to add a new flag called `s3SSEAlgo` which is of type string and has either the value `AES256` or `aws:kms`. If the algorithm is `AWS256`, then no `aws-kms-key-id` is needed and if one is provided, that will result in an error. If the algorithm is `aws:kms`, then the user must provide an `aws-kms-key-id`. According to the AWS docs, the region of CMK must be the same as the region of the bucket. If they are not, the Operator must throw an error.
This solution is backwards compatible, and should not need a lot of extra code to be implemented.

Can't think of any other alternative. Open to new ideas though.

**Additional context**
That's it.

ptzianos:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1053

### What's in this PR?
Adds a new flag to allow users to configure the AWS SSE algorithm, fixes the parameter checks to allow the new scheme and adds an example of how to use SSE in AWS

### Why?
Without this PR, a user is unable to use AWS S3 buckets with server-side encryption policies. A lot of companies enforce these policies across all their S3 buckets to ensure that the data are encrypted at rest. The operator performs client-side encryption at the moment, without any way of changing this.

### Additional context
This PR has been tested with the following cases:
1. S3 bucket with policy that enforces AES256 SSE
2. S3 bucket with policy that enforces aws:kms SSE with a CMK
3. S3 bucket with no SSE encryption enforcing

### Checklist


bonifaido:
@ptzianos please rebase the commits on with the same email address as your GitHub user otherwise the CLA check won't pass! Thanks in advance! Once ready I will start to review the changes! :)

ptzianos:
@bonifaido all the commits have been rebased and CLAs have been singed. Ready for reviewing 💪

ptzianos:
@bonifaido thank you very much for your kind words. when do you think you will get a new release out?

bonifaido:
Next week most probably, until then you can use the `master` tags for testing and stuff like that!

PR #1088, opened 2020-09-01
Allow the vault-env repository to be overridden.

### LINKED ISSUE #1075: Be able to specify repository and tag for VAULT_ENV_IMAGE in Helm chart for webhook
Right now the environment variable is hardcoded to

```
            - name: VAULT_ENV_IMAGE
              value: "banzaicloud/vault-env:{{ include "vault-secrets-webhook.bank-vaults.version" . }}"
```

in webhook-deployment.yaml - it would nice that it could be pulled from another repository. It used to work specifying VAULT_ENV_IMAGE under env: in the values file. If I do it now Helm generates two entries with different values, and in my test it seems to prefer the first in the list (the hardcoded one).

Ulrik

brewneaux:
Fixes issue #1075 to allow for the vault-env repository to be overridden via the values file.

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | yes
| Deprecations?   | yes
| Related tickets | Fixes #1075 

### What's in this PR?
Allows the repository for the vault-env container, injected into deployments etc, to be overridden.  Primarily necessary in environments where dockerhub is not available due to firewalls or access rules.

### Why?
My employer has secure environments where we need to use this chart, and make use of an Artifactory installation to serve our images.  We can't directly pull from docker.io, and upload images directly to our instance once they have been reviewed and approved.

### Checklist

### DIFF
--- a/charts/vault-secrets-webhook/values.yaml
+++ b/charts/vault-secrets-webhook/values.yaml
@@ -34,9 +34,11 @@ serviceAccount:
   # Enables GKE workload identity
   #  iam.gke.io/gcp-service-account: gsa@project.iam.gserviceaccount.com
 
+vault_env:
+  repository: banzaicloud/vault-env
+
 env:
   VAULT_IMAGE: vault:1.5.0
-  # VAULT_ENV_IMAGE: banzaicloud/vault-env:1.4.0
   # VAULT_CAPATH: /vault/tls
   # used when the pod that should get secret injected does not specify
   # an imagePullSecret
--- a/charts/vault-secrets-webhook/values.yaml
+++ b/charts/vault-secrets-webhook/values.yaml
@@ -34,7 +34,7 @@ serviceAccount:
   # Enables GKE workload identity
   #  iam.gke.io/gcp-service-account: gsa@project.iam.gserviceaccount.com
 
-vault_env:
+vaultEnv:
   repository: banzaicloud/vault-env
 
 env:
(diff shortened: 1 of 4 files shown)


bonifaido:
Hi @brewneaux, thanks for the PR, please bump the chart version as well (CI broke because of this).

brewneaux:
Hi @bonifaido - thanks for the quick reply. I also can't figure out exactly what I'm missing in the PR template that is causing that check to fail

bonifaido:
No problem, just fixed the description (Bug fix was broken).

charts/vault-secrets-webhook/values.yaml
```diff
@@ -34,9 +34,11 @@ serviceAccount:
   # Enables GKE workload identity
   #  iam.gke.io/gcp-service-account: gsa@project.iam.gserviceaccount.com
 
+vault_env:
```
bonifaido, line 37: One minor thing @brewneaux, can we make this camelCase, as the other values in this file?

brewneaux: Pushed

bonifaido (approved):
LGTM, thanks!

charts/vault-secrets-webhook/values.yaml
bonifaido: Thanks

Issue #1098, opened 2020-09-04
[vault-secrets-webhook] Add all keys in vault path as env variables


rocpatel:
**Is your feature request related to a problem? Please describe.**

We have laid out our vault paths by applications.  So, each application gets its own path, thus the keys in the path are all the env variables we want to send to the app.  Right now, we have to add those same keys as env variables in our k8s manifests.  This can be somewhat error-prone, and extra work for the manifest developers.

**Describe the solution you'd like**

It would be nice to have a way of just specifying a vault path, and have vault-env add all the keys in that path as env variables to the process.

**Describe alternatives you've considered**

Couldn't really find any alternatives.  The only one is to explicitly have all the keys as env variables.

**Additional context**

I have already started working on a solution that adds an additional annotation with a comma-delimited list of vault paths to read from.  When this annotation is present, vault-env would read those paths and add all the keys as env variables.

bonifaido:
@rocpatel I think we can close this now, right?

bonifaido:
fixed in #1100

PR #1100, opened 2020-09-06
[webhook] Add env from vault path

rocpatel:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | mentioned in #1098 

### What's in this PR?
This PR adds an annotation `vault-env-from-path`, which will mutate a pod with an additional environment variable `VAULT_ENV_FROM_PATH`.  This annotation will hold a comma-delimited list of the vault paths to read from.  When this environment variable is present the `vault-env` binary will call the `InjectSecretsFromVaultPath` method on the `injector` package.  The `InjectSecretsFromVaultPath` method will inject all the secrets from the vault paths into the environment.  It does not sanitize the keys but adds them as is to the environment.  Also, if a later path has a duplicate key to a previous path, that key will be overwritten by the later path value.  This PR probably can be refined a bit more, but wanted to put something together to start a discussion.

### Additional context
I have added some unit tests to test the functionality.  The only thing I don't have is an error if the path doesn't exist.  I also do not take into account the transit encryption for the paths.

### Checklist

### To Do

* mutation of configmaps/secrets
* refactoring of code to make it DRY
* possibly uppercasing keys when adding to environment variables
* updating of documentation, need some direction as to where that is
* error handling around missing paths to make it easier to debug missing path issues


bonifaido:
Thanks, I will check this one today! 👍

internal/injector/inejctor.go
```diff
@@ -215,3 +215,81 @@ func (i SecretInjector) InjectSecretsFromVault(references map[string]string, inj
 
 	return nil
 }
+
+func (i SecretInjector) InjectSecretsFromVaultPath(paths string, inject SecretInjectorFunc) error {
```
bonifaido, line 219: Basically this function is the same as the original `InjectSecretsFromVault`, just wrapped inside another `for` loop, am I right? We should reuse the same logic extracted from there I think.

bonifaido: Really nice tests though! 👍

rocpatel: Yes, as I was putting it together, noticed that it's doing the same thing. I'll look into refactoring it a bit

rocpatel: Ok, so i went ahead and refactored it a bit.  I didn't find it as simple as just re-using InjectSecretsFromVault with extra parameters.  However, I did pull out some common logic to a new method.  Let me know what you think

bonifaido (dismissed):
LGTM, my only question is regarding this comment:
> The only thing I don't have is an error if the path doesn't exist.

Do we still miss this?

pbalogh-sa (approved):
Thanks @rocpatel !

rocpatel:
> LGTM, my only question is regarding this comment:
> 
> > The only thing I don't have is an error if the path doesn't exist.
> 
> Do we still miss this?

It was actually doing a hard fail when the path is missing.  Just pushed up something that uses the same logic to check `IgnoreMissingSecrets`.  Also, is there a spot to update the documentation?  Would need to add the extra annotation.

bonifaido (approved):
LGTM, thank you very much @rocpatel 

The documentation is available at: https://github.com/banzaicloud/bank-vaults-docs

PR #1117, opened 2020-09-22
fix: move VAULT_ENV_FROM_PATH to mutateContainers

rocpatel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
When adding the env-from-path annotation, I was incorrectly mutating the pod instead of the container.  This fix is to actually mutate the container.  The work around for this is to explicitly add `VAULT_ENV_FROM_PATH` environment variable to the container.

### Checklist


bonifaido (approved):
Thank you for the fix!

PR #1119, opened 2020-09-24
vault-mutating-webhook: allow specifying pod securityContext and full container securityContext

dbeal-wiser:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | none

### What's in this PR?

It is not possible to specify full gamut of `pod.spec.container[].securityContext` or `pod.spec.securityContext` without modifying the chart itself. In our case, we needed to set `pod.spec.securityContext.fsGroup` to the same group as the user in order to be compatible with AWS [OIDC EKS IAM](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html). This PR includes a minor fix to allow specifying all `securityContext` options on the webhook deployment. There should be no effective change unless custom options are set.

### Additional context

Users who may have set custom options and not realized they were unset may be affected. We tested the implementation on our in-house setup and verified options can be set appropriately.

### Checklist

- [x] (not needed) User guide and development docs updated (if needed)

### DIFF
--- a/charts/vault-secrets-webhook/templates/webhook-deployment.yaml
+++ b/charts/vault-secrets-webhook/templates/webhook-deployment.yaml
@@ -98,3 +98,7 @@ spec:
       affinity:
 {{ toYaml .Values.affinity | indent 8 }}
   {{- end }}
+  {{- if .Values.podSecurityContext }}
+      securityContext:
+{{ toYaml .Values.podSecurityContext | indent 8 }}
+  {{- end }}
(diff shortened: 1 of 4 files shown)


charts/vault-secrets-webhook/templates/webhook-deployment.yaml
```diff
@@ -102,3 +98,7 @@ spec:
       affinity:
 {{ toYaml .Values.affinity | indent 8 }}
   {{- end }}
+  {{- if .Values.podSecurityContext }}
```
bonifaido, line 101: `podSecurityContext` should be mentioned in the values.yaml file with a default empty value. Also in the README.

bonifaido (changes requested):
Please bump the chart version as well.

dbeal-wiser:
btw im not sure about if I should update the spacing for the README.md table, let me know if that should be adjusted or if the default value is just so long that its not worth it.

bonifaido (approved):
It is absolutely fine @dbeal-wiser, thanks!

Issue #966, opened 2020-05-06
[vault-secrets-webhook] Webhook should support major clouds container registries (ACR, etc...)


amit-o:
Right now, the webhook only supports ECR's authentication out of the box.
It should probably support GCR and ACR, as well.

leominov:
Look's like a part of my closed pull request #654 :D

bonifaido:
Still happy to merge it, if it up-to-date! :)

bonifaido:
GCR got implemented already.

sebader:
are there any news around this for ACR (Azure Container Registry)? We are using an AKS cluster and have "AcrPull" permissions assigned to the kubelet identity which works fine for image pulling but it does not look like the webhook takes that into account

bonifaido:
Hi @sebader we have migrated the registry code to [go-containerregistry](https://github.com/vdemeester/k8s-pkg-credentialprovider/tree/v1.21.0/azure) already, which should support ACR as well. If not that is a bug, please provide some logs from the webhook and some environment issues (versions used) on a new Github issue.

bonifaido:
Originally fixed by: https://github.com/banzaicloud/bank-vaults/pull/1236

sebader:
thanks @bonifaido! Looking [at the code](https://github.com/vdemeester/k8s-pkg-credentialprovider/blob/74c721a3e6ab7a5eea42230cdb92412b8df78537/azure/azure_credentials.go#L254), there seems to be some reference to MSI (which is what we are looking for). I just don't see any docs on how to configure this. Do you have any pointers to a sample config/docs? thanks!

bonifaido:
Unfortunately no, I don't have, since this library is not really well documented. From go-containerregistry there is no way to programatically configure it that is for sure.

Can you enable some logging in the webhook? This file [https://github.com/vdemeester/k8s-pkg-credentialprovider/blob/master/azure/azure_credentials.go#L254] seems to have some useful log messages.

bonifaido:
A snippet from the docs of go-containerregistry:
>>> k8schain

>>> This is an implementation of the github.com/google/go-containerregistry library's authn.Keychain interface based on the authentication semantics used by the Kubelet when performing the pull of a Pod's images.

bonifaido:
I think this is the same issue: https://github.com/banzaicloud/bank-vaults/issues/927#issuecomment-878857469

PR #1120, opened 2020-09-25
Add support for GCR default credentials

viktorradnai:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | partially #966

### What's in this PR?

This PR adds support for authenticating against Google Container Registry without explicit username / password. 

It works similarly to the already present Amazon ECR support: If there are no username / password / imagePullSecrets specified, then it checks the registry hostname against the known GCR hotsts. If the pattern matches then it tries to obtain an auth token from the metadata service endpoint when running inside Google Cloud (which also works with workload identity if running in GKE).

### Why?

Using default credentials avoids the need for pre-shared credentials (passwords or manually generated service account tokens) which would need to be managed.

### Checklist

### To Do

This PR does not address authenticating with an user supplied Google service account token (although that may be possible using the username/password mechanism but was not tested).

### DIFF
--- a/cmd/vault-secrets-webhook/registry/registry.go
+++ b/cmd/vault-secrets-webhook/registry/registry.go
@@ -57,7 +57,7 @@ func init() {
 	ecrHostPattern = regexp.MustCompile(`([a-zA-Z0-9][a-zA-Z0-9-_]*)\.dkr\.ecr(\-fips)?\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?`)
 
 	// From https://cloud.google.com/container-registry/docs/overview
-	gcrHostPattern = regexp.MustCompile(`((us|eu|asia)\.)?gcr.io`)
+	gcrHostPattern = regexp.MustCompile(`^https://((us|eu|asia)\.)?gcr\.io$`)
 }
 
 // ImageRegistry is a docker registry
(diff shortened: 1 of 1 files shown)


cmd/vault-secrets-webhook/registry/registry.go
```diff
@@ -52,6 +55,9 @@ func init() {
 
 	// Adapted from https://github.com/awslabs/amazon-ecr-credential-helper/blob/master/ecr-login/api/client.go#L34
 	ecrHostPattern = regexp.MustCompile(`([a-zA-Z0-9][a-zA-Z0-9-_]*)\.dkr\.ecr(\-fips)?\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?`)
+
+	// From https://cloud.google.com/container-registry/docs/overview
+	gcrHostPattern = regexp.MustCompile(`((us|eu|asia)\.)?gcr.io`)
```
bonifaido, line 60: Code scanning just gave some hints here about the unescaped dot, which is highly unlikely to cause any issues, but we probably should escape it ;)

bonifaido (dismissed):
LGTM, other than the small code scanning hint. Thank you @viktorradnai !

bonifaido:
Once this gets merged, we should add it to the docs, that this works automatically for GCR as well here: https://github.com/banzaicloud/bank-vaults-docs/blob/12b992e5262acad355767a72371d3449a408960a/docs/mutating-webhook/_index.md#registry-access

bonifaido (approved):
👍

PR #1140, opened 2020-10-07
Fix linting in Helm 3.3

cablespaghetti:
Signed-off-by: Sam Weston <weston.sam@gmail.com>

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | yes
| Deprecations?   | no

### What's in this PR?
We run all our deployments through Helm's Linting and in 3.3 there are some additional rules which the charts in this repo fail. I don't think any of these changes are particularly controversial other than the change in the vault-operator chart to move away from the apiextensions.k8s.io/v1beta1 deprecated in Kubernetes 1.16, which does mean that the chart won't work on older Kubernetes versions. However I would hope most people using this chart would be on 1.16 or newer anyway.

### Why?
Pass Helm linting for best practises.

### Additional context
- vault-operator chart `type: application` removal is because this is only valid for v2 charts
- vault-operator chart `apiextensions.k8s.io/v1beta1` to `v1` change is because this API is deprecated since v1.16 of Kubernetes
- icon addition was just a warning but thought it made sense to add
- vault chart RoleBinding change away from a list is due to this https://github.com/helm/helm/issues/8615. We could potentially just add a name to the List instead.

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -216,14 +216,14 @@ workflows:
 
   helm-chart:
     jobs:
-      - helm2/lint-chart:
+      - helm/lint-chart:
           executor: helm/helm2
           charts-dir: charts
           filters:
             tags:
               ignore: /.*/
 
-      - helm3/lint-chart:
+      - helm/lint-chart:
           executor: helm/helm3
           charts-dir: charts
           filters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -216,8 +216,15 @@ workflows:
 
   helm-chart:
     jobs:
-      - helm/lint-chart:
+      - helm2/lint-chart:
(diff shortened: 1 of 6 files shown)


bonifaido:
I'm just in the middle of removing Helm2 support in #1139 , so the type will be changed to Helm2, can we postpone this merge until that?

cablespaghetti:
Hi @bonifaido, sure this isn't massively urgent. :+1: I'll get this to pass and then rebase it until that PR is merged.

bonifaido:
I just merged the Helm 3 fixes for the operator @cablespaghetti

cablespaghetti:
Looks good to me. Everything passes linting in 1.6.0 now. :)

PR #1143, opened 2020-10-14
Possibility to define multiple inline-secrets for mutation

### LINKED ISSUE #1001: Support for multiple vault secrets in the same generic K8s Secret key
Using the vault-secrets mutating webhook, I'm unable to replace multiple secrets inside the same key, or replace secrets inside a larger string of text.

This would be useful for projects that expect config files embedded inside a single, such as AlertManager and ArgoCD. I would like to keep my secret values in Vault simple and segregated, so that I do not have to update the entire contents of my secret in Vault each time a single value changes.

I would like the following example to function as expected, only the vault value is replaced. Multiple vault values are replaced successfully.

```
---
apiVersion: v1
kind: Secret
metadata:
  name: argocd-notifications-secret
  namespace: argo-cd
  annotations:
    vault.security.banzaicloud.io/vault-addr: <omitted>
    vault.security.banzaicloud.io/vault-path: <omitted>
    vault.security.banzaicloud.io/vault-role: vault-secrets-webhook_vault-secrets-webhook
stringData:
  # this works
  # key: vault:secret/data/project/argocd/prod/notifications#SLACK_TOKEN#1

  # this does not
  notifiers.yaml: |
    slack:
      token: vault:secret/data/project/argocd/prod/notifications#SLACK_TOKEN#1
```

I believe this is the code that would need to be modified, and that docker registry credentials have some special handling that appear to be similar to what we would want for generic Secrets.
https://github.com/banzaicloud/bank-vaults/blob/master/cmd/vault-secrets-webhook/secret.go#L31-L76

I can work around this today by placing the entire multi line text block into Vault as a single value.

kschu91:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1001 

### What's in this PR?
https://github.com/banzaicloud/bank-vaults/issues/1001#issuecomment-707641069

### Why?
https://github.com/banzaicloud/bank-vaults/issues/1001#issuecomment-707641069

### Checklist

### To Do

- [x] Please review implementation

### DIFF
--- a/cmd/vault-secrets-webhook/common.go
+++ b/cmd/vault-secrets-webhook/common.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"regexp"
 	"strings"
 
 	"github.com/sirupsen/logrus"
@@ -22,6 +23,8 @@ import (
 	"github.com/banzaicloud/bank-vaults/internal/injector"
 	"github.com/banzaicloud/bank-vaults/pkg/sdk/vault"
 )
+
+var InlineMutationRegex = regexp.MustCompile("\\${{(.*?)}}")
 
 func getDataFromVault(data map[string]string, vaultClient *vault.Client, vaultConfig VaultConfig, logger logrus.FieldLogger) (map[string]string, error) {
 	vaultData := make(map[string]string, len(data))
@@ -44,5 +47,9 @@ func hasVaultPrefix(value string) bool {
 }
 
 func hasInlineVaultDelimiters(value string) bool {
-	return strings.Contains(value, "${{") && strings.Contains(value, "}}")
+	return len(findInlineVaultDelimiters(value)) > 0
+}
+
+func findInlineVaultDelimiters(value string) [][]string {
+	return InlineMutationRegex.FindAllStringSubmatch(value, -1)
 }
--- a/cmd/vault-secrets-webhook/common.go
+++ b/cmd/vault-secrets-webhook/common.go
@@ -24,7 +24,7 @@ import (
 	"github.com/banzaicloud/bank-vaults/pkg/sdk/vault"
 )
 
-var InlineMutationRegex = regexp.MustCompile("\\${([>]{0,2}vault:.*?)}")
+var InlineMutationRegex = regexp.MustCompile(`\${([>]{0,2}vault:.*?)}`)
 
 func getDataFromVault(data map[string]string, vaultClient *vault.Client, vaultConfig VaultConfig, logger logrus.FieldLogger) (map[string]string, error) {
 	vaultData := make(map[string]string, len(data))
--- a/cmd/vault-secrets-webhook/common.go
+++ b/cmd/vault-secrets-webhook/common.go
@@ -24,7 +24,7 @@ import (
 	"github.com/banzaicloud/bank-vaults/pkg/sdk/vault"
 )
 
-var InlineMutationRegex = regexp.MustCompile("\\${{(.*?)}}")
+var InlineMutationRegex = regexp.MustCompile("\\${([>]{0,2}vault:.*?)}")
 
 func getDataFromVault(data map[string]string, vaultClient *vault.Client, vaultConfig VaultConfig, logger logrus.FieldLogger) (map[string]string, error) {
 	vaultData := make(map[string]string, len(data))
--- a/cmd/vault-secrets-webhook/common.go
+++ b/cmd/vault-secrets-webhook/common.go
@@ -42,3 +42,7 @@ func getDataFromVault(data map[string]string, vaultClient *vault.Client, vaultCo
 func hasVaultPrefix(value string) bool {
 	return strings.HasPrefix(value, "vault:") || strings.HasPrefix(value, ">>vault:")
 }
+
+func hasInlineVaultDelimiters(value string) bool {
+	return strings.Contains(value, "${{") && strings.Contains(value, "}}")
+}
--- a/cmd/vault-secrets-webhook/configmap.go
+++ b/cmd/vault-secrets-webhook/configmap.go
@@ -16,7 +16,6 @@ package main
 
 import (
 	"encoding/base64"
-	"regexp"
 	"strings"
 
 	"emperror.dev/errors"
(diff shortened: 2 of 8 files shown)


bonifaido:
Hi @kschu91, regarding the failed build here: https://github.com/banzaicloud/bank-vaults/pull/1143/checks?check_run_id=1252775994

Could you please add the required ignore line to the license check config? https://github.com/banzaicloud/bank-vaults/blob/master/.licensei.toml#L27

bonifaido:
Don't worry about my previous comment, just added it in https://github.com/banzaicloud/bank-vaults/pull/1144/commits/e2f8a3566803dda479b6ced57c1d218d89457938

Please rebase on master once this gets merged (a few minutes from now on).

cmd/vault-secrets-webhook/common.go
```diff
@@ -42,3 +42,7 @@ func getDataFromVault(data map[string]string, vaultClient *vault.Client, vaultCo
 func hasVaultPrefix(value string) bool {
 	return strings.HasPrefix(value, "vault:") || strings.HasPrefix(value, ">>vault:")
 }
+
+func hasInlineVaultDelimiters(value string) bool {
```
bonifaido, line 46: Sorry, I just checked that we use `${` `}` instead of the double `{{`.

https://github.com/banzaicloud/bank-vaults/blob/9c8215ecde27930d10e20605ea817cf2756cbf30/internal/configuration/template.go#L46

cmd/vault-secrets-webhook/configmap.go
```diff
@@ -88,6 +101,28 @@ func (mw *mutatingWebhook) mutateConfigMapData(configMap *corev1.ConfigMap, data
 	return nil
 }
 
+func (mw *mutatingWebhook) mutateInlineConfigMapData(configMap *corev1.ConfigMap, data map[string]string, vaultClient *vault.Client, vaultConfig VaultConfig) error {
+	r, err := regexp.Compile("\\${{(.*?)}}")
```
bonifaido, line 105: We should compile the regex beforehand into "constant" variable I think.

cmd/vault-secrets-webhook/object.go
```diff
@@ -95,6 +98,22 @@ func traverseObject(o interface{}, vaultClient *vault.Client, vaultConfig VaultC
 				}
 
 				e.Set(dataFromVault["data"])
+			} else if hasInlineVaultDelimiters(s) {
+				r, err := regexp.Compile("\\${{(.*?)}}")
```
bonifaido, line 102: And here as well, also this code part looks very much the same as for Secret.

bonifaido (changes requested):

kschu91:
@bonifaido

I have extracted the matching logic into its own function to reduce duplicate code. However, I am not that familiar with go and have no idea how to do the same with the setting logic as there are different types of data coming in for configmaps, secrets and objects. But maybe this is good enough already?

Also, I globalized the compiliation of the regex, changed the behaviour to match only single brackets `${...}` and optimized the regex to match only single brackets containing `vault:` and `>>vault:`.

bonifaido (approved):
LGTM, the last thing it would be nice to have to add some kind of small test-case for this to make sure it doesn't break in the future: https://github.com/banzaicloud/bank-vaults/blob/master/hack/acceptance-test.sh#L137

pbalogh-sa (approved):

hack/acceptance-test.sh
```diff
@@ -140,11 +140,13 @@ helm upgrade --install vault-secrets-webhook ./charts/vault-secrets-webhook \
 kubectl apply -f deploy/test-secret.yaml
 test `kubectl get secrets sample-secret -o jsonpath='{.data.\.dockerconfigjson}' | base64 --decode | jq -r '.auths[].username'` = "dockerrepouser"
 test `kubectl get secrets sample-secret -o jsonpath='{.data.\.dockerconfigjson}' | base64 --decode | jq -r '.auths[].password'` = "dockerrepopassword"
+test `kubectl get secrets sample-secret -o jsonpath='{.data.inline}' | base64 --decode` = "Inline: secretId AWS_ACCESS_KEY_ID"
```
bonifaido, line 143: The first argument needs to be quoted I think otherwise thanks for the test!

```
++ kubectl get secrets sample-secret -o 'jsonpath={.data.inline}'
++ base64 --decode
+ test Inline: secretId AWS_ACCESS_KEY_ID = 'Inline: secretId AWS_ACCESS_KEY_ID'
hack/acceptance-test.sh: line 140: test: too many arguments
```

hack/acceptance-test.sh
```diff
@@ -140,11 +140,13 @@ helm upgrade --install vault-secrets-webhook ./charts/vault-secrets-webhook \
...
 
 kubectl apply -f deploy/test-configmap.yaml
 test `kubectl get cm sample-configmap -o jsonpath='{.data.aws-access-key-id}'` = "secretId"
 test "$(kubectl get cm sample-configmap -o jsonpath='{.data.aws-access-key-id-formatted}')" = "AWS key in base64: c2VjcmV0SWQ="
 test `kubectl get cm sample-configmap -o jsonpath='{.binaryData.aws-access-key-id-binary}'` = "secretId"
+test `kubectl get cm sample-configmap -o jsonpath='{.data.inline}'` = "Inline: secretId AWS_ACCESS_KEY_ID"
```
bonifaido, line 149: And here as well.

kschu91:
> LGTM, the last thing it would be nice to have to add some kind of small test-case for this to make sure it doesn't break in the future: https://github.com/banzaicloud/bank-vaults/blob/master/hack/acceptance-test.sh#L137

@bonifaido We have updated the acceptance tests. Would be great if we can hava a merge along with a new release soon!

bonifaido:
No worries, a new release will performed tomorrow I think!

bonifaido:
https://github.com/banzaicloud/bank-vaults/pull/1157

PR #1145, opened 2020-10-14
add check for EnvFromPath to be empty for continuation

rocpatel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
The pod mutation checks to see if there are any env variables
that have vault keys in them, if there are no env variables it
continues on and doesn't mutate the pod.  This fix also checks if
the EnvFromPath annotation is empty, if it's empty and there are no 
env variables that reference vault keys it continues without mutating the pod.

### Why?
If you have a pod that only has the envFromPath annotation and no env variables that reference vault keys, the pod does not get mutated.

### Checklist

### DIFF
--- a/cmd/vault-secrets-webhook/pod.go
+++ b/cmd/vault-secrets-webhook/pod.go
@@ -224,7 +224,7 @@ func (mw *mutatingWebhook) mutateContainers(containers []corev1.Container, podSp
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


bonifaido (approved):
LGTM, thanks for the fix!

PR #1163, opened 2020-10-28
feat: annotations to control agent and env images

michael-todorovic:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
This PR adds 4 new annotations so we can alter the images for vault agent and vault-env at pod level.

### Why?
This is useful when you need to test new images without breaking everything or deploy a whole new environment. This can be useful when you have multiple vaults with different versions, depending on the team. It can be useful if team A has vault v2 with a breaking change from v1 but team B has still vault v1.

### Checklist

### To Do

- Review the doc for this PR: https://github.com/banzaicloud/bank-vaults-docs/pull/44

### DIFF
--- a/cmd/vault-secrets-webhook/main.go
+++ b/cmd/vault-secrets-webhook/main.go
@@ -82,6 +82,8 @@ type VaultConfig struct {
 	AgentMemory                 resource.Quantity
 	AgentImage                  string
 	AgentImagePullPolicy        corev1.PullPolicy
+	EnvImage                    string
+	EnvImagePullPolicy          corev1.PullPolicy
 	Skip                        bool
 	VaultEnvFromPath            string
 	TokenAuthMount              string
@@ -89,9 +91,7 @@ type VaultConfig struct {
 
 func init() {
 	viper.SetDefault("vault_image", "vault:latest")
-	viper.SetDefault("vault_image_pull_policy", string(corev1.PullIfNotPresent))
 	viper.SetDefault("vault_env_image", "banzaicloud/vault-env:latest")
-	viper.SetDefault("vault_env_image_pull_policy", string(corev1.PullIfNotPresent))
 	viper.SetDefault("vault_ct_image", "hashicorp/consul-template:0.24.1-alpine")
 	viper.SetDefault("vault_addr", "https://vault:8200")
 	viper.SetDefault("vault_skip_verify", "false")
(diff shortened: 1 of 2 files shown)


cmd/vault-secrets-webhook/main.go
```diff
@@ -82,16 +82,16 @@ type VaultConfig struct {
...
+	EnvImage                    string
+	EnvImagePullPolicy          corev1.PullPolicy
 	Skip                        bool
 	VaultEnvFromPath            string
 	TokenAuthMount              string
 }
 
 func init() {
 	viper.SetDefault("vault_image", "vault:latest")
-	viper.SetDefault("vault_image_pull_policy", string(corev1.PullIfNotPresent))
```
bonifaido, line 92: I think these shouldn't be removed to keep backward compatibility if people have defined this in a global manner.

cmd/vault-secrets-webhook/main.go
```diff
@@ -341,8 +341,41 @@ func parseVaultConfig(obj metav1.Object) VaultConfig {
...
+		switch val {
+		case "Never", "never":
+			vaultConfig.EnvImagePullPolicy = corev1.PullNever
+		case "Always", "always":
+			vaultConfig.EnvImagePullPolicy = corev1.PullAlways
+		case "IfNotPresent", "ifnotpresent":
+			vaultConfig.EnvImagePullPolicy = corev1.PullIfNotPresent
+		}
+	} else {
+		vaultConfig.EnvImagePullPolicy = corev1.PullIfNotPresent
```
bonifaido, line 359: And the aforementioned viper key should be used here. (`vault_env_image_pull_policy`)

cmd/vault-secrets-webhook/main.go
```diff
@@ -341,8 +341,41 @@ func parseVaultConfig(obj metav1.Object) VaultConfig {
...
+		switch val {
+		case "Never", "never":
+			vaultConfig.AgentImagePullPolicy = corev1.PullNever
+		case "Always", "always":
+			vaultConfig.AgentImagePullPolicy = corev1.PullAlways
+		case "IfNotPresent", "ifnotpresent":
+			vaultConfig.AgentImagePullPolicy = corev1.PullIfNotPresent
+		}
+	} else {
+		vaultConfig.AgentImagePullPolicy = corev1.PullIfNotPresent
```
bonifaido, line 377: And the aforementioned viper key should be used here. (`vault_image_pull_policy `)

bonifaido (changes requested):
LGTM, thank you! Just the policy `imagePullPolicy` default values shouldn't be hardcoded.

cmd/vault-secrets-webhook/main.go
michael-todorovic: I wanted to use viper at first but got issues (I'm not experience with go yet) so I actually copied the code from consul-template [pull policy](https://github.com/banzaicloud/bank-vaults/blob/master/cmd/vault-secrets-webhook/main.go#L227-L237) :)
I'll dig a bit more to get back viper

bonifaido (approved):
LGTM, thank you @michael-todorovic

PR #1168, opened 2020-10-30
fix vault agent `-once` command to exit after auth

joshdvir:
Q | A
-- | --
Bug fix | Yes
New feature | no
API breaks | no
Deprecations | no
Related tickets | none
License | Apache 2.0

### What's in this PR?
This PR fixes the command passing the `vault agent` command to exit after the auth process. 

### Why?
`vault agent` command doesn't support the current `-once` attribute, the correct attribute is `-exit-after-auth`

### Checklist

### DIFF
--- a/cmd/vault-secrets-webhook/pod.go
+++ b/cmd/vault-secrets-webhook/pod.go
@@ -680,7 +680,7 @@ func getContainers(vaultConfig VaultConfig, containerEnvVars []corev1.EnvVar, co
 
 	var ctCommandString []string
 	if vaultConfig.CtOnce {
-		ctCommandString = []string{"-config", "/vault/ct-config/config.hcl", "-exit-after-auth"}
+		ctCommandString = []string{"-config", "/vault/ct-config/config.hcl", "-once"}
 	} else {
 		ctCommandString = []string{"-config", "/vault/ct-config/config.hcl"}
 	}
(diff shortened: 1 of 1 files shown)


cmd/vault-secrets-webhook/pod.go
```diff
@@ -680,7 +680,7 @@ func getContainers(vaultConfig VaultConfig, containerEnvVars []corev1.EnvVar, co
 
 	var ctCommandString []string
 	if vaultConfig.CtOnce {
-		ctCommandString = []string{"-config", "/vault/ct-config/config.hcl", "-once"}
+		ctCommandString = []string{"-config", "/vault/ct-config/config.hcl", "-exit-after-auth"}
```
bonifaido, line 683: I think this line should be changed instead: https://github.com/banzaicloud/bank-vaults/blob/761e771169aee388f38f745d855ea28b65bdbc77/cmd/vault-secrets-webhook/pod.go#L744

bonifaido (changes requested):
see comment ☝️

cmd/vault-secrets-webhook/pod.go
joshdvir: @bonifaido you are correct, fixed!

joshdvir:
@bonifaido fixed, thanks!

bonifaido (approved):
Thanks for the fix!

PR #1181, opened 2020-11-12
Use agent ConfigMap when agent is in the InitContainers

joshdvir:
| Q               | A
| --------------- | ---
| Bug fix        | no
| New feature    | yes
| API breaks     | no
| Deprecations   | no
| Related tickets | none

### What's in this PR?
This PR allows to use a ConfigMap with the agent when the agent is in the initContainers

### Why?
I want to make sure the agent runs before all other containers and exits since I want to use the token the agent sinks directly in my application. 

### Checklist


bonifaido (approved):
LGTM @joshdvir, thanks for this!

PR #1196, opened 2020-11-26
allow the vault-webhook to be contacted over loadbalancer or ingress

AtzeDeVries:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no|
| Deprecations?   | nos

### What's in this PR?
MutatingWebhook can also connect to an url. Currently this can only be an (k8s) service. This PR add the ability to create 
a k8s Loadbalancer or k8s Ingress and use the url method

### Why?
Using the Service method can cause the webhook to be temporary unreachable. Check the following issues
https://github.com/kubernetes/kubernetes/issues/80313
https://github.com/kubernetes/client-go/issues/374#issuecomment-632457187
In short, the client-go library caches the ip of the pod so if the pod gets  an different ip the cache needs to invalidate
which takes time. Using a url instead of the service is a workaround until this is fixed in k8s. 

### Checklist

### DIFF
--- a/charts/vault-secrets-webhook/Chart.yaml
+++ b/charts/vault-secrets-webhook/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: vault-secrets-webhook
-version: 1.8.1
+version: 1.8.2
 appVersion: 1.8.0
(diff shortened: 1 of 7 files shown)


bonifaido (approved):
Great work @AtzeDeVries ! Thank you very much!

PR #1239, opened 2021-01-20
[webhook] Kill container process when non-renewable lease expired.

moskitone:
Signed-off-by: Volodymyr Moseichuk <moskitone@gmail.com>

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | Messages in slack

### What's in this PR?
[webhook] vault.security.banzaicloud.io/vault-env-daemon-exit-when-expired - New annotation for terminating pods with expired non-renewable secrets in vault-env daemon mode.
[chart-webhook] - extend RBAC with ability to get serviceaccounts (required by new go-containerregistry).

### Why?
Due to limitations in vault 
https://github.com/hashicorp/vault/blob/master/builtin/logical/aws/secret_access_keys.go#L200-L201
and https://github.com/banzaicloud/bank-vaults/blob/master/internal/injector/injector.go#L254 
this part https://github.com/banzaicloud/bank-vaults/blob/master/cmd/vault-env/main.go#L104-L111 newer starts. After lease expiration container remains in RUNNING state with expired credentials.

### Additional context
Tested only with aws sts credentials.

### Checklist

### DIFF
--- a/charts/vault-secrets-webhook/Chart.yaml
+++ b/charts/vault-secrets-webhook/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: vault-secrets-webhook
-version: 1.11.0
+version: 1.11.1
 appVersion: 1.11.0
 description: A Helm chart that deploys a mutating admission webhook that configures applications to request env vars from Vault Secrets
 home: https://banzaicloud.com/products/bank-vaults/
--- a/charts/vault-secrets-webhook/templates/webhook-rbac.yaml
+++ b/charts/vault-secrets-webhook/templates/webhook-rbac.yaml
@@ -33,6 +33,12 @@ rules:
   - apiGroups:
       - ""
     resources:
+      - serviceaccounts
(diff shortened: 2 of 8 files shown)


bonifaido:
- I think this should be the default behavior, so I wouldn't introduce a new annotation for this, what do you think @moskitone?
- Could you please bump the chart version?

bonifaido (changes requested):
See my comments above.

moskitone:
> * I think this should be the default behavior, so I wouldn't introduce a new annotation for this, what do you think @moskitone?

Just wanted to be on the safe side and didn't want to break default behaviour.

> * Could you please bump the chart version?

Sure

moskitone:
Hi @bonifaido 
Could you please check it?

bonifaido (approved):
LGTM! Thanks @moskitone nice fix!

PR #1255, opened 2021-02-11
Delete obsolete reference to try pipeline

fekete-robert:
.


bonifaido (commented):
I think the core parts section could remain here since it is still true. I would delete only after
>  Check out...

bonifaido (approved):

PR #1271, opened 2021-02-25
[webhook] Fix for inline mutations in custom resources

dervoeti:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Fix for inline mutations in custom resources

### Why?
Inline mutations in custom resources did not work for me (values were not replaced), i looked into the source code and it seems like the data gets successfully retrieved from vault, but the object (custom resource) is not updated afterwards. I tested the fixed version in my cluster and it now works as expected.

To test it, create a custom resource definition, add it to the "customResourceMutations" list in the helm chart values and create a custom resource with an annotation, like:
```
metadata:
  annotations:
    foo: "secret is ${vault:secrets/data/foo#bar}"
```

It does not work for me with the current version, when i apply the fix from this PR it works.

### Checklist


bonifaido (approved):
Nice catch!

PR #1287, opened 2021-03-31
vault-secrets-webhook: Add missing objectSelector configuration options

gw0:
| Q               | A
| --------------- | ---
| Bug fix?        | kind of
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | no

### What's in this PR?

This PR adds missing `objectSelector` configuration options to the MutatingWebhookConfiguration resource. On one hand the `objectSelector` was completely missing for some hooks, on the other it did not support the `matchLabels` selector option (to provide the same options as `namespaceSelector`).

In addition, this PR adds a default `objectSelector` rule to not trigger on internal Helm v3 state resources (can be in Secrets or ConfigMaps).

### Why?

My attempts to use the `objectSelector` configuration to prevent the `vault-secrets-webhook` to trigger on internal Helm v3 state resources have failed. By default Helm v3 state is stored in Secrets with a label `owner=helm`, so one would expect this to work:
```yaml
objectSelector:
  matchExpressions:
    - key: owner
      operator: NotIn
      values:
        - helm
```

Unfortunately, it did not work, because hook for the Secret resources was missing the `objectSelector` configuration (it only had
`namespaceSelector`).

### Checklist

### DIFF
--- a/charts/vault-secrets-webhook/Chart.yaml
+++ b/charts/vault-secrets-webhook/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: vault-secrets-webhook
-version: 1.11.6
+version: 1.11.7
 appVersion: 1.11.3
 description: A Helm chart that deploys a mutating admission webhook that configures applications to request env vars from Vault Secrets
 home: https://banzaicloud.com/products/bank-vaults/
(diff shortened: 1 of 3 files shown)


bonifaido (approved):
LGTM, great fix! Thank you

PR #1295, opened 2021-04-09
[webhook] vault-ct-once as initContainer

moskitone:
| Q               | A
| --------------- | ---
| Bug fix?        | maybe
| New feature?    | maybe
| API breaks?     | yes
| Deprecations?   | no

### What's in this PR?

Changed order of running consul-template when `vault.security.banzaicloud.io/vault-ct-once` annotation is set.
Now it runs after all initContainers

### Why?

In some cases, for example cronjobs, where there is no restart policy or restarts are disabled, consul-template runs after the main container and it causes wrong behaviour other then expected from pod lifecycle. 
I suggest to run consul-template in case `vault.security.banzaicloud.io/vault-ct-once` is set as the last of initContainers to fix this issue.

### Checklist


bonifaido (approved):
Thank you very much @moskitone ! Nice fix.

pbalogh-sa (approved):

Issue #1290, opened 2021-04-01
vault agent templates for the init containers


dzirg44:
**Is your feature request related to a problem? Please describe.**
Hi. As I discovered vault agent doesn't modify  'init containers' , it can be useful in applications where we have DB migrations.  

**Describe the solution you'd like**
maybe something like this
```yaml
        vault.security.banzaicloud.io/vault-addr: "https://vault.default:8200"
        vault.security.banzaicloud.io/vault-path: "kubernetes"
        vault.security.banzaicloud.io/vault-role: "dev"
        vault.security.banzaicloud.io/vault-skip-verify: "true"
        vault.security.banzaicloud.io/vault-agent-configmap: vault-agent-pki
       vault.security.banzaicloud.io/vault-init-agent-configmap: vault-agent-pki
```
of course I would like to have rendered secrets  via  agent only in certain containers , but I don't know how to achieve it.

**Describe alternatives you've considered**
I made a custom chart where I have one short lived init container, where I get token, and another init container where I use it token via consul-template. If everything works fine,a deployment will be created as well.

PR #1302, opened 2021-04-28
feat(consul-template): add configuration for template injecting for in init containers

kratisto:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no 
| Related tickets | Partially : #1290 (covers only the consul template part) 

### What's in this PR?

The goal is to add the injection of the configuration generated by consul template in the init containers. 
Obviously this is only possible when not in daemon mode.
I had to move the block injecting the consul-template part in order to guarantee the correct execution order of the init containers. That is to say : 
- vault_agent
- consul_template
- user init containers. 

If I hadn't done so, here is the execution order: 
- consul_template
- vault_agent
- user init containers. 

In order to solve this problem, it was easier to move this code than to go through the list and insert it after the vault-agent. 

I chose to add an activation configuration in order not to make the injection automatic. 

### Why?
I have developed an api that uses liquibase. In order to avoid checking the database every time the application container starts, we decided to use an init container for liquibase. So we need to have access to the configuration generated by consul-template in this init container
### Checklist


bonifaido (changes requested):
Hi @kratisto, there are still some formatting errors, and also could you please sign the CLA?

bonifaido (approved):
LGTM! We now just need to have the CLA signed ;)

kratisto:
@bonifaido 
Sorry, I thought I had. 

I just signed it.

Are you planning a new version soon?

pbalogh-sa (approved):

bonifaido:
Thanks for the feature @kratisto !

Release: maybe today, but Monday at the latest.

PR #1324, opened 2021-05-14
Add awskms seal configuration to example

miguelaferreira:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | None


### What's in this PR?
This PR adds a piece of configuration to a Vault CR configuration.


### Why?
I'm not sure, but I think this configuration is needed to enable the `awskms` seal, which seems to be the purpose of the said configuration.


### Additional context
I used the configuration as it was and when starting, vault did not print out the expected lines in the logs. Expected lines being:
```
==> Vault server configuration:
         AWS KMS KeyID: some-key-id                                                                                                                                                                                                                                                        
        AWS KMS Region: a-region
```


### Checklist
<!-- Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields. -->

### DIFF
--- a/operator/deploy/cr-awskms.yaml
+++ b/operator/deploy/cr-awskms.yaml
@@ -51,6 +51,12 @@ spec:
         tls_cert_file: /vault/tls/server.crt
         tls_key_file: /vault/tls/server.key
     ui: true
+    # Configure Vault to use AWS KMS seal
+    # See: https://www.vaultproject.io/docs/configuration/seal/awskms for more information.
+    # seal:
+    #   awskms:
+    #     kms_key_id: some-key-id
+    #     region: us-east-1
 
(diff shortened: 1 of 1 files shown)


bonifaido:
This part is only needed if you are using the Vault auto-unseal option, but since Bank-Vaults is using its own unseal mechanism as well (with more options that are built into Vault) this is not required (but works well together). We can leave the seal stanza in as commented out mentioning this in a comment. Thanks for this!

miguelaferreira:
Thanks for getting back to me so quickly @bonifaido. I gave it a shot at creating the comment.

bonifaido (approved):
LGTM, thanks!

PR #1355, opened 2021-06-11
charts/vault: openshift support

### LINKED ISSUE #1353: [Helm Vault] Unable to deploy to OpenShift due to hard-coded fsgroup
It is not currently possible to deploy this chart to OpenShift, due to the pod fsGroup being hard-coded to 65534.

By default, each project in Openshift has an individual allocation of allowed group IDs, meaning a hard-coded default will not work, as a project will not generally have permission to use the requested group.

My preference for working around this would be to follow the solution used by the Hashicorp Vault chart, and define a global flag that is set if the chart is being deployed to OpenShift, i.e. `globals.openshift` The securityContext block would then be wrapped within an if statement which checks that this flag is false, i.e.:

```
{{- if not .Values.global.openshift }}
securityContext:
  fsGroup: 65534
{{- end }}
```

An alternative would be to add a value for the fsGroup, setting the default in `values.yaml`, i.e.:

```
securityContext:
  fsGroup: {{ .Values.securityContext.fsGroup }}
```

**Additional context**
Happy to raise a PR for either of these changes

chris1786:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1353 

### What's in this PR?

- Adds a global `openshift` flag to be set when deploying to OpenShift. This removes the hard-coded fsGroup from the stateful set
- Explicitly sets the required verbs in the secret-access Role - the default OpenShift config rejects the wildcard currently used.

### DIFF
--- a/charts/vault/Chart.yaml
+++ b/charts/vault/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: vault
-version: 1.13.0
+version: 1.13.1
 appVersion: 1.13.0
 description: A Helm chart for Vault, a tool for managing secrets
 home: https://www.vaultproject.io/
--- a/charts/vault/templates/role.yaml
+++ b/charts/vault/templates/role.yaml
@@ -13,7 +13,7 @@ rules:
 - apiGroups: [""]
   resources: ["secrets"]
   resourceNames: ["bank-vaults"]
-  verbs: ["*"]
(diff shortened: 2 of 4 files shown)


bonifaido (changes requested):
Could you please bump the chart version?

bonifaido (approved):

PR #1401, opened 2021-08-18
[webhook] Draft of namespaced service accounts

### LINKED ISSUE #929: [vault-secrets-webhook] namespaced access to secrets
I'm currently trying out the vault-secrets-webhook on top of an existing Vault deployment.
I'm currently using a vault-template sidecar along with existing deployments to load secrets and trying to figure out how to migrate to the webhook.

For authentication to Vault I'm using a dedicated ServiceAccount in each namespace. Therefore the vault policies and the roles used to authenticate are also namespaced to the specific deployments running in the cluster. Apps can only access the secret their individual ServiceAccount from their namespace has access to.

As far as I understand the vault-secrets-webhook, access to all Vault data is authenticated with a single ServiceAccount, residing in the namespace of the webhook.

From a security perspective, I would like to have an option to let the webhook use a ServiceAccount in the namespace of the to-be-mutated resource. Otherwise, it would be possible for any resource in any namespace to query for any secret that the webhook has access to.

It is quite possible, that this is already possible with the webhook and I didn't figure out. If not I would propose to allow the webhook to be configured to use a ServiceAccount in the namespace of the resource requesting a secret, possibly also auto-creating this service account

timbuchwaldt:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #929

### What's in this PR?

Add a new authentication method that uses a token from the same namespace as the resource.

### Why?

Non-pod resources are mutated with access privileges of the mutating webhook container. This doesn't allow for least privilege principles.

### Additional context

This is just a first draft to gather some feedback on the general direction.

### Checklist

- [ ] Code meets the [Developer Guide](https://github.com/banzaicloud/developer-guide)

### To Do

- [ ] Validate general idea
- [ ] Add tests

### DIFF
--- a/pkg/webhook/config.go
+++ b/pkg/webhook/config.go
@@ -16,12 +16,14 @@ package webhook
 
 import (
 	"strconv"
+	"context"
 	"time"
 
 	"github.com/spf13/viper"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
 // VaultConfig represents vault options
@@ -72,9 +74,10 @@ type VaultConfig struct {
 	EnvCPULimit                 resource.Quantity
 	EnvMemoryLimit              resource.Quantity
 	VaultNamespace              string
+	ServiceAccountJWT              string
 }
 
-func parseVaultConfig(obj metav1.Object) VaultConfig {
+func parseVaultConfig(ctx context.Context, obj metav1.Object, k8s_client kubernetes.Interface) VaultConfig {
 	var vaultConfig VaultConfig
 	annotations := obj.GetAnnotations()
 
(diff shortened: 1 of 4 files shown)


bonifaido (commented):
I think the idea is really good, we got this request every once in a while, so thank you very much for taking the time to implement it!

pkg/webhook/config.go
```diff
@@ -117,6 +120,26 @@ func parseVaultConfig(obj metav1.Object) VaultConfig {
 		vaultConfig.Path = viper.GetString("vault_path")
 	}
 
+	// TODO: Check for flag to verify we want to use namespace-local SAs instead of the vault webhook namespaces SA
+	if val, ok := annotations["vault.security.banzaicloud.io/vault-serviceaccount"]; ok {
+		vaultConfig.AuthMethod = "namespaced"
+		sa, err := k8s_client.CoreV1().ServiceAccounts(obj.GetNamespace()).Get(ctx,val, metav1.GetOptions{})
```
bonifaido, line 126: I would move the reading of the Secret ot the client.go parts since here no real error handling is present (except logging) and it is not even expected to touch Kubernetes here, we are just parsing the config.

bonifaido (changes requested):
Please move the Secret reading to `newVaultClient()`

timbuchwaldt:
Hey @bonifaido - thanks for the feedback, that sounds doable :) I'll see what I can get working tomorrow :)

pkg/webhook/webhook.go
```diff
@@ -207,13 +207,35 @@ func (mw *MutatingWebhook) newVaultClient(vaultConfig VaultConfig) (*vault.Clien
...
+		if err != nil {
+			return nil, errors.Wrap(err, "Failed to retrieve secret for specified service account")
+		}
+		return vault.NewClientFromConfig(
+			clientConfig,
+			vault.ClientRole(vaultConfig.Role),
+			vault.ClientAuthPath(vaultConfig.Path),
+			vault.ClientAuthMethod(vaultConfig.AuthMethod),
+			vault.ClientLogger(logrusadapter.NewFromEntry(mw.logger)),
+			vault.ClientToken(secret.Data["token"]),
```
timbuchwaldt, line 226: I'm almost certain there is a better way of adding this parameter here.

bonifaido:
Please format the code according to the CI check messages.

timbuchwaldt:
Please not this currently doesn't work, I'll be back to testing why today.

bonifaido:
We still have lint issue here, otherwise LGTM! :)

https://github.com/banzaicloud/bank-vaults/pull/1401/checks?check_run_id=3410442298#step:7:13

PR #1407, opened 2021-08-31
fix(webhook): wrong namespace used for serviceaccount

dquagebeur:
mutated object namespace must be used to inherit  vault policies

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
With the new feature serviceAccount, it is possible to choose which serviceAccount is used by the webhook to connect to Vault. The problem for me is that this serviceAccount is searched/used in the vault-webhook namespace instead of the mutated object namespace.

### Why?
 In our case, vault policies are namespaced and with this fix, vault-webhook namespace doesn't need to have access on all secrets...

### DIFF
--- a/pkg/sdk/vault/client.go
+++ b/pkg/sdk/vault/client.go
@@ -464,7 +464,13 @@ func NewClientFromRawClient(rawClient *vaultapi.Client, opts ...ClientOption) (*
 
 			case NamespacedSecretAuthMethod:
 				loginDataFunc = func() (map[string]interface{}, error) {
-					// Projected SA JWTs do expire, so we need to move the reading logic into the loop
+					if len(o.existingSecret) > 0 {
+						return map[string]interface{}{
+							"jwt":  o.existingSecret,
+							"role": o.role,
+						}, nil
+					}
+
 					jwt, err := ioutil.ReadFile(jwtFile)
 					if err != nil {
 						return nil, err
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -39,14 +39,17 @@ import (
 )
 
 type MutatingWebhook struct {
-	k8sClient kubernetes.Interface
-	namespace string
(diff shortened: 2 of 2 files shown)


bonifaido:
One question, does this break the previous functionality regarding the location of the secret it expected?

dquagebeur:
I'm not sure what Tim has in mind when he implements this feature : 
1- the secret has to be in the vault-webhook namespace (it is just an alternative service account): in that case, yes, there is a break
2- the secret has to be in the mutating object namespace, so the webhook works like vault-env with Pod : in that case, it is not a break, but a bug fix.
I have asked my teammates, and for everyone (including me), the logical answer is 2. Do you agree with us ?

bonifaido:
Yes TBH after giving some thought to it I agree, that option 2 is the better solution here. And yes that better represents the vault-env experience here.

Since this feature is not yet released we don't break anything just the original PR authors use-case :)

Can we make it configurable so by default it would search the Secret's namespace and if the users would like to have it otherwise it would check that namespace? I would suggest to transform this annotation this way:

```yaml
Your way:
vault.security.banzaicloud.io/vault-serviceaccount: blabla # The annotated objects namespace will be searched for the blabla SA.

Tim's way:
vault.security.banzaicloud.io/vault-serviceaccount: mynamespace/blabla # The mynamespace will be checked for the blabla SA.
```

WDYT?

dquagebeur:
TBH I don't really see a use case on a production cluster. And it would be complicated regarding rights management, as namespace A secret could use a reference to namespace B service account to access secrets of namespace B.
At least, it required a global configuration attribute to enable/disable the support of namespace in service account name.

bonifaido:
I also agree, my bad, that I haven't cached this at the initial PR of this feature.

dquagebeur:
No problem, thanks for the merge.

Issue #1398, opened 2021-08-13
[vault-env] Fetch image descriptor ignore registry mirror.


fdpeiter:
> Bugs should be filed for issues encountered whilst operating bank-vaults.
> You should first attempt to resolve your issues through the community support
> channels, e.g. Slack, in order to rule out individual configuration errors.
> Please provide as much detail as possible.

**Describe the bug**:
After DockerHub implemented it's rate limit, we've pushed via Kops a configuration to add a Registry Mirror to the nodes. Although this worked fine for pods that doesn't use vault secrets, whenever a pod need mutation, the configuration seems to be ignored:

```
time="2021-08-13T19:36:35Z" level=error msg="admission webhook error: cannot fetch image descriptor: GET https://index.docker.io/v2/tfgco/fluxcloud/manifests/v0.5.0: TOOMANYREQUESTS: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit" app=vault-secrets-webhook
```

Checking the documentation, we've noticed two environment variables that should help with this, enabling us to inject a DockerHub credential using a Kubernetes Secret: https://github.com/banzaicloud/bank-vaults/blob/master/charts/vault-secrets-webhook/values.yaml#L59

We've followed the https://banzaicloud.com/docs/bank-vaults/mutating-webhook/configuration/#registry-access tutorial, creating a secret with the DockerHub credentials:

```bash
kubectl -n vault-secrets-webhook create secret docker-registry dockerhub --docker-username=${DOCKER_USERNAME} --docker-password=$DOCKER_PASSWORD
```

We've then proceeded to define both env variables:
```yaml
        - name: DEFAULT_IMAGE_PULL_SECRET
          value: dockerhub
        - name: DEFAULT_IMAGE_PULL_SECRET_NAMESPACE
          value: vault-secrets-webhook
```
The errors still persisted and checking the code, we don't believe one of the conditions for the variables to be used will ever be true:  https://github.com/banzaicloud/bank-vaults/blob/b38c0a7d868a75f549a1307a585d425bbe1e4c06/pkg/webhook/registry.go#L108. It seems that the `containerInfo.Namespace` will never be empty.

**Expected behaviour**:
After configuring the DockerHub credentials and both default values for the pull secrets, we've expected pods that don't have the ImagePullSecrets configuration  used the default ones to authenticate in DockerHub.

**Steps to reproduce the bug**:
[...]

PR #1409, opened 2021-09-02
fix(webhook-registry): use default value if no ImagePullSecrets was specified in podspec

### LINKED ISSUE #1384: [vault-secrets-webhook] helm deployments fail on AKS
**Describe the bug**:
When doing helm deployments into the cluster replicasets fail to create in AKS with the error message seen below. Tested manually deploying a deployment using kubectl and the same image from the same registry and this deploys fine and is able to get secrets from vault. Also to note a helm deployment of that creates a job and not a replicaset didn't have any issues.

As a note we are running this config in AWS on EKS with no issues.

**Expected behaviour**:
replicaset's deploy fine with no issues

AKS cluster with vault and vault-secret webhooks charts installed
Private acr registry for images

**Additional context**:
N/A

**Environment details**:
- Kubernetes version (e.g. v1.10.2): 1.19
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc): AKS
- bank-vaults version (e.g. 0.4.17): vault-secret-webhook helm chart - 1.13.2, vault chart - 1.13.1
- Install method (e.g. helm or static manifests): helm
- Logs from the misbehaving component (and any other relevant logs): these are the logs we get from the replicaset when trying to deploy via helm
`Warning FailedCreate 110s replicaset-controller Error creating: Internal error occurred: failed calling webhook "pods.vault-secrets-webhook.admission.banzaicloud.com": an error on the server ("{\"kind\":\"AdmissionReview\",\"apiVersion\":\"admission.k8s.io/v1beta1\",\"response\":{\"uid\":\"887288aa-a5f3-46d3-b212-01b085cf2aa5\",\"allowed\":false,\"status\":{\"metadata\":{},\"status\":\"Failure\",\"message\":\"could not mutate object: cannot fetch image descriptor: GET <removed azure private registry address>: UNAUTHORIZED: authentication required, visit https://aka.ms/acr/authorization for more information.\"}}}") has prevented the request from succeeding`
```
debug: true

certificate:
  useCertManager: true
  generate: false

env:
  VAULT_IMAGE: <private registry>/vault:1.6.2

vaultEnv:
  repository: <private registry>/vault-env

metrics:
  enabled: true
  serviceMonitor:
    enabled: true
    scheme: https
    tlsConfig:
      insecureSkipVerify: true
    additionalLabels:
      release: prometheus

resources:
  requests:
    cpu: 50m
    memory: 32Mi
  limits:
    cpu: 500m
    memory: 256Mi

image:
  repository: <private registry>/vault-secrets-webhook

affinity:
  podAntiAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
    - labelSelector:
        matchLabels:
          app.kubernetes.io/name: vault-secrets-webhook
      topologyKey: "kubernetes.io/hostname"
    preferredDuringSchedulingIgnoredDuringExecution:
    - weight: 100
      podAffinityTerm:
        labelSelector:
          matchLabels:
            app.kubernetes.io/name: vault-secrets-webhook
        topologyKey: "failure-domain.beta.kubernetes.io/zone"

podsFailurePolicy: Fail

objectSelector:
  matchExpressions:
  - key: k8s-app
    operator: NotIn
    values:
    - kube-proxy

tolerations:
  - operator: Equal
    key: arch
    value: arm64
    effect: NoSchedule
```

/kind bug

kratisto:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | May fix  #1384 and #1398

### What's in this PR?
As @fdpeiter suggested, `containerInfo.Namespace` seems to never be empty.  So I changed the condition to only care about the presence of a credential in the pod spec. This will allow to use the default configuration for the auth k8s chain. 
I tried to implement tests but I ran into the use of the remote package which retrieves the image blob.

With this modification, we will always keep the priority of the presence of the configuration in the pod being deployed. 
And this will allow the default configuration to be used in case of absence in this one. 
Since the default value of the two parameters is empty. I don't see any possible impact if the default values are not configured

In order to avoid creating the secret for every possible namespace, I think you should override the namespace with the one present in the configuration. But it makes me worry from a security point of view. I'm not sure it's a good thing to make a private registry accessible to a whole cluster

**_I don't have an Azure cluster available or a private registry to test._** 

### Checklist

### DIFF
--- a/pkg/webhook/registry.go
+++ b/pkg/webhook/registry.go
@@ -105,7 +105,7 @@ func (r *Registry) GetImageConfig(
 	// The pod imagePullSecrets did not contained any credentials.
 	// Try to find matching registry credentials in the default imagePullSecret if one was provided.
 	// Otherwise cloud credential providers will be tried.
-	if containerInfo.Namespace == "" && len(containerInfo.ImagePullSecrets) == 0 {
+	if len(containerInfo.ImagePullSecrets) == 0 {
 		containerInfo.Namespace = viper.GetString("default_image_pull_secret_namespace")
 		containerInfo.ImagePullSecrets = []string{viper.GetString("default_image_pull_secret")}
 	}


bonifaido (approved):
You're right @kratisto, thank you for the fix! I will do a release shortly, probably the first part of next week.

PR #1361, opened 2021-06-21
[webhook] add Vault namespaces support

anoncam:
fixes: https://github.com/banzaicloud/bank-vaults/issues/1347

PR #1412, opened 2021-09-09
Set Vault Namespace on underlying rawClient if it has a non-empty value

### LINKED ISSUE #917: Webhook unable to mutate Secret when using external Vault with Vault-namespaces
**Describe the bug**:
Our organization uses Vault Enterprise, and our team was assigned a Vault namespace.
We used the helm chart to deploy the Vault webhook to our K8S cluster, and we are able to mutate Pods and grab values from our Vault instance in our datacenter.  As an example we are able to set the Pod's environment variable with this value from Vault:
`vault:kv/data/parent/child/gc#SOME_VAR`

The key to making this work was to set the VAULT_NAMESPACE environment variable in the POD:
```
kind: Deployment
...
  template:
    metadata:
      annotations:
        vault.security.banzaicloud.io/vault-addr: "https://external-vault.com:443"
...
        env:
        - name: VAULT_NAMESPACE
          value: Foo/Bar
        - name: SOME_VAR
          value: vault:kv/data/parent/child/gc#SOME_VAR
```

We're trying to mutate a K8S secret to read the same value
```
apiVersion: v1
kind: Secret
metadata:
  name: remote-secret
  annotations:
    vault.security.banzaicloud.io/vault-addr: "https://external-vault.com:443"
    vault.security.banzaicloud.io/vault-role: "default"
    vault.security.banzaicloud.io/vault-skip-verify: "true"
    vault.security.banzaicloud.io/vault-path: "kubernetes"
type: Opaque   
data:
  # value below is vault:kv/data/parent/child/gc#SOME_VAR base64 encoded
  vaultdata: dmF1bHQ6a3YvZGF0YS9wYXJlbnQvY2hpbGQvZ2MjU09NRV9WQVIK
```
I tried to use this secret in a Deployment
```
kind: Deployment
...
  template:
    metadata:
      annotations:
        vault.security.banzaicloud.io/vault-addr: "https://external-vault.com:443"
        vault.security.banzaicloud.io/vault-role: "default"
        vault.security.banzaicloud.io/vault-path: "kubernetes"
...
      env:
        - name: VAULT_NAMESPACE
          value: Foo/Bar
        - name: DEBUG
          value: "true"
        - name: FROM_SECRET
          valueFrom:
            secretKeyRef:
              name: remote-secret
              key: vaultdata
```

However, we get the following error from the Pod
```
$ kubectl logs pod-12345
time="2020-03-11T19:39:42Z" level=info msg="Received new Vault token"
time="2020-03-11T19:39:42Z" level=info msg="Initial Vault token arrived"
time="2020-03-11T19:39:42Z" level=info msg="Renewed Vault Token"
time="2020-03-11T19:39:42Z" level=fatal msg="key not found: SOME_VAR\n" app=vault-env
```

**Expected behaviour**:
The secret from the external Vault instance from that particular namespace should be fetched.

See above

**Additional context**:

**Environment details**:
- Kubernetes version (e.g. v1.10.2): 1.14
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc): EKS
- bank-vaults version (e.g. 0.4.17): banzaicloud/vault-secrets-webhook 0.9.0 image
- Install method (e.g. helm or static manifests): helm

/kind bug

wmene:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #917 , completes #1361 

### What's in this PR?
This completes the work in PR #1361 .  This enables mutation of non-pod resources like secrets.  It sets the vault namespace on the underlying `rawClient` used in [pkg/sdk/vault/client.go](https://github.com/banzaicloud/bank-vaults/blob/master/pkg/sdk/vault/client.go)

### Why?
Even though PR #1361 parses out `vault-namespace` from a resource's annotation, it does not set that value on the underlying vault client.

### Additional context
As described in #917, this is not an issue for pods that set the `VAULT_NAMESPACE` environment variable. Setting this environment variable does get picked up by the client, see https://github.com/hashicorp/vault/blob/v1.8.2/api/client.go#L484

### Checklist

- [x ] Code meets the [Developer Guide](https://github.com/banzaicloud/developer-guide)

### DIFF
--- a/pkg/sdk/vault/client.go
+++ b/pkg/sdk/vault/client.go
@@ -144,7 +144,7 @@ func (co ExistingSecret) apply(o *clientOptions) {
 type VaultNamespace string
 
 func (co VaultNamespace) apply(o *clientOptions) {
-	o.vaultNamespace = co
+	o.vaultNamespace = string(co)
 }
 
 const (
@@ -329,10 +329,7 @@ func NewClientFromRawClient(rawClient *vaultapi.Client, opts ...ClientOption) (*
 
 	// Set vault namespace if defined
 	if o.vaultNamespace != "" {
-		err := rawClient.SetNamespace(o.vaultNamespace)
-		if err != nil {
-			return nil, err
-		}
+		rawClient.SetNamespace(o.vaultNamespace)
 	}
 
 	// Default timeout
--- a/pkg/sdk/vault/client.go
+++ b/pkg/sdk/vault/client.go
@@ -61,6 +61,7 @@ type clientOptions struct {
 	logger         Logger
 	authMethod     ClientAuthMethod
 	existingSecret string
+	vaultNamespace string
 }
 
 // ClientOption configures a Vault client using the functional options paradigm popularized by Rob Pike and Dave Cheney.
@@ -137,6 +138,13 @@ type ExistingSecret string
 
 func (co ExistingSecret) apply(o *clientOptions) {
 	o.existingSecret = string(co)
+}
+
+// Vault Enterprise Namespace (not Kubernetes namespace)
+type VaultNamespace string
+
+func (co VaultNamespace) apply(o *clientOptions) {
+	o.vaultNamespace = co
 }
 
 const (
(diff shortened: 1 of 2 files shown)


bonifaido:
We have some compile errors now.

wmene:
@bonifaido Whoops, sorry about that.  First time working with golang, but things should be ok now

bonifaido (approved):
LGTM, and thank you @wmene !

wmene:
@bonifaido Thanks for looking!  It looks like I just missed the 1.14.0 release.  Any idea about when 1.14.1 or later will come out? 😀

bonifaido:
I think we wait for a few more items usually 1 or 2 weeks :)

bonifaido:
Releasing in 1 hour from now. 👍

PR #1415, opened 2021-09-14
charts: Fix markdown tables in chart Readmes

LinAnt:
### What's in this PR?
Fixes bad formatting in markdown tables

### Why?
The Readmes render fine in browsers etc, but looks weird when read in their raw form.,

I use this thing for linting markdown: https://github.com/DavidAnson/markdownlint
There are some other warnings, but I think tables are the only thing that makes sense fixing.

### Checklist


bonifaido (approved):
Thank you!

PR #1437, opened 2021-10-07
feat(chart): Adding initContainers support for Vault Secrets Webhook chart

fdpeiter:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | none

### What's in this PR?

A simple change in the chart to add support for init-containers.

### Why?

Having init-containers enable you to run automation tasks, validations or any other scripts prior to vault-secrets-webhook running. In my particular use case, we created an init-container that manages to create the Kubernetes auth method in Vault, so you don't have to integrate the cluster manually.

### Checklist

- [X] User guide and development docs updated (if needed)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.17.1-alpine AS builder
+FROM golang:1.17.2-alpine AS builder
 
 RUN apk add --update --no-cache build-base git mercurial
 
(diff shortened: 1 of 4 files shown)


fdpeiter:
Just updated with upstream again, seems a commit was merged to master in the meantime.

PR #1469, opened 2021-11-11
set privileged to true on vaultSA PSP

lelvisl:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
set privileged to true on vaultSA PSP

### Why?
allowPrivilegeEscalation and allowedCapabilities don't work without privileged: true

### Checklist

### DIFF
--- a/charts/vault-operator/Chart.yaml
+++ b/charts/vault-operator/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: vault-operator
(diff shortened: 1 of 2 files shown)


pbalogh-sa (approved):
You're right. Thanks.
Remark: In the future it should be configurable by values.

bonifaido (approved):

PR #1487, opened 2021-12-06
Support NonVoter option on Raft Join Request (RaftJoin) in operator. 

### LINKED ISSUE #1486: Unable to add Raft member as NonVoter (non_voter) / Request for Support / Read Only Support
There doesn't seem to be a way to tell the operator to opt-in a secondary/replica as a non-voter in the raft quorum. 

It would be ideal for multi-dc deployments to support a DC as a non-quorum member of the raft followers. Making it a read-only instance. 

It looks like since we use the `RaftJoinRequest` struct here: https://github.com/banzaicloud/bank-vaults/blob/0266c80d08838d1f2dc7d073a489687b22179ef7/internal/vault/operator_client.go#L413

This already has an param of `NonVoter`. Meaning it should be relatively simple to add something to the CRD for supporting different raft options on a per `Vault` object basis.

Example replica CR:

```
raftLeaderAddress: XYZ
raftJoinRequestOptions:
  autoJoin: false
  nonVoter: true
```

These `raftJoinOptions` can directly translate into a `RaftJoinRequest` (This could help deprecate `raftLeaderAddress` in favour of just a `raftJoinRequestOptions` in the CRD. 

I can't find a good way to automatically achieve what's necessary for a read only replica of the Vault. 

**Additional context**
Worth noting, that for enterprise licensing this `NonVoter` option does have different meanings. However, the base usage should be the same:

https://www.vaultproject.io/api/system/storage/raft#parameters

DandyDeveloper:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1486 

### What's in this PR?
This is a basic example of introducing the [non_voter](https://www.vaultproject.io/api/system/storage/raft#non_voter) flag to enable read only replicas of the Vault for the multi cluster deployment.

The changes encompass a basic addition of an environment variable and also room to expand `RaftJoin` to be init with `api.RaftJoinRequest` from the CRD itself (This has not been done as I'd want to discuss with someone first). 

### Why?

fixes #1486 - However, I think this is more of a bandaid then a full-featured fix. As the big changes could introduce deprecations. 

### Additional context
Right now, none. This is a very basic change to add a simple argument to the raft join API request.

### Checklist

### To Do

- [ ] Update the RaftJoin initialization of an `api.RaftJoinRequest` struct from the CR itself. 
- [ ] Proper Documented approach to include this (I don't know where to update this).

### DIFF
--- a/cmd/bank-vaults/unseal.go
+++ b/cmd/bank-vaults/unseal.go
@@ -18,7 +18,6 @@ import (
 	"os"
 	"time"
 
-	"github.com/hashicorp/vault/api"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
@@ -111,10 +110,7 @@ from one of the followings:
 				}
 			} else {
 				logrus.Info("joining raft cluster...")
-				if err := v.RaftJoin(
-					&api.RaftJoinRequest{
-						LeaderAPIAddr: unsealConfig.raftLeaderAddress,
-					}); err != nil {
+				if err := v.RaftJoin(unsealConfig.raftLeaderAddress); err != nil {
 					logrus.Fatalf("error joining leader vault: %s", err.Error())
 				}
 			}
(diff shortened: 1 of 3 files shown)


DandyDeveloper:
Actually, this doesn't work because it looks like our raft implementation from `bank-vaults` first tries to use the local vault instance and rejects stuff with; `vault non-voting nodes not allowed` error. Trying to work this out.

DandyDeveloper:
Closing this whilst I find a proper way to do this.

PR #1509, opened 2022-01-12
webhook: allow securityContext modification for injected containers

NeonSludge:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
This PR makes it possible to lock down the `securityContext` of containers that get injected into pods by the webhook.
This is achieved by:
1. Implementing a base `securityContext` shared by all injected containers.
2. Dropping all Linux capabilities by default and adding them back as needed.
3. Adding 4 new pod annotations recognized by the webhook that allow setting the `runAsNonRoot`, `readOnlyRootFilesystem`, `runAsUser` and `runAsGroup` parameters for injected containers.

Added pod annotations:
| Annotation                                       | Default | Explanation
| ------------------------------------------------ | ------- | -----------
| `vault.security.banzaicloud.io/run-as-non-root`  | `false` | Set this to `true` to indicate that all injected containers must run as a non-root user. 
| `vault.security.banzaicloud.io/run-as-user`      | `0`     | Set the UID for all injected containers. The default value means that no modifications will be made to the `securityContext` of injected containers.
| `vault.security.banzaicloud.io/run-as-group`     | `0`     | Set the GID for all injected containers. The default value means that no modifications will be made to the `securityContext` of injected containers.
| `vault.security.banzaicloud.io/readonly-root-fs` | `false` | Set this to `true` to indicate that all injected containers must have a read-only root filesystem. 

Changes in default behaviour:
1. All Linux capabilities are dropped by default.
2. All injected containers now start with a base `securityContext` that then gets extended as needed.

Everything else is disabled by default. This should allow existing users to keep their old webhook configuration.

### Why?
Some Kubernetes environments may use third-party policy engines (such as OPA or Kyverno) that enforce various security policies. Some of these security policies may impose additional requirements on containers or init containers such as dropping all capabilities by default or running with a read-only root FS. This may in turn make it hard to use the mutating webhook implemented in Bank Vaults because it injects containers that run with a very basic `securityContext` that can not currently be modified by the user.

### Additional context
This has been tested in a 1.23.1 Kubernetes cluster with the Kyverno policy engine in a scenario where environment variables get injected into a pod that adheres to a strict set of security policies.

### Checklist

### DIFF
--- a/pkg/webhook/config.go
+++ b/pkg/webhook/config.go
@@ -49,6 +49,10 @@ type VaultConfig struct {
 	CtCPU                       resource.Quantity
 	CtMemory                    resource.Quantity
 	PspAllowPrivilegeEscalation bool
+	RunAsNonRoot                bool
+	RunAsUser                   int64
+	RunAsGroup                  int64
+	ReadOnlyRootFilesystem      bool
 	IgnoreMissingSecrets        string
 	VaultEnvPassThrough         string
 	ConfigfilePath              string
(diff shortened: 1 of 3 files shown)


NeonSludge:
> So if I understand correctly, this will break backwards compatbility, if someone relied on the default settings for some reason previously?

Well, it shouldn't because the injected containers don't seem to need the whole default set of capabilities so they will work fine without any additional action required. Specific capabilities like `IPC_LOCK` or `SYS_PTRACE` will be added back only if the use case requires them - this behaviour has not changed.

So I don't think introducing `drop: ["ALL"]` by default will break anything unless someone has a really weird environment where they are specifically preventing all containers from dropping capabilities for some reason.

I thought about adding another annotation that would toggle this behaviour much like the `runAsNonRoot` or `readOnlRootyFilesystem` parameters but dropping all capabilities by default is widely considered a best practice so I felt that this annotation would be redundant.

bonifaido (approved):
LGMT Very nice work @NeonSludge, could you please add the annotations to the docs?

NeonSludge:
> could you please add the annotations to the docs?

Sure, I've opened a PR in the docs repository.

pbriet:
This doesn't affect consul-template injected containers.
User is still "100".

https://github.com/NeonSludge/bank-vaults/blob/f3090a18aef5019540c7b4a1b0a3485c1bad5a4f/pkg/webhook/pod.go#L661

Is it a deliberate choice?

Thanks,

juniorz:
For some reason, this changes the UID used by the `vault-agent` to `0` from `100`:

https://github.com/banzaicloud/bank-vaults/commit/63fd7a9133006514cfda515fd70907709cdd9011#diff-32802053da07acef355a6beb95cc5c4494d6d9c77ee709fb9093c7febb4bfc05L660

However, the official `vault` image runs the agent as the `vault` user (UID `100`) which causes permission errors when trying to read the `ServiceAccount` token:

```
==> Vault agent configuration:

                     Cgo: disabled
               Log Level: info
                 Version: Vault v1.10.3
             Version Sha: af866591ee60485f05d6e32dd63dde93df686dfb

2022-09-22T07:16:54.728Z [INFO]  auth.handler: starting auth handler
2022-09-22T07:16:54.731Z [INFO]  auth.handler: authenticating
2022-09-22T07:16:54.731Z [ERROR] auth.handler: error getting path or data from method: error="error reading JWT with Kubernetes Auth: open /var/run/secrets/kubernetes.io/serviceaccount/token: permission denied" backoff=1s
```

The current code also does not obey `vault.security.banzaicloud.io/run-as-user` which makes it impossible to fix this issue.

Any particular reason for this behaviour? Can we either default to UID `100` or allow it to be changed via an annotation?

Thank you

PR #1532, opened 2022-02-03
configurer: Purge Un-managed Audit Devices

wazery:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | mentioned in #1520, #1519, #1514, #1513
| License         | Apache 2.0

### What's in this PR?

This PR is part of a series of feature PRs to implement purging unmanaged resources from Vault, you can refer to the previous PR over here #1520 .

This PR handles the logic of purging unmanaged Audit devices, as well as enabling them if they are non-existent.

### Next

- Migrate Startup Secrets to the new style, it will not have a purge option because it doesn't fit it (I will work on it).
- Documentation and examples @aabouzaid will work on it.

### Checklist

- [x] Code meets the [Developer Guide](https://github.com/banzaicloud/developer-guide)

PR #1513, opened 2022-01-16
feature: Refactor auth methods handling and implement purgeUnmanagedConfig

wazery:
| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | yes|yes
| API breaks?     | no|yes
| Deprecations?   | no|yes
| Related tickets | fixes #605, mentioned in #1503 
| License         | Apache 2.0

### What's in this PR?

This PR is part of a series of implementing `purgeUnmanagedConfig` which enables for syncing Vault config with bank-vaults' external configs. As well as refactoring the cli-tool's operator_client.

### Why?

More details can be found in this first PR #605 and this issue #1503 

### Additional context

This PR is concerned with the "Auth Methods" refactoring and the implementation of `purgeUnmanagedConfig`.

#### How to configure `purgeUnmanagedConfig`?

If the user wants to remove any unmanaged config (just policies, auth methods at the moment), they just need to add `purgeUnmanagedConfig` under `externalConfig`:

```
externalConfig:
  purgeUnmanagedConfig:
    enabled: true
```

Auth methods purging can be also disabled by excluding it like so:

```
externalConfig:
  purgeUnmanagedConfig:
    enabled: true
    exclude:
      auth: true
```

### Checklist

- [x] Code meets the [Developer Guide](https://github.com/banzaicloud/developer-guide)
- [ ] User guide and development docs updated (if needed)
- [ ] Related Helm chart(s) updated (if needed)

### To Do

- [ ] If the PR is not complete but you want to discuss the approach, list what remains to be done here

PR #1535, opened 2022-02-05
configurer: refactor startup secrets

wazery:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | mentioned in #1520, #1519, #1514, #1513, #1532 

### What's in this PR?
This PR is part of a series of feature PRs to implement purging unmanaged resources from Vault, you can refer to the previous PR over here #1532.

This PR just refactors StartupSecrets to be consistent with the new structure.

### Next

- Documentation and examples @aabouzaid will work on it.

### Checklist

### DIFF
--- a/internal/vault/startup_secrets.go
+++ b/internal/vault/startup_secrets.go
@@ -1,0 +1,109 @@
+// Copyright © 2022 Banzai Cloud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package vault
+
+import (
+	"strings"
+
+	"emperror.dev/errors"
+	"github.com/spf13/cast"
+)
+
+type startupSecret struct {
+	Type string `json:"type"`
+	Path string `json:"path"`
+	Data struct {
+		Data         map[string]interface{} `json:"data"`
+		SecretKeyRef []interface{}          `json:"secretKeyRef"`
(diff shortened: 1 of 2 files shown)


internal/vault/startup_secrets.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"emperror.dev/errors"
+	"github.com/spf13/cast"
+)
+
+type startupSecret struct {
+	Type string `json:"type"`
+	Path string `json:"path"`
+	Data struct {
+		Data         map[string]interface{} `json:"data"`
```
aabouzaid, line 28: I just realized that my code here is wrong, there is no data inside data but only a special case of `secretKeyRef`.

so anything under `startupSecret.Data` should be created as secret normally unless `secretKeyRef` is present there.

I will fix it tomorrow, please don't merge till that.

aabouzaid: @bonifaido can you please confirm my observation?

so if we have this as in many examples:

```
startupSecrets:
  - type: kv
    path: secret/foo
    data:
      data:
        secret: foo
``

in the path `secret/foo` we will have a map called `{"data": {"secret": "foo"}} `.
if that is the case it's super confusing to have that as an example ... the second `data` key is actually part of the secret.

Comment (bonifaido) [reply internal/vault/startup_secrets.go:28]: If I remember correctly this used the lower level API (which has no idea about KV version 1 or 2), take this example, which would like to have the secrets in kv2 format, so you have to put manually the `/data/` section to the path and data wrapper to to the json itself:
```
    startupSecrets:
      - type: kv
        path: secret/data/accounts/aws
        data:
          data:
            AWS_ACCESS_KEY_ID: secretId
            AWS_SECRET_ACCESS_KEY: s3cr3t
```

aabouzaid: @bonifaido thanks for the clarification, I've fixed it according to that.

aabouzaid:
This PR is ready to be merged.

aabouzaid:
@bonifaido done :heavy_check_mark:

bonifaido (approved):
LGTM! Thanks for this great work for all these PRs!

PR #1562, opened 2022-03-16
webhook: Fix broken capabilities (#1531)

### LINKED ISSUE #1531: Vault Agent Templating sidecar fails to use configfile after v1.15.0 upgrade
**Describe the bug**:

After the upgrade to v1.15.0 the pod fails to start and vault-agent returns the following error:

```text
vault-agent chown: /vault/config/config.hcl: Read-only file system                                                                                                            
vault-agent chown: /vault/config: Operation not permitted                                                                                                                     
vault-agent chown: /vault/config: Operation not permitted                                                                                                                     
vault-agent Could not chown /vault/config (may not have appropriate permissions)
vault-agent unable to set CAP_SETFCAP effective capability: Operation not permitted
```

**Expected behaviour**:
We not change the underlying configuration because of the default usage and the PR description it should work.

We use the Vault Agent Templating according to the documentation in https://banzaicloud.com/docs/bank-vaults/mutating-webhook/vault-agent-templating/. For that we are setting the following annotations to our pod:

vault.security.banzaicloud.io/vault-agent-configmap: configmap-name
vault.security.banzaicloud.io/vault-configfile-path: /config

The Configmap looks something like this:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: configmap-name
  namespace: namespace
data:
  config.hcl: |
    auto_auth {
      method "kubernetes" {
        mount_path = "auth/kubernetes" 
        config = {
          role = "<role-name>" 
        }
      }
      sink "file" {
        config = {
          path = "/vault/.vault-token" 
        }
      }
    }
    template {
      left_delimiter  = "[[" 
      right_delimiter = "]]" 
      contents = <<-EOS
        SOME-CONFIG
      EOS
      destination = "/vault/secrets/pod_config.yml" 
    }
```

**Additional context**:

We currently believe that the problem has been introduced with https://github.com/banzaicloud/bank-vaults/pull/1509. We are not sure, but could it be related to setting the runAsUser to UID 100 https://github.com/banzaicloud/bank-vaults/blob/63fd7a9133006514cfda515fd70907709cdd9011/pkg/webhook/pod.go#L51? 
vault.security.banzaicloud.io/run-as-user implies to us that it should be set to 0. Unfortunately, we were not able to change the UID to root, as https://github.com/banzaicloud/bank-vaults/blob/63fd7a9133006514cfda515fd70907709cdd9011/pkg/webhook/pod.go#L832 checks > 0 and https://github.com/banzaicloud/bank-vaults/blob/63fd7a9133006514cfda515fd70907709cdd9011/pkg/webhook/pod.go#L662 overwrites it afterwards anyway.

The securityContext for the vault-agent container created by the webhook looks like this:

```yaml
securityContext:                                                                                                                                                          
       allowPrivilegeEscalation: false                                                                                                                                         
       capabilities:                                                                                                                                                           
         add:                                                                                                                                                                  
         - IPC_LOCK                                                                                                                                                            
         - SYS_PTRACE                                                                                                                                                          
         drop:                                                                                                                                                                 
         - ALL                                                                                                                                                                 
       readOnlyRootFilesystem: false                                                                                                                                           
       runAsNonRoot: false                                                                                                                                                     
       runAsUser: 100 
```
**Environment details**:

- bank-vaults version: 1.15.0

/kind bug

### LINKED ISSUE #1540: vault-secret-webhook creating vault-agent without enough permissions
Since upgrading to vault-webhook 1.15, we get the following error when trying to inject secrets using vault-agent:

(logs from the gent injected to the pod)
```
chown: /vault/config/config.hcl: Read-only file system
chown: /vault/config: Operation not permitted
Could not chown /vault/config (may not have appropriate permissions)
unable to set CAP_SETFCAP effective capability: Operation not permitted
```

**Expected behaviour**:
Inject secret

Use vault agent-config-map in the pod annotation:
```
      annotations:
        "vault.security.banzaicloud.io/vault-agent-configmap": x-vault-agent
```
create the configmap
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: x-vault-agent
  labels:
    app.kubernetes.io/name: x-vault-agent
data:
  config.hcl: |
    auto_auth {
      method "kubernetes" {
        mount_path = "auth/kubernetes"
        config = {
          role = "x"
        }
      }
      sink "file" {
        config = {
          path = "/vault/.vault-token"
        }
      }
    }
    template {
      contents = <<EOH
      {{`{{- with secret "secret/data/shared/xxx/secret" -}}
      {{ .Data.data.credentials }}{{- end -}}`}}
      EOH
      destination = "/vault/secrets/credentials.json"
    }
```

**Additional context**:

It looks like it was introduced by https://github.com/banzaicloud/bank-vaults/commit/63fd7a9133006514cfda515fd70907709cdd9011.

The "drop all" prevents the `CAP_SETFCAP`. vault-agent injected wth 1.14 do not drop capabilities.

**Environment details**:
- Kubernetes version : 1.18
- Cloud-provider/provisioner: EKS
- bank-vaults version: vault-secret-webhook 1.15

/kind bug

ernadh:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1531, fixes #1540

### What's in this PR?
* Fix for capabilities regression introduced in 1.15.0
* Additional acceptance test

### Why?
1.15.0 introduced a regression in dropping all capabilities necessary for Vault Agent Templating. The PR fixes that by keeping the capabilities drop while adding the bare minimum that is necessary for proper functionality of templating.
In addition to that, the PR adds a new acceptance test that goes through the process of templating. This test will fail the build in case same or similar regression happens in the future.

### Checklist

### DIFF
--- a/pkg/webhook/pod.go
+++ b/pkg/webhook/pod.go
@@ -48,7 +48,7 @@ auto_auth {
                 }
         }
 }`
-	vaultAgentUID int64 = 100
+	vaultAgentUID int64 = 0
 
 	VaultEnvVolumeName = "vault-env"
 )
@@ -817,8 +817,16 @@ func getBaseSecurityContext(podSecurityContext *corev1.PodSecurityContext, vault
 	context := &corev1.SecurityContext{
 		AllowPrivilegeEscalation: &vaultConfig.PspAllowPrivilegeEscalation,
 		RunAsNonRoot:             &vaultConfig.RunAsNonRoot,
+		RunAsUser:                &vaultConfig.RunAsUser,
 		ReadOnlyRootFilesystem:   &vaultConfig.ReadOnlyRootFilesystem,
 		Capabilities: &corev1.Capabilities{
+			Add: []corev1.Capability{
+				"CHOWN",
+				"SETFCAP",
+				"SETGID",
+				"SETPCAP",
+				"SETUID",
+			},
 			Drop: []corev1.Capability{
 				"ALL",
 			},
--- a/deploy/test-deploy-templating.yaml
+++ b/deploy/test-deploy-templating.yaml
@@ -1,0 +1,77 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/name: my-app
+    my-app.kubernetes.io/name: my-app-vault-agent
+    branches: "true"
+  name: my-app-vault-agent
+  namespace: default
+data:
+  config.hcl: |
+    vault {
+      // This is needed until https://github.com/hashicorp/vault/issues/7889
+      // gets fixed, otherwise it is automated by the webhook.
+      ca_cert = "/vault/tls/ca.crt"
+    }
+    auto_auth {
+      method "kubernetes" {
+        mount_path = "auth/kubernetes"
+        config = {
+          role = "default"
+        }
+      }
+      sink "file" {
+        config = {
+          path = "/vault/.vault-token"
+        }
+      }
+    }
+    template {
+      contents = <<EOH
+        {{- with secret "secret/accounts/aws" }}
+        {
+          "id": "{{ .Data.data.AWS_ACCESS_KEY_ID }}",
+          "key": "{{ .Data.data.AWS_SECRET_ACCESS_KEY }}"
+        }
+        {{ end }}
+      EOH
+      destination = "/vault/secrets/config.yaml"
+      command     = "/bin/sh -c \"kill -HUP $(pidof vault-demo-app) || true\""
+    }
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: vault-test
+  namespace: default
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: test-templating
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: test-templating
+      annotations:
+        vault.security.banzaicloud.io/vault-addr: "https://vault:8200" # optional, the address of the Vault service, default values is https://vault:8200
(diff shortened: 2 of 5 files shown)


bonifaido (approved):
LGTM!

pkg/webhook/pod.go
```diff
@@ -817,8 +817,16 @@ func getBaseSecurityContext(podSecurityContext *corev1.PodSecurityContext, vault
 	context := &corev1.SecurityContext{
 		AllowPrivilegeEscalation: &vaultConfig.PspAllowPrivilegeEscalation,
 		RunAsNonRoot:             &vaultConfig.RunAsNonRoot,
+		RunAsUser:                &vaultConfig.RunAsUser,
 		ReadOnlyRootFilesystem:   &vaultConfig.ReadOnlyRootFilesystem,
 		Capabilities: &corev1.Capabilities{
+			Add: []corev1.Capability{
+				"CHOWN",
```
madianas21, line 824: These added capabilities are causing an issue:
```
Error creating: pods "xxxxxx-service-db-migration-" is forbidden: PodSecurityPolicy: unable to admit pod: [spec.initContainers[0].securityContext.capabilities.add: Invalid value: "CHOWN": capability may not be added spec.initContainers[0].securityContext.capabilities.add: Invalid value: "SETFCAP": capability may not be added spec.initContainers[0].securityContext.capabilities.add: Invalid value: "SETGID": capability may not be added spec.initContainers[0].securityContext.capabilities.add: Invalid value: "SETPCAP": capability may not be added spec.initContainers[0].securityContext.capabilities.add: Invalid value: "SETUID": capability may not be added]
```

ernadh: Can you create an issue and fill out all the required fields to give us an opportunity to reproduce the problem using the platform you are using?

Issue #1573, opened 2022-03-28
Multiple equal signs in vaultEnvsConfig values parsed incorrectly


avoidik:
**Describe the bug**:

bank-vaults incorrectly parses configuration template with multiple equal signs in `vaultEnvsConfig` values, for example

```
param: "${ .Env.VAULT_ABC }"
```

where

```
  vaultEnvsConfig:
    - name: VAULT_ABC
      value: "abc=123 xyz=456"
```

will be rendered as

```
{"param": "abc"}
```

**Expected behaviour**:

I'm expecting it to render

```
{"param": "abc=123 xyz=456"}
```

**Steps to reproduce the bug**:

```go
package main

import (
	"fmt"
	"os"
	"strings"
)

# https://github.com/banzaicloud/bank-vaults/blob/9e797bd85a33845d6301eb1d907fb216198c99be/internal/configuration/template.go#L67-L79
func EnvTemplate() struct{ Env map[string]string } {
	var env struct {
		Env map[string]string
	}
	env.Env = make(map[string]string, len(os.Environ()))

	for _, v := range os.Environ() {
		split := strings.Split(v, "=")
		env.Env[split[0]] = split[1]
	}
	return env
}

func main() {
	os.Setenv("VAULT_ABC", `abc=123 xyz=456`)
	env := EnvTemplate()
	fmt.Printf("%+v\n", env.Env["VAULT_ABC"])
}
```

**Additional context**:

None

**Environment details**:

- Kubernetes version (e.g. v1.10.2): N/A
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc): N/A
- bank-vaults version (e.g. 0.4.17): 1.15.2
- Install method (e.g. helm or static manifests): Helm
- Logs from the misbehaving component (and any other relevant logs): N/A
- Resource definition (possibly in YAML format) that caused the issue, without sensitive data: N/A

/kind bug

PR #1574, opened 2022-03-28
properly split template variables with multiple equal signs

avoidik:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | https://github.com/banzaicloud/bank-vaults/issues/1573

### What's in this PR?

Use `SplitN` instead of `Split` to split environment variables


bonifaido (approved):
So true, Thank you!

PR #1615, opened 2022-05-10
webhook: Add .Values.secretsMutation to control whether to mutate Secrets

### LINKED ISSUE #1614: Allow users to control whether to mutate Secrets or not
We would like all secrets to only be accessible at the Vault and at the applications' environment variables. The Pod mutator achieves this quite nicely for us, even when we define a Kubernetes Secret with a vault path and consume it in the Pod.

However, if we allow the Secret mutator to take action, it will replace the vault path on the Secret's data with the actual secret value from Vault.

Sometimes this is undesired, as this sensitive data might not be encrypted at rest and K8s RBAC might not be sufficient to prevent users from reading data they should not have access to, data which would otherwise be protected via authorization in Vault.
 
I would like the chart to allow me to control this behavior by not creating a MutationWebhookConfiguration resource that takes into account Secrets.

This is the same behavior that already exists for [ConfigMap mutation](https://github.com/banzaicloud/bank-vaults/blob/main/charts/vault-secrets-webhook/templates/apiservice-webhook.yaml#L176) and [CRD mutations](https://github.com/banzaicloud/bank-vaults/blob/main/charts/vault-secrets-webhook/templates/apiservice-webhook.yaml#L241).

I haven't considered other alternatives, I think this would solve the issue quite nicely and in line with other configuration already present at the chart.

**Additional context**
N/A

LCaparelli:
In order to control if secret values should be persisted in K8s.

Sometimes this is undesired, as this sensitive data might not
be encrypted at rest and RBAC might not be sufficient to prevent
users from reading data they should not have access to.

Signed-off-by: Lucas Caparelli <lucas.caparelli@gympass.com>

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1614

### What's in this PR?
A way to control whether to register the Secrets in the MutationWebhookConfiguration.

### Additional context
I've kept the default as `true` so it doesn't break current behavior for users that do not care about this.

### Checklist

### DIFF
--- a/charts/vault-secrets-webhook/Chart.yaml
+++ b/charts/vault-secrets-webhook/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: vault-secrets-webhook
-version: 1.15.8
+version: 1.15.9
 appVersion: 1.15.2
 description: A Helm chart that deploys a mutating admission webhook that configures applications to request secrets from Vault
 icon: https://raw.githubusercontent.com/banzaicloud/bank-vaults/main/docs/images/logo/bank-vaults-logo.svg
--- a/charts/vault-secrets-webhook/README.md
+++ b/charts/vault-secrets-webhook/README.md
@@ -127,6 +127,7 @@ The following tables lists configurable parameters of the vault-secrets-webhook 
 | volumes                            | extra volume definitions                                                      | `[]`                                                     |
 | volumeMounts                       | extra volume mounts                                                           | `[]`                                                     |
 | configMapMutation                  | enable injecting values from Vault to ConfigMaps                              | `false`                                                  |
+| secretsMutation                    | enable injecting values from Vault to Secrets                                 | `true`                                                   |
 | customResourceMutations            | list of CustomResources to inject values from Vault                           | `[]`                                                     |
 | podDisruptionBudget.enabled        | enable PodDisruptionBudget                                                    | `true`                                                   |
 | podDisruptionBudget.minAvailable   | represents the number of Pods that must be available (integer or percentage)  | `1`                                                      |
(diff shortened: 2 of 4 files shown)


LCaparelli:
```
+ kubectl wait --namespace vswh --for=condition=ready --timeout=150s pods -l app.kubernetes.io/name=vault-secrets-webhook
pod/vault-secrets-webhook-6d8d999565-sd574 condition met
pod/vault-secrets-webhook-6d8d999565-x8hlm condition met
+ kubectl apply -f deploy/test-secret.yaml
Error from server (InternalError): error when creating "deploy/test-secret.yaml": Internal error occurred: failed calling webhook "secrets.vault-secrets-webhook.admission.banzaicloud.com": failed to call webhook: Post "[https://vault-secrets-webhook.vswh.svc:443/secrets?timeout=10s](https://vault-secrets-webhook.vswh.svc/secrets?timeout=10s)": dial tcp 10.96.66.68:443: connect: connection refused
```

From the acceptance tests. I'm not sure if this is a consequence of the proposed changes, looks like the webhook wasn't up since the error is connection refused (though if I'm reading this correctly we did wait for it to be ready. Strange). Either that or we're hitting the wrong port, which seems unlikely.

I'll bump the chart version to fix the error from CI/build.

bonifaido:
Don't worry, the changes do look fine! However the current test framework is a bit flaky, because of the lack of resources on GitHub actions...

LCaparelli:
Got it @bonifaido, thanks! Please let me know if something else needs changing :-)

bonifaido (approved):
Nice, thank you @LCaparelli !

PR #1646, opened 2022-07-02
Fix: use the correct path var for audit and secrets engines

### LINKED ISSUE #1612: Configurer fails to create file audit device after v1.15.1
**Describe the bug**:
After upgrading from v1.14.4 to v1.15.1, we noticed that the configurer fails with the following error message:
> adding audit device  (file)
> error configuring vault: error configuring audit devices for vault: error configuring managed audits: error enabling audit device  in vault: Error making API request.\n\nURL: PUT https://vault.vault:8200/v1/sys/audit\nCode: 405. Errors:\n\n* 1 error occurred:\n\t* unsupported operation\n\n
> Failed applying configuration file: /config/vault-configurer/vault-config.yml

**Expected behaviour**:
The configurer should be able to configure the audit device. This was working until we tried to upgrade to v1.15.1

* Use Bank Vaults >= v1.15.1
* Create a new Vault CR that contains the following external config to send logs to stdout:
```yaml
audit:
  - description: send audit logs to STDOUT
    options:
      file_path: STDOUT
    type: file
```

**Additional context**:
I had a look into the Vault logs and noticed that the request path has changed after upgrading to v1.15.1: Before the audit device was successfully created when sending a request to `sys/audit/file`. Now requests are sent to `sys/audit`, which appears to be the problem. Therefore it might have been introduced with https://github.com/banzaicloud/bank-vaults/pull/1538. When downgrading to v1.15.0 the problem does not exist.

**Environment details**:
- Kubernetes version: v1.21.9
- Cloud-provider/provisioner : EKS
- bank-vaults version: v1.15.1
- Install method: static method

/kind bug

aabouzaid:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no/
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1612

### What's in this PR?
The Audit path default value was overwritten after it was set correctly because using the wrong var to trim.

In the trim process, the var `configs[index].Path` should be used because it was set by the previous if condition.

And to avoid a such issue in the future, I've also added that case (when the "path" would have defaulted to "type") to E2E tests. 

### Checklist

### DIFF
--- a/.github/workflows/ci-config-crud-e2e.yml
+++ b/.github/workflows/ci-config-crud-e2e.yml
@@ -73,37 +73,37 @@ jobs:
             --vault-config-file "${BANK_VAULTS_CONFIG_FILE}" \
             &> "${BANK_VAULTS_LOG_FILE}" & disown
 
-      - name: Audit tests
+      - name: Test group - Audit
         if: always()
         run: |
           bash scripts/validate-config-crud/validate-config-crud.sh "audit"
 
-      - name: Auth tests
+      - name: Test group - Auth
         if: always()
         run: |
           bash scripts/validate-config-crud/validate-config-crud.sh "auth"
 
-      - name: Groups tests
+      - name: Test group - Groups
         if: always()
         run: |
           bash scripts/validate-config-crud/validate-config-crud.sh "groups"
 
-      - name: Group-aliases tests
+      - name: Test group - Group-Aliases
         if: always()
         run: |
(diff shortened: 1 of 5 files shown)


aabouzaid:
@bonifaido this is a tiny fix for the audit issue.

bonifaido (approved):
Thank you for fixin!

PR #1651, opened 2022-07-08
Parameterize mount_accessor

### LINKED ISSUE #1533: Automatically set the mount_accessor for a vault policy
When creating the vault CRD we have to provide mount accessor paths for any policies that use the the authentication method in the policy template. It would be nice to automatically retrieve this information. 

In my situation, I have multiple vault servers to setup and have to provide different values for each one.

For example:
```
externalConfig:
    policies:
      - name: read-namespace-secrets
        rules:  
            path "secret/data/namespace/{{identity.entity.aliases.auth_kubernetes_0a00d852.metadata.service_account_namespace}}/*" {
              capabilities = ["read"]
            }
    auth:
      - type: kubernetes
        path: kubernetes
        roles: 
         - name: test
            bound_service_account_names = ["*"]
            bound_service_account_namespaces = ["*"]
            policies = ["read-namespace-secrets"]
            ttl = "1h"
```
It would be nice to be able to reference the mount_accessor value in the CRD somehow. Maybe via some sort of templating parameters?

ex:
```
externalConfig:
    policies:
      - name: read-namespace-secrets
        rules:  
            path "secret/data/namespace/{{identity.entity.aliases.{{ vault.kubernetes.mount_accessor }}.metadata.service_account_namespace}}/*" {
              capabilities = ["read"]
            }
    auth:
      - type: kubernetes
        path: kubernetes
        roles: 
         - name: test
            bound_service_account_names = ["*"]
            bound_service_account_namespaces = ["*"]
            policies = ["read-namespace-secrets"]
            ttl = "1h"
```

* maybe the mount_accessor value could be stored in a confimap & then referenced via an ENV variable

**Additional context**

patoarvizu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1533

### What's in this PR?

This is a first pass at trying to automate the injection of the mount_accessor id so it can be used in [templated policies](https://www.vaultproject.io/docs/concepts/policies#templated-policies). I'm looking for early feedback on this change.

Templating directly like we do with the other functions like `blob`, `awskms`, etc., is very difficult to do because it will probably introduce a circular dependency, since fetching the mount accessor requires Vault to have already been configured. There are two non-mutually exclusive options here.

1. Instead of using the normal `${` `}` notation for functions, we use a placeholder with a specific format, for example `__accessor__kubernetes`, which would be replaced by the actual mount accessor later on. This works because that placeholder would make any policy documents still valid HCL so they would pass formatting, and since policy creation happens after the auth engine creation, bank-vaults would already have the accessor ids by the time that happens. Even if the order of operations is swapped in the future, or if there is a race condition, the worst that would happen is that a valid but incorrect policy would exist for only one reconciliation cycle, and only the very first time it's created.
2. If we wanted to keep consistency, we could create a `${ accessor ... }` function that just replaces the path with the `__accessor__` placeholder, and the replacement logic kicks in later on. This option requires the logic for option 1 either way, but the templating mechanism can stay consistent, and documented together. Someone with knowledge of the underlying placeholder replacing logic could potentially use that notation directly, but it would provide no advantage, so the risk of having to support unexpected use cases is practically null.

This PR implements option 2, but if option 1 is preferred due to simplicity or to minimize the number of changes, that can be done easily.

### Why?

We wish to be able to use Vault templated policies without requiring manual intervention, or multi-stage configuration automation. Mount accessor ids are not known in advance, so abstracting them away from the definition, and letting the bank-vaults control plane discover and inject them is ideal.

### Additional context

I've validated this in a local Kubernetes cluster. If it helps, I could create a reproduction repo and post it here.

### Checklist

### To Do

### DIFF
--- a/operator/deploy/cr-policy-with-accessor.yaml
+++ b/operator/deploy/cr-policy-with-accessor.yaml
@@ -1,0 +1,48 @@
+apiVersion: "vault.banzaicloud.com/v1alpha1"
+kind: "Vault"
+metadata:
+  name: "vault"
+spec:
+  size: 1
+  image: vault:1.10.0
+
+  # Specify the ServiceAccount where the Vault Pod and the Bank-Vaults configurer/unsealer is running
+  serviceAccount: vault
+
+  # Describe where you would like to store the Vault unseal keys and root token.
+  unsealConfig:
+    kubernetes:
+      secretNamespace: default
+
+  # A YAML representation of a final vault config file.
+  # See https://www.vaultproject.io/docs/configuration/ for more information.
+  config:
+    storage:
+      file:
+        path: "/vault/file"
+    listener:
+      tcp:
+        address: "0.0.0.0:8200"
+        # Uncommenting the following line and deleting tls_cert_file and tls_key_file disables TLS
+        # tls_disable: true
+        tls_cert_file: /vault/tls/server.crt
+        tls_key_file: /vault/tls/server.key
+
+  # See: https://banzaicloud.com/docs/bank-vaults/cli-tool/#example-external-vault-configuration
+  # The repository also contains a lot examples in the deploy/ and operator/deploy directories.
+  externalConfig:
+    policies:
+      - name: allow_secrets
+        rules: |
+          path "secret/data/{{identity.entity.aliases.${ accessor `kubernetes/` }.metadata.service_account_namespace}}/*" {
+            capabilities = ["create", "read", "update", "delete", "list"]
+          }
+    auth:
+      - type: kubernetes
+        roles:
+          # Allow every pod in the default namespace to use the secret kv store
+          - name: default
+            bound_service_account_names: default
+            bound_service_account_namespaces: default
+            policies: allow_secrets
+            ttl: 1h


patoarvizu:
Hello! I created [this](https://github.com/patoarvizu/bank-vaults-repro) reproduction repo to show the proposed feature in action, I hope it helps! cc @bonifaido

patoarvizu:
Hi! Just following up here, is there any chance someone can take a look and give feedback? Thanks! 🙏 @bonifaido @pbalogh-sa

bonifaido:
Hi @patoarvizu, sorry for the delay, all of us were having vacations, but now we are back, we will take a look at this this week with @Rasek91

patoarvizu:
Thank you!

Rasek91:
Hi @patoarvizu. sorry for the delay. The draft looks good. Thank you!

Rasek91:
In my opinion approach 2 is better. We don't have unit test for the code. We have end to end test cases: [acceptance-test.sh](https://github.com/banzaicloud/bank-vaults/blob/main/hack/acceptance-test.sh), and some of them cover the templating for example: [test-deploy-templating.yaml](https://github.com/banzaicloud/bank-vaults/blob/main/deploy/test-deploy-templating.yaml), [cr-oidc.yaml](https://github.com/banzaicloud/bank-vaults/blob/main/operator/deploy/cr-oidc.yaml).

patoarvizu:
I just pushed [PR #94](https://github.com/banzaicloud/bank-vaults-docs/pull/94) for bank-vaults-docs. I'm going to remove the draft label from this PR.

I'm not sure what to do about the failed checks though? I think one has to do with the checklist at the beginning, but I don't know if there's something I need to do about `CI / Accepttance test (pull_request)`, can you advise on what I should do, if anything?

Rasek91:
We have a problem with resources in our workflows, but I rerun and passed the tests. thank you for the great work.

Rasek91 (approved):

bonifaido (approved):
LGTM!

patoarvizu:
Thanks! There's also the corresponding documentation PR [here](https://github.com/banzaicloud/bank-vaults-docs/pull/94) that's already approved.

PR #1697, opened 2022-10-12
webhook: transit engine use batch api calls for better performance

BojanZelic:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | n/a

### What's in this PR?
This PR improves performance when using the transit engine with lots of secrets coming in from vault;

### Why?
When you have lots of secrets that use the transit engine, the webhook is really slow to respond; This is because it's making N number of API calls to vault; We can leverage batch_input to send multiple secrets at once, reducing the overhead and improving the response time of the webhook.

For example - If I have a secret with 20 Values coming from vault... previously this would take about 1.5-2 seconds for the webhook to respond; After this PR since it's only making 1 API call, it takes < 250 ms

### Additional context
I've tested it extensively with ConfigMaps, and Secrets; The code for secret mutation was simplified in order to share the logic for prefetching transit secrets;  

Mutating Custom Resources or Pods still has the old behavior

### Checklist

### DIFF
--- a/internal/injector/injector.go
+++ b/internal/injector/injector.go
@@ -37,6 +37,7 @@ type SecretRenewer interface {
 type Config struct {
 	TransitKeyID         string
 	TransitPath          string
+	TransitBatchSize     int
 	IgnoreMissingSecrets bool
 	DaemonMode           bool
 }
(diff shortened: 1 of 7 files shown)


bonifaido (approved):
Nice work, thank you!

PR #1702, opened 2022-10-25
Don't crash trying to rotate aws config/root.

### LINKED ISSUE #1481: AWS secret engine rotation
**Describe the bug**:
It's not clear how to configure the AWS secrets engine with rotation. The only doc I've found is the [example](https://github.com/banzaicloud/bank-vaults/blob/master/vault-config.yml#L279) config file. In this example, `rotate` is commented out.

When I set `rotate: true`, I get a panic at [this line](https://github.com/banzaicloud/bank-vaults/blob/master/internal/vault/operator_client.go#L1279) because `config/root` has no `name` attribute. (The example also has no `name` attribute.)

If I set `rotate: true` and also set `name: foo` then I get an error because bank-vaults tries to create an invalid config entry `aws/config/root/foo`.

**Expected behaviour**:
The AWS secrets engine should be able to rotate its credentials.

```yaml
secrets:
  - path: aws
    type: aws
    description: Ephemeral AWS access credentials.
    configuration:
      config/root:
        - access_key: AKIAfoo
          secret_key: foo
          rotate: true
      roles: []
```

**Additional context**:
With no name, the panic looks like this:
```
{"level":"info","msg":"tuning already existing mount: aws/","time":"2021-11-30T23:35:34Z"}
{"level":"info","msg":"Secret at configpath aws/config/root already exists, rotate was set so this will not be updated","time":"2021-11-30T23:35:34Z"}
panic: interface conversion: interface {} is nil, not string

goroutine 1 [running]:
github.com/banzaicloud/bank-vaults/internal/vault.(*vault).configureSecretEngines(0xc000626a50, 0xc000860360)
	/build/internal/vault/operator_client.go:1279 +0x2005
github.com/banzaicloud/bank-vaults/internal/vault.(*vault).Configure(0xc000626a50, 0x4)
	/build/internal/vault/operator_client.go:475 +0x25e
main.glob..func2.2({0x23c13d0, 0xc000626a50}, 0xc0007afcc8, 0xc000945c40, 0x0, 0xc0004d2ae0, 0xc0003304e0)
	/build/cmd/bank-vaults/configure.go:127 +0x23b
main.glob..func2(0x34573c0, {0x1ffe0e4, 0x10, 0x10})
	/build/cmd/bank-vaults/configure.go:147 +0x54c
github.com/spf13/cobra.(*Command).execute(0x34573c0, {0xc00013b400, 0x10, 0x10})
	/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:854 +0x5f8
github.com/spf13/cobra.(*Command).ExecuteC(0x3457120)
	/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:958 +0x3ad
github.com/spf13/cobra.(*Command).Execute(...)
	/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:895
main.execute()
	/build/cmd/bank-vaults/main.go:138 +0xde
main.main()
	/build/cmd/bank-vaults/main.go:278 +0x5c
```

With name set, the error looks like this:
```json
{
  "level":"error",
  "msg":"error configuring vault: error configuring secret engines for vault: error configuring aws/config/root/foo config in vault: Error making API request.\n\nURL: PUT https://vault.vault.svc.cluster.local:8200/v1/aws/config/root/foo\nCode: 404. Errors:\n\n* 1 error occurred:\n\t* unsupported path\n\n",
  "time":"2021-11-30T23:27:37Z"
}
```

**Environment details**:
- Kubernetes version (e.g. v1.10.2): v1.21.5-gke.1302
- Cloud-provider/provisioner (e.g. AKS, GKE, EKS, PKE etc): GKE
- bank-vaults version (e.g. 0.4.17): 1.14.3
- Install method (e.g. helm or static manifests): static manifests
- Logs from the misbehaving component (and any other relevant logs): See above.
- Resource definition (possibly in YAML format) that caused the issue, without sensitive data: See above.

Since `rotateSecretEngineCredentials` ignores `name` when `secretEngineType == "aws"`, it seems like a dummy value should be passed instead of trying to cast `name` to `string`.

/kind bug

cjyar:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1481

### What's in this PR?
Check for nil.

### Why?
We can't cast `name` to `string` if it's nil. And the user can't work around the problem by providing a `name` to the config, because then we try to POST the config to the wrong path.

### Checklist

### DIFF
--- a/internal/vault/secrets_engines.go
+++ b/internal/vault/secrets_engines.go
@@ -315,7 +315,7 @@ func (v *vault) addManagedSecretsEngines(managedSecretsEngines []secretEngine) e
 					}
 				}
 
-				// For secret engines where the root credentials are rotatable we don't wan't to reconfigure again
+				// For secret engines where the root credentials are rotatable we don't want to reconfigure again
 				// with the old credentials, because that would cause access denied issues. Currently these are:
 				// - AWS
 				// - Database
@@ -323,7 +323,11 @@ func (v *vault) addManagedSecretsEngines(managedSecretsEngines []secretEngine) e
 					((secretEngine.Type == "database" && configOption == "config") ||
 						(secretEngine.Type == "aws" && configOption == "config/root")) {
 					// TODO we need to find out if it was rotated or not
-					err = v.rotateSecretEngineCredentials(secretEngine.Type, secretEngine.Path, name.(string), configPath)
+					nameStr := ""
+					if name != nil {
+						nameStr = name.(string)
+					}
+					err = v.rotateSecretEngineCredentials(secretEngine.Type, secretEngine.Path, nameStr, configPath)
 					if err != nil {
 						return errors.Wrapf(err, "error rotating credentials for '%s' config in vault", configPath)
 					}


bonifaido (approved):
nice catch! thx!

PR #1712, opened 2022-11-06
webhook: Add VAULT_TOKEN support

mooncube:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | n/a

### What's in this PR?

If VAULT_TOKEN env variable is defined for the webhook pod, forward it to all additional init and sidecar containers. Do not add Vault Agent container if the variable is defined because looks like it's impossible to reuse an existing token when using Vault agent.

It's possible to inject secrets into pod, secrets, configmaps and use Consul templates to generate configuration files when explicitly specifying VAULT_TOKEN, so you bypass kubernetes authentication method if using Minikube.

### Checklist

### DIFF
--- a/pkg/webhook/config.go
+++ b/pkg/webhook/config.go
@@ -83,6 +83,7 @@ type VaultConfig struct {
 	VaultServiceAccount         string
 	ObjectNamespace             string
 	MutateProbes                bool
+	Token                       string
 }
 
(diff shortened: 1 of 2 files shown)


bonifaido:
Thank you! Please sign the CLA @mooncube.

PR #1715, opened 2022-11-10
fix: log with warning level instead of error if missing secrets are allowed

comfortablynumb:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | no

### What's in this PR?

This PR changes the log level from ERROR to WARNING when missing secrets are allowed.

### Why?

Currently, when using annotation `vault.security.banzaicloud.io/vault-ignore-missing-secrets`, an error is logged with message `path not found: %s`, but in the [documentation](https://banzaicloud.com/docs/bank-vaults/mutating-webhook/annotations/) of this annotation says: `When enabled will only log warnings when Vault secrets are missing` 

### Checklist

### DIFF
--- a/internal/injector/injector.go
+++ b/internal/injector/injector.go
@@ -269,7 +269,7 @@ func (i SecretInjector) InjectSecretsFromVault(references map[string]string, inj
 				return errors.Errorf("path not found: %s", valuePath)
 			}
 
-			i.logger.Errorf("path not found: %s", valuePath)
+			i.logger.Warnf("path not found: %s", valuePath)
 
 			continue
 		}
@@ -323,7 +323,7 @@ func (i SecretInjector) InjectSecretsFromVaultPath(paths string, inject SecretIn
 				return errors.Errorf("path not found: %s", valuePath)
 			}
 
-			i.logger.Errorln("path not found:", valuePath)
+			i.logger.Warnln("path not found:", valuePath)
 
 			continue
 		}
--- a/internal/injector/injector.go
+++ b/internal/injector/injector.go
@@ -274,9 +274,9 @@ func (i SecretInjector) InjectSecretsFromVault(references map[string]string, inj
 			}
 
 			i.logger.Warnf(
-				"We couldn't find a secret path. This is not an error since missing secrets can be ignored according to the configuration you've set (annotation: %s). Path not found: %s",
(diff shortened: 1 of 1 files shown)


internal/injector/injector.go
```diff
@@ -269,7 +273,11 @@ func (i SecretInjector) InjectSecretsFromVault(references map[string]string, inj
 				return errors.Errorf("path not found: %s", valuePath)
 			}
 
-			i.logger.Errorf("path not found: %s", valuePath)
+			i.logger.Warnf(
+				"We couldn't find a secret path. This is not an error since missing secrets can be ignored according to the configuration you've set (annotation: %s). Path not found: %s",
```
akijakya, line 277: I'd propose replacing the first sentence with the last so the message starts with the most important part followed by a human friendly explanation.

internal/injector/injector.go
```diff
@@ -323,7 +331,11 @@ func (i SecretInjector) InjectSecretsFromVaultPath(paths string, inject SecretIn
 				return errors.Errorf("path not found: %s", valuePath)
 			}
 
-			i.logger.Errorln("path not found:", valuePath)
+			i.logger.Warnf(
+				"We couldn't find a secret path. This is not an error since missing secrets can be ignored according to the configuration you've set (annotation: %s). Path not found: %s",
```
akijakya, line 335: I'd propose replacing the first sentence with the last so the message starts with the most important part followed by a human friendly explanation.

akijakya (commented):
Hi @comfortablynumb, thanks for your contribution, also sorry for the delay in response!

internal/injector/injector.go
comfortablynumb: Done!

internal/injector/injector.go
comfortablynumb: Done!

comfortablynumb:
> Hi @comfortablynumb, thanks for your contribution, also sorry for the delay in response!

Hi @akijakya! I've updated the PR with your suggestions. Let me know if you need anything else.

Thanks!

akijakya (approved):
LGTM

bonifaido (approved):

PR #1779, opened 2023-01-27
vault-env: fix vault-env generated env which may contain duplicates

NitriKx:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

I have noticed that the new `preprocessTransitSecrets` is calling `inject` even if nothing has been replaced with values from the transit cache. 

In case you are working with "maps" (ConfigMap, Secret) this does not matter as the "old" code will still overwrite the key with the secret. But in the case of `vault-env` you will have the same env var set twice and most the the time only the first occurence (without the secret) is taken into account.

### Why?

Since https://github.com/banzaicloud/bank-vaults/pull/1697 we are experimenting issues with `vault-env`: the generated environment has all the env vars containing `vault:` tokens duplicated, one version in its original version and a second one with the secret instead of the token.

```
root@pod:/# /vault/vault-env printenv
{"addr":"","app":"vault-env","level":"info","msg":"received new Vault token","path":"...","role":"...","time":"2023-01-27T15:25:12Z"}
{"app":"vault-env","level":"info","msg":"initial Vault token arrived","time":"2023-01-27T15:25:12Z"}
{"app":"vault-env","level":"info","msg":"spawning process: [printenv]","time":"2023-01-27T15:25:12Z"}
RMQ_PASSWORD=${vault:secret/data/rmqpassword#password}
OTHER=VAL
BLA=BLA
......
RMQ_PASSWORD=suPeRSecrEt
......
```

### Checklist

### DIFF
--- a/internal/injector/injector.go
+++ b/internal/injector/injector.go
@@ -143,12 +143,20 @@ func (i SecretInjector) preprocessTransitSecrets(references *map[string]string, 
 
 	for name, value := range *references {
 		if HasInlineVaultDelimiters(value) {
+			newValue := value
 			for _, vaultSecretReference := range FindInlineVaultDelimiters(value) {
 				if v, ok := i.transitCache[vaultSecretReference[0]]; ok {
-					value = strings.Replace(value, vaultSecretReference[0], string(v), -1)
+					newValue = strings.Replace(value, vaultSecretReference[0], string(v), -1)
 				}
 			}
-			inject(name, value)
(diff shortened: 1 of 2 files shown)


akijakya (approved):
LGTM, thanks for the contribution!

PR #1784, opened 2023-02-02
1777: Make service account mount configurable for vault-agent sidecar

### LINKED ISSUE #1777: Make service account mount configurable for vault-agent sidecar
I want to use JWT auth for the vault agent sidecar, my app already has a service account token it uses to authenticate with Vault.

I want the vault agent to use the same service account token.  However, looking at the code the token path is [hardcoded](https://github.com/banzaicloud/bank-vaults/blob/main/pkg/webhook/pod.go#L622) to `/var/run/secrets/kubernetes.io/serviceaccount/token`. 

Can we make this configurable using annotations e.g `vault.security.banzaicloud.io/vault-agent-token-path`.  If this is not specified then default to `/var/run/secrets/kubernetes.io/serviceaccount/token`

malachiobadeyi:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1777

### What's in this PR?
This PR make the service account mount configurable for init containers

### Why?
Allows you to specify a non-default service account token mount path for authenticating using JWT.  
#1777

### Checklist

- ~User guide and development docs updated (if needed) TODO~


akijakya (approved):
LGTM, thanks for the contribution!

Issue #1803, opened 2023-02-17
[vault-secrets-webhook]: Dynamically update TLS certificates


fran-wl:
**Is your feature request related to a problem? Please describe.**
The problem I'm facing is that when a certificate used by the vault-secrets-webhook server expires and it's renewed then I need to restart the pods for the server to load the new certificate.

**Describe the solution you'd like**
I would like for certificates to be reloaded without having to restart the vault-secrets-webhook pods.

This could be achieved by overriding the server `TLSConfig.GetCertificate` function so that it reloads the certificate every time it changes. A basic implementation of this feature could be done by changing the line [vault-secrets-webhook/main.go#L134](https://github.com/banzaicloud/bank-vaults/blob/main/cmd/vault-secrets-webhook/main.go#L134) so that the certificate is loaded on every call:
```golang
srv := &http.Server{
	Addr:    listenAddress,
	Handler: mux,
	TLSConfig: &tls.Config{
		GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
			// TODO: improve to avoid reloading the cert on every call.
			cert, err := tls.LoadX509KeyPair(tlsCertFile, tlsPrivateKeyFile)
			if err != nil {
				return nil, err
			}
			return &cert, nil
		},
	},
}
err = srv.ListenAndServeTLS("", "")
```

**Describe alternatives you've considered**
A more efficient way would be to have a routine to monitor the cert file and reload it once it changes.

bonifaido:
What changes your certificate in the given Secret? People tend to use some kind external tools for this, instead of implementing resource watching in the given application. https://github.com/stakater/Reloader is one of these tools.

Also if you renew the certificate with some CI/CD automation you can simply restart the pods (by default they run in two instances if deployed by the chart) and that causes zero downtime.

fran-wl:
Hi @bonifaido, thanks for taking a look at this. We are using cert-manager to handle changes... We thought it would make sense for this component to be able to keep working by itself without having to rely on other add-ons or do a rollout.

rsafonseca:
[...]

PR #1809, opened 2023-02-24
feature: reload vault-secret-webhook tls certicate on renewal

fran-wl:
| Q               | A
| --------------- | ---
| Bug fix?        | no/yes
| New feature?    | no/yes
| API breaks?     | no/yes
| Deprecations?   | no/yes

### Why?

When the tls certificate used by the vault-secrets-webhook server expires and it's renewed then, It's necessary to restart the pods to load the new certificate.

### Checklist

- [ ] Code meets the [Developer Guide](https://github.com/banzaicloud/developer-guide)

### To Do


universam1:
Looks like we are experiencing this issue, any plan to rollout?

fran-wl:
> Looks like we are experiencing this issue, any plan to rollout?

I think you should ask in the thread of the related [ticket](https://github.com/banzaicloud/bank-vaults/issues/1803). BTW I've been running this code in production without issues for the last 2 months.

PR #1810, opened 2023-02-28
1782: allow env to be set in vault init container

### LINKED ISSUE #1782: Vault Agent fails to start if you change the vault user name
**Describe the bug**:
If you try to change the vault user using "vault.security.banzaicloud.io/run-as-user" then the vault-agent will fail to start up with the following error message `unable to set CAP_SETFCAP effective capability: Operation not permitted`.   This happens because the vault [entrypoint.s](https://github.com/hashicorp/vault/blob/v1.12.2/.release/docker/ubi-docker-entrypoint.sh#L90) script attemps to run `setcap cap_ipc_lock=+ep $(readlink -f /bin/vault)` but fails if it's not the vault user.   This step can be bypassed by setting the env variable SKIP_SETCAP=1, however, I can not see any way to set env variables using the mutating webhook.   

Can we add a way to set ENV variables by annotating the pod spec.  e.g `vault.security.banzaicloud.io/vault-agent-env-variables: "[{\"Name\": \"SKIP_SETCAP\",\"Value\": \"1\"}]"`

malachiobadeyi:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1782 

### What's in this PR?
This feature enables environment variables to be set In the vault init container

### Checklist

### DIFF
--- a/pkg/webhook/config.go
+++ b/pkg/webhook/config.go
@@ -69,6 +69,7 @@ type VaultConfig struct {
 	AgentMemory                   resource.Quantity
 	AgentImage                    string
 	AgentImagePullPolicy          corev1.PullPolicy
+	AgentEnvVariables             string
 	ServiceAccountTokenVolumeName string
 	EnvImage                      string
 	EnvImagePullPolicy            corev1.PullPolicy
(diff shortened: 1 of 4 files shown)


akijakya:
Hi @malachiobadeyi, thanks for the contribution! Could you please sing the CLA, also rebase the PR onto the main branch so this could be merged soon? Thanks!

akijakya (approved):

Issue #954, opened 2020-04-17
[vault-secrets-webhook] Make default vault role setting configurable


javanthropus:
**Is your feature request related to a problem? Please describe.**
Our users currently need to set the vault role for their pod explicitly even though it should always be *namespace_serviceaccount* in our scenarios.  We have been able to define defaults for all of the other settings we need so that our users do not need to worry about them, but not this one since the default is simply the serviceaccount name at this time.  We want to ensure uniqueness of role names across namespaces and serviceaccounts.

**Describe the solution you'd like**
We would like to be able to define the vault role setting for a pod with a default of the form *namespace_serviceaccount*, where namespace is the namespace of the pod and serviceaccount is the serviceaccount the pod is configured to use.  Being able to supply a template for the vault role in the settings for the webhook would grant sufficient flexibility.

We suggest allowing the `VAULT_ROLE` environment variable of the webhook to be set to something like `{{ .namespace }}_{{ .serviceAccount }}` as a template would give us what we need while also allowing its default value to implicitly be `{{ .serviceAccount }}` to preserve backward compatibility.

**Describe alternatives you've considered**
We are currently using the explicit method, but this is additional toil for our users which can be overlooked or incorrectly set, leading to errors that our users must investigate.  We cannot use the current default because it is set to only the name of the serviceaccount for the pod which could be replicated in another namespace and need to remain separate.  Including the namespace in the role name allows for roles to be used that distinguish between the different serviceaccounts.

**Additional context**
None.

bonifaido:
Hi @javanthropus, good idea, I'm happy to have something like this in webhook instead of just the `{{.serviceAccount}}`. I can imagine some go templating pattern, that could be provided to the webhook to create the role name. Are you about to submit a PR for this, that could probably speed up the process?

javanthropus:
Sadly, I'm not a golang developer, so this would take a significant amount of time for me to attempt.  I can raise this issue with others at my company though and see if any of them are able to provide a PR.

csatib02:
Completed in: https://github.com/bank-vaults/bank-vaults/pull/1870

PR #1870, opened 2023-04-06
templating support for vault-role setting

### LINKED ISSUE #1160: [vault-secrets-webhook] Configurable default behavior of vault-role
Vault-secrets-webhook uses the env var `VAULT_ROLE` to provide the default role used for the Kubernetes auth method. If it is empty, it uses the ServiceAccount of the Pod. This is excellent for environments with tight security, but requires that ServiceAccounts and roles in Vault are always kept in sync. But in a bit more relaxed environments, one might want to use the Namespace of the Pod as a default role (which is currently not supported). It is true, that this can be accomplished by assigning a default ServiceAccount to all Pods, but it is troublesome to always set this in all Helm charts.

**Suggestion**
The env var `VAULT_ROLE` could accept two special values to specify what type of automatic default roles should be used. For example:

- `VAULT_ROLE="auto-serviceaccount"` (also `VAULT_ROLE=""`): uses the ServiceAccount of the Pod as the default role
- `VAULT_ROLE="auto-namespace"`: uses the Namespace of the Pod as the default role (maybe prefixed with `namespace-${namespace}`)

The logic needs to be added here:
- https://github.com/banzaicloud/bank-vaults/blob/7c55c84d06533d4273486f216d9a01adf67953cf/cmd/vault-secrets-webhook/main.go#L135-L144

universam1:
As described in the mentioned issues, there is a need to define a role based on variables such as the Namespace or ServiceAccount involved.

Instead of manually training developers to set the right annotation, we could define a default using templating to ensure the right role for the right namespace or other concepts.

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1160 #954

### What's in this PR?
We want to be able to define the vault role setting with a default of the form _"k8s.\<namespace\>"_, where namespace is the namespace of the pod. Being able to supply a template for the vault role in the settings for the webhook would grant sufficient flexibility.

Hence a very handy **default** configuration could be set so that every deployment has its namespaced role automatically:

```yaml
      containers:
      - env:
        - name: VAULT_ROLE
          value: 'k8s.{{.namespace}}'
```

Or at deployment level:
```yaml
      annotations:
        vault.security.banzaicloud.io/vault-role: '{{ .namespace }}-{{ serviceaccount }}'
```

### Why?
Our users currently need to set the vault role for their pod explicitly even though it should always be _k8s.\<namespace\>_ in our scenarios. We have been able to define defaults for all of the other settings we need so that our users do not need to worry about them, but not this one since the default is simply the serviceaccount name at this time. 

We need to ensure uniqueness of role names across namespaces and serviceaccounts!

### Checklist

For now this is a naive approach, to get feedback from the stakeholders @bonifaido how the implementation is expected.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -20,6 +20,8 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"strings"
+	"text/template"
 
 	"emperror.dev/errors"
 	vaultapi "github.com/hashicorp/vault/api"
@@ -52,6 +54,24 @@ func (mw *MutatingWebhook) VaultSecretsMutator(ctx context.Context, ar *model.Ad
 	if vaultConfig.Skip {
 		return &mutating.MutatorResult{}, nil
 	}
+
+	// parse resulting vaultConfig.Role as potential template with fields of vaultConfig
+	tmpl, err := template.New("vaultRole").Option("missingkey=error").Parse(vaultConfig.Role)
+	if err != nil {
+		return &mutating.MutatorResult{}, errors.Wrap(err, "error parsing vault_role")
+	}
+	var vRoleBuf strings.Builder
+	if err = tmpl.Execute(&vRoleBuf, map[string]string{
+		"authmethod":     vaultConfig.AuthMethod,
+		"name":           obj.GetName(),
+		"namespace":      vaultConfig.ObjectNamespace,
+		"path":           vaultConfig.Path,
+		"serviceaccount": vaultConfig.VaultServiceAccount,
+	}); err != nil {
+		return &mutating.MutatorResult{}, errors.Wrap(err, "error templating vault_role")
+	}
+	vaultConfig.Role = vRoleBuf.String()
+	mw.logger.Debugf("vaultConfig.Role = '%s'", vaultConfig.Role)
 
 	switch v := obj.(type) {
 	case *corev1.Pod:


pkg/webhook/webhook.go
```diff
@@ -17,9 +17,12 @@ package webhook
 import (
 	"context"
 	"crypto/x509"
+	"fmt"
+	"html/template"
```
AndreasSko, line 21: Did you on purpose choose `html` instead of `text`? `html/template` seems to be [mainly recommended for HTML templates](https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/html/template/doc.go;l=7-9), so we could also use `text/template`.
```suggestion
	"text/template"
```

icereed:
That looks very useful! It helps us to avoid rendering templates before we apply the YAMLs.

AndreasSko:
Thanks, that's a very handy feature! :sparkles: Looking forward to using it :slightly_smiling_face:

pkg/webhook/webhook.go
universam1: thanks

universam1:
@bonifaido Please let me know if there is anything else you like me to include in this PR, thanks!

akijakya (commented):
Hi @universam1, thanks for the contribution, also sorry for the late response!
I was wondering if this logic would fit more [here](https://github.com/banzaicloud/bank-vaults/blob/7eb7b7d0c0d1e30c533fff8f0a104dcdabf94cc7/pkg/webhook/config.go#L110) where the config is being parsed, what do you think?

universam1:
Thank you @akijakya  for the reply, sorry my late response now.

> logic would fit more [here](https://github.com/banzaicloud/bank-vaults/blob/7eb7b7d0c0d1e30c533fff8f0a104dcdabf94cc7/pkg/webhook/config.go#L110) where the config is being parsed

I thought it would be best after the config has been evaluated, but if you prefer I can move it inside the function as well

akijakya:
@universam1, looking at it again you are right about leaving this out from the config parsing function. If there will be another annotation that can be templated these could be moved to a separate function but until then it can stay here I think.
Could you please also open a PR to the docs repo and add some words about this new feature [here](https://github.com/bank-vaults/bank-vaults.dev/blob/302cc5a791be0bb94364e1c30839862990ddca6f/content/docs/mutating-webhook/annotations.md?plain=1#L13)? Thanks a lot! 🙂

akijakya (approved):
Thanks for the contribution!

sagikazarmark (approved):

PR #1808, opened 2023-02-23
fix: Add missing parameters when using auto unseal

lusu007:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1787 
| License         | Apache 2.0

### What's in this PR?

This PR checks the seal type and uses the according parameters in the init API call. Furthermore, I updated the vault go api.

### Additional context

This PR is the successor of #1795. I had to create a new one because GitHub force-pushed to my fork and automatically closed the PR when syncing the branch. Sorry for the confusion.

### Checklist

- [x] Code meets the [Developer Guide](https://github.com/banzaicloud/developer-guide)
- [x] User guide and development docs updated (if needed)
- [x] Related Helm chart(s) updated (if needed)

### To Do

- [x]  Test it with all unseal methods (tested with shamir)

PR #1871, opened 2023-04-10
fix: Honor vault operator init parameter validation

kirilvasilev:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1808 

### What's in this PR?
This PR checks the seal recovery key support and uses the according parameters in the init API call.

### Why?
This PR fixes the initialization of vaults that use a seal type that supports only recovery keys and vice versa.

### Additional context
Vault 1.12.0 introduced parameter validation in the operator init request. In the cases where the unseal type does not support recovery keys and recovery keys are provided, the initialization will fail. The same case applies to unseal types that do support recovery keys and secret keys are provided. The related Vault change can be found [here](https://github.com/hashicorp/vault/blob/main/http/sys_init.go#L170).

### Checklist

### To Do
- [x] Test it with all unseal methods (tested with shamir and awskms)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,6 @@ require (
 	github.com/google/go-containerregistry v0.8.0
 	github.com/google/go-containerregistry/pkg/authn/k8schain v0.0.0-20220105220605-d9bfbcb99e52
 	github.com/hashicorp/go-uuid v1.0.2
-	github.com/hashicorp/go-version v1.6.0
 	github.com/hashicorp/hcl v1.0.0
 	github.com/hashicorp/vault/api v1.8.2
 	github.com/hashicorp/vault/sdk v0.6.0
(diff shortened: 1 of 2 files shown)


ats0stv:
Thanks for this PR @kirilvasilev

akijakya:
Hi @kirilvasilev, thanks for the contribution! Please run `go mod tidy` (again) because there seems to be a problem, and locally that fixed the issue for me, thanks!

kirilvasilev:
Hi @akijakya, thanks for letting me know, I pushed the updated `go.mod` and `go.sum` files.

akijakya (approved):
Thanks for the contribution!

PR #1884, opened 2023-04-24
Update sdk to use the code from the new module

sagikazarmark:
This PR removes most of the `pkg/sdk` code in a backward compatible way, relying on the new github.com/bank-vaults/vault-dsk module under the hood.


akijakya (approved):
Thanks! 💪

PR #1886, opened 2023-05-02
Secret syncer basics

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Implementing the basics for a new operator working together with the webhook, which - if enabled - will periodically check if the secrets used by a resource that has the `alpha.vault.security.banzaicloud.io/reload-on-secret-change` annotation set to `true` are still the same version as in Vault, and if not, updates the resource ensuring it has the the latest version of the secrets.

This feature will support only deployments first. Implemented in this PR:
- a new logic in the webhook that adds a new `alpha.vault.security.banzaicloud.io/secret-version-hash` annotation with the hashed value of the path and current version of the Vault secrets to the deployment
- a temporary CLI tool that accepts the name (and namespace) of a deployment, and collects the used Vault secrets again, generates the hash, compares it to the one in the annotation, and if they are not matching, updates the annotation which causes the pods to restart, and the webhook doing its job again, but now with the latest version of the secrets. The logic implemented here will be the part of an operator later.

The new `collector` package can find Vault secret paths to collect that are specified:
- as `env` in the pod template in the following format: `vault:secret/data/mysql#MYSQL_PASSWORD`
- in a secret or configmap through `valueFrom` or `envFrom` in the following format: `vault:secret/data/mysql#MYSQL_PASSWORD`. It also uses the contents of the `kubectl.kubernetes.io/last-applied-configuration` annotation, because if the secret or configmap mutation is also enabled in the webhook, the value of these Vault secrets already being injected into them so the original Vault secret path can only be found there. Also if the secret change affected a secret stored in a k8s secret or configmap, that also needs to be re-applied which achieved by patching them with the contents of the `last-applied-configuration` annotation.
- in the `vault.security.banzaicloud.io/vault-env-from-path` annotation in the pod template
- in vault templates from configmaps specified in the `vault.security.banzaicloud.io/vault-agent-configmap` annotation in the pod template

### Why?

To make the first steps towards a new operator that keeps resources mutated by the Bank-Vaults webhook up to date with Vault.

### Additional context

Steps for local testing:
1. Prepare a kubernetes cluster (recommended with Docker desktop or kind)
2. Start the operator with `make operator-up`
3. Create the namespace `vswh` (`kubectl create namespace vswh`)
4. Create a Vault instance with `kubectl apply -f operator/deploy/cr.yaml`
5. Build a local version of the webhook: `docker build . -t ghcr.io/banzaicloud/vault-secrets-webhook:local -f Dockerfile.webhook`
6. Install the webhook:
```
helm upgrade --install vault-secrets-webhook charts/vault-secrets-webhook \
    --set replicaCount=1 \
    --set image.tag=local \
    --set image.pullPolicy=IfNotPresent \
    --set configMapMutation=true \
    --set deploymentSecretSync=true \
    --set configmapFailurePolicy=Fail \
    --set podsFailurePolicy=Fail \
    --set secretsFailurePolicy=Fail \
    --set vaultEnv.tag=latest \
    --namespace vswh
```
7. Apply the test deployment: `kubectl apply -f test/deploy/test-deployment-sync.yaml`
8. Check if the annotation is present in the deployment: `kubectl get deploy hello-secrets-sync -o jsonpath='{ .metadata.annotations.alpha\.vault\.security\.banzaicloud\.io/secret-version-hash }'`
9. Start port-forwarding Vault: `kubectl port-forward vault-0 8200:8200` and in another terminal instance, export some important envs that the CLI tool needs:
```
export VAULT_TOKEN=$(kubectl get secrets vault-unseal-keys -o jsonpath={.data.vault-root} | base64 --decode)

kubectl get secret vault-tls -o jsonpath="{.data.ca\.crt}" | base64 --decode > $PWD/vault-ca.crt
export VAULT_CACERT=$PWD/vault-ca.crt

export VAULT_ADDR=https://127.0.0.1:8200
```
10. Run the CLI tool: `go run cmd/vault-secrets-syncer/main.go deploy hello-secrets-sync`. The last log should say `Secrets are up to date`
11. Now change a secret in Vault in the same terminal window: `vault kv patch secret/mysql MYSQL_PASSWORD=updated3xtr3ms3cr3t`
12. Run the CLI tool again and observe in its logs and also with `kubectl get deploy hello-secrets-sync -o jsonpath='{ .metadata.annotations.alpha\.vault\.security\.banzaicloud\.io/secret-version-hash }'` that the annotation is indeed has been changed and the pods restarted, and in the pod logs the new secret has been echoed.

### Checklist

- ~Related Helm chart(s) updated (if needed)~

### DIFF
--- a/charts/vault-secrets-webhook/Chart.yaml
+++ b/charts/vault-secrets-webhook/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: vault-secrets-webhook
-version: 1.19.0
+version: 1.19.1
 appVersion: 1.19.0
 description: A Helm chart that deploys a mutating admission webhook that configures applications to request secrets from Vault
 icon: https://raw.githubusercontent.com/banzaicloud/bank-vaults/main/docs/images/logo/bank-vaults-logo.svg
--- a/charts/vault-secrets-webhook/templates/apiservice-webhook.yaml
+++ b/charts/vault-secrets-webhook/templates/apiservice-webhook.yaml
@@ -365,3 +365,64 @@ webhooks:
   sideEffects: {{ .Values.apiSideEffectValue }}
 {{- end }}
 {{- end }}
+{{- if .Values.deploymentSecretSync }}
+- name: sync.{{ template "vault-secrets-webhook.name" . }}.admission.banzaicloud.com
+  {{- if semverCompare ">=1.14-0" (include "vault-secrets-webhook.capabilities.kubeVersion" .) }}
+  admissionReviewVersions: ["v1beta1"]
+  {{- if .Values.timeoutSeconds }}
+  timeoutSeconds: {{ .Values.timeoutSeconds }}
+  {{- end }}
+  {{- end }}
+  clientConfig:
+    {{- if .Values.webhookClientConfig.useUrl }}
+    url: {{ .Values.webhookClientConfig.url }}
+    {{- else }}
+    service:
+      namespace: {{ .Release.Namespace }}
+      name: {{ template "vault-secrets-webhook.fullname" . }}
+      path: /deployment-secret-sync
+    {{- end }}
+    caBundle: {{ $caCrt }}
+  rules:
+  - operations:
+      - CREATE
+      - UPDATE
+    apiGroups:
+      - "*"
+    apiVersions:
+      - "*"
+    resources:
+      - deployments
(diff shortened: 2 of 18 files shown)


akijakya:
This POC is now in a draft PR in its own repo: https://github.com/bank-vaults/vault-secrets-reloader/pull/2

PR #1903, opened 2023-05-12
configurer: use `azure-sdk-for-go` package

### LINKED ISSUE #2041: Fix MSI not available error in bank-vaults configurer
There was a change in a version we upgraded to of the azure `go-autorest` package which has also been deprecated from which this error message is originates, switching to the new `azure-sdk-for-go` package should solve the problem but needs testing.

https://github.com/banzaicloud/bank-vaults/pull/1903

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #2041

### What's in this PR?

Use `azure-sdk-for-go` instead of deprecated `go-autorest` package for creating a client for Azure key vault (also fixes the error message `MSI not available`).

### Why?

A [version bump of the Azure auth package](https://github.com/banzaicloud/bank-vaults/pull/1887) causes failure when starting the bank-vaults image on a local kind cluster e.g. when running the acceptance tests locally. This new error was introduced [here](https://github.com/Azure/go-autorest/pull/670). Kind cluster on Github runners and also local cluster provided for Docker desktop for Mac seems not to be affected.

### Checklist

- ~Related Helm chart(s) updated (if needed)~

### To Do
- [X] Test against Azure


pkg/kv/azurekv/iam.go
```diff
@@ -34,6 +34,12 @@ func init() {
 		keyvaultAuthorizer, err = auth.NewAuthorizerFromFileWithResource(resource)
 	} else {
 		keyvaultAuthorizer, err = auth.NewAuthorizerFromEnvironmentWithResource(resource)
+		if err.Error() == "MSI not available" {
```
sagikazarmark, line 37: Matching errors like this is not really elegant. Please figure out a better way to match the error (the Azure library ought to provide a solution for this).

Also, I wonder if a recent change caused this, maybe we need to do something differently here.

Tombar:
We are running into this same issue, the steps to reproduce are straightforward 

```
docker run --rm -it ghcr.io/banzaicloud/bank-vaults:latest
2023/06/23 19:03:19 MSI not available

docker run --rm -it ghcr.io/banzaicloud/bank-vaults:1.18.0
This is a CLI tool to help automate the setup and management of Hashicorp Vault.

Usage:
  bank-vaults [command]
...
```

Is there a plan to fix this or should we stick to 1.18.0 ?

akijakya:
> Is there a plan to fix this or should we stick to 1.18.0 ?

I am currently looking into this problem, so we definitely plan to fix it, however this code will also be moved under the new `bank-vaults` organisation and a new release is expected only when that is done, so if it is not problem, please stick with the older version in the meantime, thanks!

pkg/kv/azurekv/iam.go
akijakya: Looking again into the problem, the fix would be to rewrite this part entirely using the newer `azure-sdk-for-go` package.

pkg/kv/azurekv/keyvault.go
```diff
@@ -42,19 +44,39 @@ func New(name string) (kv.Service, error) {
...
-	keyClient.Authorizer = GetKeyvaultAuthorizer()
+	// File based auth is not supported in the new SDK, hence this workaround
+	if _, ok := os.LookupEnv("AZURE_AUTH_LOCATION"); ok {
+		settings, err := auth.GetSettingsFromFile()
+		if err != nil {
+			log.Fatal(err)
+		}
+		os.Setenv(auth.TenantID, settings.Values[auth.TenantID])
+		os.Setenv(auth.ClientID, settings.Values[auth.ClientID])
+		os.Setenv(auth.ClientSecret, settings.Values[auth.ClientSecret])
```
sagikazarmark, line 55: I guess this is OK for now, because I don't want to block this PR anymore. I've opened #2056 describing a slightly better solution.

sagikazarmark (approved):

PR #1941, opened 2023-06-08
Remove vault-env (and some operator leftovers)

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | yes
| Related tickets | -

### What's in this PR?

- removed some more operator-related stuff
- removed vault-env, set the location of the new image in webhook chart

### Why?

Both of these got their own repos under the [bank-vaults org](https://github.com/bank-vaults)

### Checklist


sagikazarmark (changes requested):
Let's keep internal in this repo for now. We can figure out what to do with the duplicate code later.

charts/vault-secrets-webhook/values.yaml
```diff
@@ -48,7 +48,7 @@ webhookClientConfig:
   # url: https://example.com
 
 vaultEnv:
-  repository: ghcr.io/banzaicloud/vault-env
+  repository: ghcr.io/bank-vaults/vault-env
```
sagikazarmark, line 51: This is already done: https://github.com/banzaicloud/bank-vaults/blob/main/charts/vault-secrets-webhook/values.yaml#L51

akijakya: ok

sagikazarmark (approved):

Issue #1990, opened 2023-06-28
Startup secrets keys created in lowercase


akijakya:
> Bugs should be filed for issues encountered whilst operating bank-vaults.
> You should first attempt to resolve your issues through the community support
> channels, e.g. Slack, in order to rule out individual configuration errors.
> Please provide as much detail as possible.

**Describe the bug**:
The latest version of bank-vaults creates startup secrets, but they appear in lowercase in Vault resulting in vault-env not finding them when applying our default resources we use for testing.

**Expected behaviour**:
Getting this:
```
vault kv get secret/accounts/aws
Handling connection for 8200
====== Secret Path ======
secret/data/accounts/aws

======= Metadata =======
Key                Value
---                -----
created_time       2023-06-28T09:21:21.574452303Z
custom_metadata    <nil>
deletion_time      n/a
destroyed          false
version            1

============ Data ============
Key                      Value
---                      -----
AWS_ACCESS_KEY_ID        secretId
AWS_SECRET_ACCESS_KEY    s3cr3t
```

instead of this:

```
vault kv get secret/accounts/aws
Handling connection for 8200
====== Secret Path ======
secret/data/accounts/aws

======= Metadata =======
Key                Value
---                -----
created_time       2023-06-28T08:48:28.447984417Z
custom_metadata    <nil>
deletion_time      n/a
destroyed          false
version            1

============ Data ============
Key                      Value
---                      -----
aws_access_key_id        secretId
aws_secret_access_key    s3cr3t
```

/kind bug

akijakya:
The issue was introduced by a [viper version upgrade](https://github.com/banzaicloud/bank-vaults/pull/1965), turns out it was a bug that it allowed uppercase characters in the first place (see [issue](https://github.com/spf13/viper/issues/1486)). Short term solution is to revert the upgrade.

PR #1995, opened 2023-06-29
ci: dependabot to ignore viper and gocloud.dev

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | #1990

### What's in this PR?

Dependabot to ignore upgrading viper and gocloud.dev until we figure out how to solve the related lowercasing issue.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -7,6 +7,9 @@ updates:
       - "area/dependencies"
     schedule:
       interval: "daily"
+    ignore:
+      - dependency-name: "github.com/spf13/viper"
+      - dependency-name: "gocloud.dev"
 
   - package-ecosystem: "docker"
     directory: "/"


sagikazarmark:
There is no need for this. Just close the PRs for now. The next version will contain a fix.

It would be great however to have some sort of test that catches the problem in case we introduce it again in the future.

akijakya:
> It would be great however to have some sort of test that catches the problem in case we introduce it again in the future.

The old e2e test would have caught this, the webhook test would fail because that used the current version of everything. 
I think the `validate-config-crud` e2e test could be extended to check the lowercasing issue, also the usage of v1 vs v2 kv engines with startup secrets, I am going to create a todo item for this.

PR #2071, opened 2023-07-26
feat: add back softhsm as a separate stage

### LINKED ISSUE #2070: Missing softhsm
Put sofhsm back to the image

sagikazarmark:
Fixes #2070

### DIFF
--- a/.github/workflows/artifacts.yaml
+++ b/.github/workflows/artifacts.yaml
@@ -35,6 +35,10 @@ jobs:
     name: Container image
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        variant: ["", "softhsm"]
+
     permissions:
       contents: read
       packages: write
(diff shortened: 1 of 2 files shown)


ramizpolic:
The CI uploads and overwrites the same artifacts for both regular and softhsm steps, which is not the expected behaviour.

sagikazarmark:
@ramizpolic Can you elaborate? I don't see how that's happening.

They push different tags: https://github.com/bank-vaults/bank-vaults/actions/runs/5666983993/job/15354810683?pr=2071#step:7:71

SoftHSM variants receive a -softhsm suffix.

sagikazarmark:
Oh, you mean build artifacts?

https://github.com/bank-vaults/bank-vaults/actions/runs/5666983993

ramizpolic:
The artifacts found here https://github.com/bank-vaults/bank-vaults/actions/runs/5666983993 belong to which image? From my understanding, they are overwritten by whichever action (either upload artifact for "" or "softhsm") finishes the last.

sagikazarmark:
That's a fair point. We probably don't need artifacts for variants at this point, so I'll just disable uploading those for now.

Good catch!

sagikazarmark:
We may add the variant to the artifact names later if it becomes necessary to publish them.

Check again please!

ramizpolic (approved):

akijakya:
I built the image and ran the acceptance test locally, which still failed, but this time with this error message:

```
time="2023-07-26T10:59:34Z" level=info msg="HSM Information {CryptokiVersion:{Major:2 Minor:40} ManufacturerID:SoftHSM Flags:0 LibraryDescription:Implementation of PKCS11 LibraryVersion:{Major:2 Minor:6}}" 
time="2023-07-26T10:59:34Z" level=info msg="HSM Searching for slot in HSM slots [{ctx: 0x40006640a8 id:0}]" 
{"level":"fatal","msg":"error creating kv store: error creating HSM kv store: can't find HSM slot","time":"2023-07-26T10:59:34Z"}
```

sagikazarmark:
@akijakya note that I changed the pin from "banzai" to "bank-vaults", so we will have to change those in the CR as well.

akijakya:
> @akijakya note that I changed the pin from "banzai" to "bank-vaults", so we will have to change those in the CR as well.

Yep I just realised that the pin changed, let me update the CR and run it again!

akijakya:
> > @akijakya note that I changed the pin from "banzai" to "bank-vaults", so we will have to change those in the CR as well.
> 
> Yep I just realised that the pin changed, let me update the CR and run it again!

Still no luck with this BTW, the error message is the same despite using the new pin, also tried to specify the `slotId` field in the CR and looked at the old Dockerfile as well, whether we left something crucial out, but still have no idea unfortunately...

PR #2086, opened 2023-08-01
feat: implement a file provider for Azure auth

### LINKED ISSUE #2056: Implement a file provider for Azure auth
#1903 replaces the previous Azure auth library with a new one. In the new library, authentication changed slightly (namely file based auth is no longer available).

The current workaround is to read the file and set some details as env vars.

While this works for now, it's not an ideal solution: normally, when a credential provider fails, it's supposed to fall back to the next one in the chain. In this case, however, that's not possible, because we overwrite environment variables.

A better solution would be implementing a custom credential provider and adding it to the chain.

Some details about how we can implement such provider:

- https://github.com/Azure/azure-sdk-for-go/blob/sdk/azidentity/v1.3.0/sdk/azidentity/default_azure_credential.go#L55
- https://github.com/Azure/azure-sdk-for-go/blob/sdk/azidentity/v1.3.0/sdk/azidentity/environment_credential.go
- https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#DefaultAzureCredential

akijakya:
Fixes #2056

Tested with Azure


pkg/kv/azurekv/keyvault.go
```diff
@@ -44,21 +50,25 @@ func New(name string) (kv.Service, error) {
...
+	if defaultAzureCredErr == nil {
+		_, getTokenErr = cred.GetToken(context.Background(), policy.TokenRequestOptions{})
+	}
+	// If getting credential with the default options fails, try it with the use of auth file
+	if defaultAzureCredErr != nil || getTokenErr != nil {
+		fbc, err := NewFileBasedCredential()
 		if err != nil {
-			log.Fatal(err)
+			log.Fatalf("failed to obtain a credential: %v", errors.Combine(defaultAzureCredErr, getTokenErr, err))
 		}
```
ramizpolic, line 70: This looks cleaner
```go
	// Creating the default credential chain or auth file
	cred, err := azidentity.NewDefaultAzureCredential(nil)
	if err != nil {
		log.Debugf("falling back to creating auth file due to err=%v", err)
		if cred, err = NewFileBasedCredential(); err != nil {
			log.Fatalf("failed to create cred chain: %v", err)
		}
	}

	// Get credential
	_, err = cred.GetToken(context.Background(), policy.TokenRequestOptions{})
	if err != nil {
		log.Fatalf("failed to obtain a credential: %v", err)
	}

```

pkg/kv/azurekv/keyvault.go
```diff
@@ -93,3 +103,63 @@ func (a *azureKeyVault) Set(key string, val []byte) error {
 	_, err := a.client.SetSecret(context.Background(), key, parameters, nil)
 	return errors.Wrapf(err, "failed to set key: %s", key)
 }
+
+func NewFileBasedCredential() (azcore.TokenCredential, error) {
+	// Implementation based on github.com/Azure/go-autorest/autorest/azure/auth.GetSettingsFromFile()
+	fileLocation := os.Getenv("AZURE_AUTH_LOCATION")
```
ramizpolic, line 109: Can we have this env variable defined as constant?

pkg/kv/azurekv/keyvault.go
```diff
@@ -93,3 +103,63 @@ func (a *azureKeyVault) Set(key string, val []byte) error {
...
+	decoded, err := decode(contents)
+	if err != nil {
+		return nil, err
+	}
+
+	type azureAuthFile struct {
+		TenantID     string `json:"tenantId"`
+		ClientID     string `json:"clientId"`
+		ClientSecret string `json:"clientSecret"`
+	}
```
ramizpolic, line 129: This can be a package-defined type named e.g. `AuthConfig`

pkg/kv/azurekv/keyvault.go
```diff
@@ -93,3 +103,63 @@ func (a *azureKeyVault) Set(key string, val []byte) error {
...
 }
+
+func NewFileBasedCredential() (azcore.TokenCredential, error) {
+	// Implementation based on github.com/Azure/go-autorest/autorest/azure/auth.GetSettingsFromFile()
+	fileLocation := os.Getenv("AZURE_AUTH_LOCATION")
+	if fileLocation == "" {
+		return nil, errors.New("environment variable AZURE_AUTH_LOCATION is not set")
+	}
+
+	contents, err := ioutil.ReadFile(fileLocation)
```
ramizpolic, line 114: use `os.ReadFile` instead of deprecated ioutil package

ramizpolic (changes requested):

pkg/kv/azurekv/keyvault.go
akijakya: The problem was that `NewDefaultAzureCredential` only creates the chain, normally it wouldn't error, but the `GetToken` method can still fail. Looking at the problem again I came up with a better solution though!

pkg/kv/azurekv/keyvault.go
akijakya: These mostly came from the old Azure code I copied blindly BTW, nice catch!

pkg/kv/azurekv/keyvault.go
```diff
@@ -93,3 +96,110 @@ func (a *azureKeyVault) Set(key string, val []byte) error {
 	_, err := a.client.SetSecret(context.Background(), key, parameters, nil)
 	return errors.Wrapf(err, "failed to set key: %s", key)
 }
+
+type DefaultWithFileBasedCredential struct {
+	defaultCred   azcore.TokenCredential
+	fileBasedCred azcore.TokenCredential
```
ramizpolic, line 102: An idea: If you use `collection map[string]azcore.TokenCredential{"file": file, "default": default}` instead of specific fields, then you can have reusable logic in GetToken (you can loop over this map). This would also rename the `DefaultWithFileBasedCredential` to `MultiTokenCredential`.

Apart from that, looks good.

ramizpolic (approved):

ramizpolic (approved):

PR #2105, opened 2023-08-08
chore: remove viper from vault file reloaders

### LINKED ISSUE #2032: Consider dropping Viper
Due to #1990 we had to revert to an older Viper version.

After a brief look at the code, I'm not entirely sure we need Viper.

We only load files with Viper and use it's watch feature to some extent, but those can probably be replaced by just fsnotify.

If we really don't use any other configuration source (flag, env var, etc), I vote for dropping Viper from BV at this time.

ramizpolic:
## Overview

Fixes #2032

### DIFF
--- a/cmd/bank-vaults/configure.go
+++ b/cmd/bank-vaults/configure.go
@@ -238,7 +238,7 @@ func parseConfiguration(parser multiparser.Parser, vaultConfigFile string) *conf
 
 	// Load raw data into map
 	var data map[string]interface{}
-	if err := parser.Unmarshal(buffer.Bytes(), &data); err != nil {
+	if err := parser.Parse(buffer.Bytes(), &data); err != nil {
 		logrus.Fatalf("error parsing vault config file: %v", err)
 	}
 
--- a/cmd/bank-vaults/configure.go
+++ b/cmd/bank-vaults/configure.go
@@ -16,7 +16,8 @@ package main
 
 import (
 	"github.com/bank-vaults/bank-vaults/internal/configuration"
-	"gopkg.in/yaml.v3"
+	"github.com/ramizpolic/multiparser"
+	"github.com/ramizpolic/multiparser/parser"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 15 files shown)


sagikazarmark (changes requested):
@ramizpolic I'm not quite sure I understand the JSON file loading part. Can you explain? Did we support that previously?

cmd/bank-vaults/configure.go
```diff
@@ -15,21 +15,21 @@
 package main
 
 import (
+	"encoding/json"
+	"github.com/ghodss/yaml"
 	"io/ioutil"
```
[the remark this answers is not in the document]
ramizpolic: Same question on my end, @sagikazarmark. It appeared to me that since we used Viper, we supported YAML/TOML/JSON. Do we only support YAML config files?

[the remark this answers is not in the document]
ramizpolic: Note, I have reverted to YAML-only vault config file support.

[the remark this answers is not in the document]
ramizpolic: Working on adding JSON support as well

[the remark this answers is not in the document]
ramizpolic: Added support for both JSON and YAML

cmd/bank-vaults/configure.go
```diff
@@ -15,21 +15,20 @@
...
+	internalVault "github.com/bank-vaults/bank-vaults/internal/vault"
 	"github.com/bank-vaults/vault-sdk/vault"
 	"github.com/fsnotify/fsnotify"
 	"github.com/jpillora/backoff"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
-
-	"github.com/bank-vaults/bank-vaults/internal/configuration"
-	internalVault "github.com/bank-vaults/bank-vaults/internal/vault"
```
sagikazarmark, line 32: Proper import formatting:

- stdlib
- third-party
- module

internal/vault/operator_client.go
```diff
@@ -18,22 +18,45 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
-	"io/ioutil"
+	"github.com/mitchellh/mapstructure"
```
sagikazarmark, line 21: Same as above

cmd/bank-vaults/configure.go
sagikazarmark: golangci-lint run --fix should fix it, no? How is your IDE setup?

sagikazarmark (changes requested):

cmd/bank-vaults/configure.go
ramizpolic: Ran lint, didn't complain. Using Goland.

cmd/bank-vaults/configure.go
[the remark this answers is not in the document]
ramizpolic: Marked resolved from meeting notes

cmd/bank-vaults/configure.go
ramizpolic: I manually updated it, linting still didn't help.

internal/vault/operator_client.go
ramizpolic: Same as above

ramizpolic: Same notes like above, linting didn't help, manual did the trick, duh.

sagikazarmark (approved):

PR #2247, opened 2023-10-04
feat: add docs generator

ramizpolic:
## Overview

Adds `make gen-docs` for CLI docks generation.

## Notes for reviewer

`cmd/` packages have been split based on usage (`cmd` -> the actual CLI, and `hack` -> CLI extras/tools).


akijakya (approved):
LGTM, but do we need both a `scripts` and a `hack` directory? I'd personally only keep one for development-related scripts and code.

PR #2270, opened 2023-10-17
dev: add gen-docs command

ramizpolic:
## Overview

Enables generating CLI docs via `make gen-docs`.

### DIFF
--- a/.licensei.toml
+++ b/.licensei.toml
@@ -34,7 +34,7 @@ ignored = [
 ]
 
 [header]
-authors = ["Banzai Cloud"]
+authors = ["Banzai Cloud", "Bank-Vaults Maintainers"]
 ignorePaths = [".direnv", ".devenv", "vendor"]
 ignoreFiles = ["zz_generated.*.go"]
 template = """// Copyright © :YEAR: :AUTHOR:
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,11 @@ build: ## Build binary
 	@mkdir -p build
 	go build -race -o build/ ./cmd/bank-vaults
 
+.PHONY: gen-docs
+gen-docs: ## Generate CLI documentation
+	@mkdir -p "build/docs"
+	go run -tags=gen_docs ./cmd/bank-vaults gen-docs "build/docs"
+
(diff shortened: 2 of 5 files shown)


.licensei.toml
```diff
@@ -34,7 +34,7 @@ ignored = [
 ]
 
 [header]
-authors = ["Banzai Cloud"]
+authors = ["Banzai Cloud", "Bank-Vaults"]
```
akijakya, line 37: We usually refer to the org as authors as "Bank-Vaults Maintainers"!

cmd/bank-vaults/gen.go
```diff
@@ -0,0 +1,68 @@
+// Copyright © 2023 Bank-Vaults
```
akijakya, line 1: Same here...

cmd/bank-vaults/gen.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+var genDocsCmd = &cobra.Command{
+	Use:    "gen-docs",
+	Hidden: true,
+	Short:  "Generate documentation",
+	Args:   cobra.ExactArgs(1), // requires output folder path
+	RunE: func(cmd *cobra.Command, args []string) error {
+		outputDir := args[0]
+
```
akijakya, line 37: Right now if the output dir does not exist yet, `make gen-docs` errors, I don't know if cobra can create the target dir with using the correct settings, if not, something like this can be used:

```
		if _, err := os.Stat(outputDir); errors.Is(err, os.ErrNotExist) {
			err := os.Mkdir(outputDir, os.ModePerm)
			if err != nil {
				logrus.Error(err)
			}
		}
```

akijakya (changes requested):

cmd/bank-vaults/gen.go
ramizpolic: Yup, nice catch, forgot to add dir in Makefile

akijakya: Even better :)

akijakya (approved):

PR #2506, opened 2024-02-14
fix: Use correct unseal key name in cofigurer

kirilvasilev:
## Overview
This PR fixes a bug in the configurer in cases where a recovery type unseal is used.
The configurer first checks what key type to use before getting it from the key store.
I've also added a fix for canceling any inflight root token generations that are remnants from previous attempts.
Another fix I added was the missing KMS key ID from the decrypt function.


akijakya (approved):
Thanks for the contribution!

ramizpolic (approved):
LGTM, thanks @kirilvasilev!

PR #2798, opened 2024-06-25
chore: bump goreleaser

csatib02:
## Overview

- This PR bumps Goreleaser to 2.0.0

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -16,7 +16,7 @@ before:
 builds:
   - env:
       - CGO_ENABLED=0
-    ldflags: "-s -w -X main.version={{ .Version }} -X main.commitHash={{ .ShortCommit }} -X main.buildDate={{ .Date }}"
+    ldflags: "-s -w -X main.Version={{ .Version }} -X main.commitHash={{ .ShortCommit }} -X main.buildDate={{ .Date }}"
     main: ./cmd/template
     goos:
       - linux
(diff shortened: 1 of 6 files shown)


.goreleaser.yaml
```diff
@@ -21,20 +26,25 @@ builds:
       - arm64
     hooks:
       post:
-        - xgo -targets {{ .Os }}/{{ .Arch }} -dest build/dist -ldflags '-s -w -X main.version={{ .Version }} -X main.commitHash={{ .ShortCommit }} -X main.buildDate={{ .Date }}' -pkg cmd/bank-vaults .
+        - xgo -targets {{ .Os }}/{{ .Arch }} -dest build/dist -ldflags '-s -w -X main.Version={{ .Version }} -X main.commitHash={{ .ShortCommit }} -X main.buildDate={{ .Date }}' -pkg cmd/bank-vaults .
```
ramizpolic, line 29: any reason why this is different?

.goreleaser.yaml
```diff
@@ -21,20 +26,25 @@ builds:
...
         - sudo bash -c "mkdir -p build/dist/bank-vaults_{{ .Os }}_{{ .Arch }}; mv build/dist/github.com/bank-vaults/bank-vaults-{{ .Os }}-{{ .Arch }} build/dist/bank-vaults_{{ .Os }}_{{ .Arch }}/bank-vaults"
 
 archives:
-  - name_template: "{{ .Binary }}-{{ .Os }}-{{ .Arch }}"
-    format_overrides:
-      - goos: windows
-        format: zip
```
ramizpolic, line 31: what happens to windows binary?

cmd/template/main.go
```diff
@@ -24,6 +24,8 @@ import (
 	"github.com/bank-vaults/internal/configuration"
 )
 
+var Version = "v1.31.1"
```
ramizpolic, line 27: i see, this has changed as well. probably better to have dev version as default and only replace on build
```suggestion
var Version = "dev"
```

csatib02: Should I change on all other PR's to dev?

.goreleaser.yaml
csatib02: https://github.com/bank-vaults/secret-sync/pull/211#discussion_r1661226873

.goreleaser.yaml
csatib02: Since we are not building for windows, these lines where unnecessary, I verified it in both `release` and `snapshot` mode.
```yaml
    goos:
      - linux
      - darwin
    goarch:
      - amd64
      - arm64
```

cmd/template/main.go
ramizpolic: yes pleases

ramizpolic (approved):

cmd/template/main.go
akijakya: So... shouldn't it be changed here as well? 🙂

csatib02: https://github.com/bank-vaults/bank-vaults/issues/2832

PR #2826, opened 2024-07-08
chore(build): check calling repository before cosign signing

richardfan1126:
## Overview

This PR adds a check into the artifact signing step of the build process, so the workflow will only sign the artifact (using cosign) when it's called by the repositories within the same GitHub organization.

This fix is to prevent other arbitrary repositories from using the certificate identity `https://github.com/bank-vaults/bank-vaults/.github/workflows/artifacts.yaml@refs/heads/main` to sign their artifacts

## Notes for reviewer

The cosign feature was originally added by this PR: https://github.com/bank-vaults/bank-vaults/pull/2756

### DIFF
--- a/.github/workflows/artifacts.yaml
+++ b/.github/workflows/artifacts.yaml
@@ -121,7 +121,7 @@ jobs:
           # push: ${{ inputs.publish }}
 
       - name: Sign image with GitHub OIDC Token
-        if: inputs.publish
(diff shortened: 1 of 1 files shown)


ramizpolic (approved):
thanks a lot @richardfan1126!

csatib02 (approved):

PR #2840, opened 2024-07-15
fix: don't set static token_reviewer_jwt

### LINKED ISSUE #1603: Unable to pull secrets when vault-configurer is down
**Describe the bug**:
When vault-configurer is down, other pods in the cluster can not pull secrets.
Here is the log from vault pod while the other pods trying to pull secrets and the configurer is not up:

`login unauthorized due to: lookup failed: service account unauthorized; this could mean it has been deleted or recreated with a new token
`

This happens because vault configurer is configuring k8s_auth with `token_reviewer_jwt` of its own (if we aren't set it explicitly) which prevents us using the local **Vault pod** token as reviewer JWT as described [here](https://www.vaultproject.io/docs/auth/kubernetes#use-local-service-account-token-as-the-reviewer-jwt)

https://github.com/banzaicloud/bank-vaults/blob/52cf23528eb30c90c6612d94c046697dcb3f06aa/internal/vault/auth_methods.go#L217-L222

So when the configurer is down, k8s deletes the token used, to configure k8s_auth method so then when Vault k8s auth tries to use the token with the API, it says that token has expired.

The only way to overcome it is to set `kubernetes_host` key and avoid using the default config as you can see here
https://github.com/banzaicloud/bank-vaults/blob/52cf23528eb30c90c6612d94c046697dcb3f06aa/internal/vault/auth_methods.go#L64-L66

**Expected behaviour**:
Pods should be able to pull secrets when configurer is down as well. 
We should not set `kubernetes_host` when running vault on the same cluster.

1 ) Omit `kubernetes_host`, `token_reviewer_jwt` from your kubernetes auth config. 
2) Sync vault configurer 
3) Scale vault configurer replicas to 0
4) Try to pull secrets with k8s_auth

**Additional context**:

**Environment details**:
- Kubernetes version: v1.21.5
- Cloud-provider/provisioner: EKS
- bank-vaults version: 1.15.2

/kind bug

msvticket:
## Overview

Vault supports reading its own /var/run/secrets/kubernetes.io/serviceaccount/token and since version 1.9.0 it can handle short lived tokens. The removed code prevented that handling to occur.

Looking at the compatibility matrix for vault operator I'd say loosing functionality for vault pre 1.9.0 should be fine:

https://github.com/bank-vaults/vault-operator/blob/main/VERSIONS.md

More details on how vault is handling Kubernetes tokens: https://developer.hashicorp.com/vault/docs/auth/kubernetes#kubernetes-1-21

Fixes #1603

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,13 +10,11 @@ require (
 	github.com/aws/aws-sdk-go v1.54.18
 	github.com/bank-vaults/internal v0.3.1
 	github.com/bank-vaults/vault-sdk v0.9.4
-	github.com/cristalhq/jwt/v3 v3.1.0
 	github.com/fsnotify/fsnotify v1.7.0
 	github.com/hashicorp/go-uuid v1.0.3
 	github.com/hashicorp/hcl v1.0.1-vault-5
 	github.com/hashicorp/vault/api v1.14.0
 	github.com/jpillora/backoff v1.0.0
-	github.com/json-iterator/go v1.1.12
 	github.com/miekg/pkcs11 v1.1.1
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/prometheus/client_golang v1.19.1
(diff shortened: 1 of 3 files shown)


ramizpolic (approved):
thanks for submitting a fix for this @msvticket, lgtm

PR #2888, opened 2024-08-07
refactor: general codebase improvements

csatib02:
## Overview

- Mostly style changes, but also replaced some old/deprecated code.

### DIFF
--- a/pkg/kv/gcs/gcs.go
+++ b/pkg/kv/gcs/gcs.go
@@ -42,9 +42,8 @@ func New(bucket, prefix string) (kv.Service, error) {
 }
 
 func (g *gcsStorage) Set(key string, val []byte) error {
-	ctx := context.Background()
 	n := objectNameWithPrefix(g.prefix, key)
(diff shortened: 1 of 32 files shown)


pkg/kv/gcs/gcs.go
```diff
@@ -55,18 +54,15 @@ func (g *gcsStorage) Set(key string, val []byte) error {
 }
 
 func (g *gcsStorage) Get(key string) ([]byte, error) {
-	ctx := context.Background()
 	n := objectNameWithPrefix(g.prefix, key)
-
```
ramizpolic, line 60: TODO: we should extend Get/Set interface to allow passing the context

ramizpolic (approved):
LGTM!

pkg/kv/gcs/gcs.go
csatib02: https://github.com/bank-vaults/bank-vaults/issues/2893

ramizpolic (approved):

PR #2897, opened 2024-08-09
feat: switch from xgo to gorelaser-cross

### LINKED ISSUE #2074: Replace current goreleaser build with goreleaser-cross
The current workflow is rather error prone and had issues in the past.

There is however an "official" cross-compile setup for goreleaser that we could try:

https://github.com/goreleaser/goreleaser-cross

csatib02:
## Overview
- Building of Bank-Vaults was a bit hacky since the pkcs11 package can't be built with simple Go Crosscompiling, so we used xgo, but that is not directly supported in GoReleaser, so we did the actual compilation with xgo in post hooks for all targets.
- This PR introduces [Goreleaser-cross](https://github.com/goreleaser/goreleaser-cross)

Fixes #2074

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -22,8 +22,6 @@ builds:
       - PKG_CONFIG_PATH=/sysroot/macos/amd64/usr/local/lib/pkgconfig
       - CC=o64-clang
       - CXX=o64-clang++
-    flags:
-      - -mod=readonly
     ldflags:
       - -s -w -X main.Version={{.Version}} -X main.commitHash={{.ShortCommit}} -X main.buildDate={{.Date}}
 
@@ -40,8 +38,6 @@ builds:
       - PKG_CONFIG_PATH=/sysroot/macos/arm64/usr/local/lib/pkgconfig
       - CC=oa64-clang
       - CXX=oa64-clang++
-    flags:
-      - -mod=readonly
     ldflags:
       - -s -w -X main.Version={{.Version}} -X main.commitHash={{.ShortCommit}} -X main.buildDate={{.Date}}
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -39,7 +42,25 @@ container-image: ## Build container image
...
+		-v `pwd`:/go/src/$(PACKAGE_NAME) \
+		-w /go/src/$(PACKAGE_NAME) \
+		ghcr.io/goreleaser/goreleaser-cross:${GOLANG_CROSS_VERSION} \
+		--clean --skip=publish --snapshot
+
+.PHONY: release
+release: ## Release the project
+	@docker run \
+		--rm \
+		-e CGO_ENABLED=1 \
```
sagikazarmark, line 58: You will probably need to pass a GITHUB_TOKEN to the container as well.

.goreleaser.yaml
```diff
@@ -10,38 +7,90 @@ dist: build/dist
...
-    hooks:
-      post:
-        - xgo -targets {{ .Os }}/{{ .Arch }} -dest build/dist -ldflags '-s -w -X main.Version={{ .Version }} -X main.commitHash={{ .ShortCommit }} -X main.buildDate={{ .Date }}' -pkg cmd/bank-vaults .
-        - sudo bash -c "mkdir -p build/dist/bank-vaults_{{ .Os }}_{{ .Arch }}; mv build/dist/github.com/bank-vaults/bank-vaults-{{ .Os }}-{{ .Arch }} build/dist/bank-vaults_{{ .Os }}_{{ .Arch }}/bank-vaults"
+    env:
+      - CGO_ENABLED=1
+      - CC=aarch64-linux-gnu-gcc
+      - CXX=aarch64-linux-gnu-g++
+    flags:
+      - -mod=readonly
```
sagikazarmark, line 76: Isn't this the default for go build commands?

Makefile
sagikazarmark, line 58: CGO is enabled above. Does it have to be here as well?

ramizpolic (approved):

PR #2958, opened 2024-09-24
fix: binary release

csatib02:


lgecse (approved):

.github/workflows/artifacts.yaml
```diff
@@ -213,9 +213,9 @@ jobs:
           if_false: make binary-snapshot
 
       - name: Build
-        run: nix develop --impure .#ci -c ${{ steps.build-command.outputs.value }}
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
larsks, line 218: Why does setting the environment variable here not work? Does `nix develop ...` reset the environment?

csatib02: IDK for sure...
There is something to do with Goreleaser-cross, this is the first time we are releasing with that.
According to this example repo: https://github.com/goreleaser/goreleaser-cross-example it reads from an env-file, taking a wild guess it is because with Goreleaser-cross the whole release is done in a distinct container.

https://github.com/bank-vaults/bank-vaults/actions/runs/11016261582/job/30592704732

PR #3308, opened 2025-05-15
Release 1.32.0 dd v1.32.0 dd

pl-jankowskimichal:
## Overview

Fixes #(issue)
1. Extend RaftInitialized method
  If the root key is not stored in the bucket, cluster is not getting up during an initial bootsrap if the root key is not in the gcs bucket.
  With the current code, a second pod will assume that it's actually a
  first one to exist and it will panic once it notices the vault-unseal-0
  key in the bucket.
  
  In this PR we try to fix it by trying to contact
  the vault leader in case the root key doesn't exist -> this is a valid
  config option.


pl-jankowskimichal:
sorry, I will submit an issue and a PR with just a relevant changes

PR #3310, opened 2025-05-15
fix: secondary pod tries to initalize Vault when vault-root key is not stored

pl-jankowskimichal:
## Overview
If the root key is not stored in the bucket, cluster is not getting up during an initial bootsrap if the root key is not in the gcs bucket. With the current code, a second pod will assume that it's actually a first one to exist and it will panic once it notices the vault-unseal-0 key in the bucket.

In this PR we try to fix it by trying to contact
the vault leader in case the root key doesn't exist -> this is a valid config option.

Fixes #(issue)
https://github.com/bank-vaults/bank-vaults/issues/3309

### DIFF
--- a/internal/vault/operator_client.go
+++ b/internal/vault/operator_client.go
@@ -372,21 +372,34 @@ func (v *vault) Init() error {
 }
 
 // RaftInitialized in our case Vault is initialized when root key is stored in the Cloud KMS
+// or when just unseal keys are stored in the Cloud KMS
 func (v *vault) RaftInitialized() (bool, error) {
-	rootToken, err := v.keyStore.Get(keyRootToken)
-	if err != nil {
-		if isNotFoundError(err) {
-			return false, nil
+	if v.config.StoreRootToken {
+		rootToken, err := v.keyStore.Get(keyRootToken)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


pl-jankowskimichal:
Seems like I was a bit to fast with this approach. I need to do a deeper dive into this

csatib02 (approved):
LGTM!

csatib02:
I believe this will also solve: https://github.com/bank-vaults/vault-operator/issues/316

pl-jankowskimichal:
regarding https://github.com/bank-vaults/vault-operator/issues/317 
I think this is an unrelated issue. I'm not sure what kustomization they have applied on top of the `cr-raft-ha-storage.yaml` but it would help us to understand what did go wrong there.

regarding https://github.com/bank-vaults/vault-operator/issues/316
It seems to me this is about a first pod (vault-0) not being able to initialise Vault due to some issues with the seal (wrap) config. So unfortunately IMHO also unrelated.

To satisfy the linters I ended up force pushing -> would you mind re-approving?

csatib02:
@pl-jankowskimichal Thanks so much for checking!

csatib02 (approved):

Issue #3309, opened 2025-05-15
Vault raft bootstrap fails when secondary pod starts and root key is missing from bucket


pl-jankowskimichal:
### Preflight Checklist

- [x] I have searched the [issue tracker](https://www.github.com/bank-vaults/bank-vaults/issues) for an issue that matches the one I want to file, without success.
- [x] I am not looking for support or already pursued the available [support channels](https://bank-vaults.dev/docs/support) without success.
- [x] I agree to follow the [Code of Conduct](https://bank-vaults.dev/docs/code-of-conduct).

### Version

1.31.3

## Description
When initializing a Vault cluster in raft mode, secondary pods can incorrectly attempt to initialize the vault cluster when the root key is not found in the GCS bucket. This leads to initialization failures as the pods detect conflicting state.

in the unseal.go we end up on a code path that is reserved for a first vault instance in the  vault cluster (before vault is initialised)
https://github.com/bank-vaults/bank-vaults/blob/main/cmd/bank-vaults/unseal.go#L112 
the reason we end up on that path is few lines up in the unseal.go (L101) inside the v.RaftInitialized() 
 in here bank vaults only check if the vault root key is stored to figure out if a cluster is initialised https://github.com/bank-vaults/bank-vaults/blob/v1.31.3/internal/vault/operator_client.go#L371 

### Actual Behavior

If the root key is not stored in the bucket during initial bootstrap, the cluster fails to come up
A secondary pod incorrectly assumes it's the first one in the cluster
The pod then panics when it notices the vault-unseal-0 key already exists in the bucket

### Steps To Reproduce

I'm testing a configuraton where bank-vaults is a sidecar in a vault pod (backed by GCS and KMS) running folowing command
```
- bank-vaults
  - unseal
  - --init
  - --raft
  - --raft-leader-address=https://<address>
  - --mode=google-cloud-kms-gcs
  - --google-cloud-kms-crypto-key=<crypto-key>
  - --google-cloud-kms-key-ring=<key-ring>
  - --google-cloud-kms-location=<region>
  - --google-cloud-kms-project=<project>
  - --google-cloud-storage-bucket=<bucket-name>
  - --google-cloud-storage-prefix=<prefix>
  - --store-root-token=false
```
Notice the `--store-root-token=false`

### Logs

```shell
❯ k get po
NAME                   READY   STATUS             RESTARTS          AGE
vault-0                3/3     Running            0                 22h
[...]

PR #3334, opened 2025-06-06
fix: keys may not exist yet

pl-jankowskimichal:
## Overview

Fixes #(issue)

We should not throw error when unseal keys are not found or else first pod can't initialize the Vault Cluster
I've introduced this error in PR #3310. Related to issue #3309


csatib02 (approved):

csatib02:
@pl-jankowskimichal 

Could you please rebase this PR, as we have successfully unclogged the pipeline.

PR #3498, opened 2025-10-06
build with Go 1.24.7 to mitigate CVEs

ddl-r-abdulaziz:
## Overview

Bumped the go-version to 1.24.7

Earlier go 1.24 versions trigger scanners on unrelated high severity CVEs.
Organizations using scanner benefit from recently patched Go versions in their own security reports.


csatib02:
hey @ddl-r-abdulaziz,

We usually bump go-versions around new releases.
We are planning on cutting a new release for every project, around early December.

PR #3621, opened 2026-01-09
fix: prevent substring collision in accessor template replacement

dmount-clari:
## Overview

Fix substring collision bug in `__accessor__` template replacement that occurs 
when mount paths share a common prefix (e.g., `kubernetes`/`kubernetes_cluster`, 
`aws`/`aws2`, `ldap`/`ldap_prod`).

**Root Cause**: The code iterated over the mounts map in undefined order. If a 
shorter path was processed before a longer path with the same prefix, the 
replacement would incorrectly modify the longer path's placeholder.

**Solution**: Sort mount paths by length (longest first) before performing 
replacements, ensuring longer paths are always processed before their prefixes.

### Code Changes
- `internal/vault/policies.go`: Sort mounts by length
- `internal/vault/secrets_engines.go`: Sort mounts by length

### Test Coverage
- `internal/vault/policies_test.go`: Added comprehensive test coverage
  - Substring collision handling (2 test cases)
  - HCL/JSON formatting with accessor replacement (2 test cases)
  - Error handling for invalid policy rules (3 test cases)
- `internal/vault/secrets_engines_test.go`: Added test coverage
  - Substring collision scenarios (4 test cases)

## Notes for reviewer

- Non-breaking change - no API modifications
- No changes needed to existing user configurations
- Comprehensive test coverage including error paths
- Handles any prefix collision pattern (not just kubernetes/kubernetes_cluster)

### DIFF
--- a/internal/vault/policies.go
+++ b/internal/vault/policies.go
@@ -17,7 +17,8 @@ package vault
 import (
 	"fmt"
 	"log/slog"
-	"sort"
+	"maps"
+	"slices"
 	"strings"
 
 	"emperror.dev/errors"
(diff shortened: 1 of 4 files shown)


internal/vault/policies.go
```diff
@@ -32,23 +33,34 @@ type policy struct {
...
+		mountPaths = append(mountPaths, k)
+	}
+	sort.Slice(mountPaths, func(i, j int) bool {
+		return len(mountPaths[i]) > len(mountPaths[j])
+	})
+
+	for index := range policiesConfig {
+		for _, k := range mountPaths {
+			v := mounts[k]
+			policiesConfig[index].Rules = strings.ReplaceAll(policiesConfig[index].Rules, fmt.Sprintf("__accessor__%s", strings.TrimRight(k, "/")), v.Accessor)
```
csatib02: The whole section is rather old, and can be made much cleaner:

```go
for i := range policiesConfig {
	policy := &policiesConfig[i]
	
	// Replace accessor placeholders
	for _, mountPath := range mountPaths {
		placeholder := fmt.Sprintf("__accessor__%s", strings.TrimSuffix(mountPath, "/"))
		policy.Rules = strings.ReplaceAll(policy.Rules, placeholder, mounts[mountPath].Accessor)
	}

	// Format as HCL, falling back to original if it's valid JSON
	formatted, err := hclPrinter.Format([]byte(policy.Rules))
	if err != nil {
		if _, parseErr := hcl.Parse(policy.Rules); parseErr != nil {
			return nil, fmt.Errorf("parsing %s policy rules: %w", policy.Name, parseErr)
		}
		slog.Debug("could not HCL-format policy rules (may be JSON)", 
			"policy", policy.Name, 
			"error", err)
		formatted = []byte(policy.Rules)
	}
	policy.RulesFormatted = string(formatted)
}
```

internal/vault/secrets_engines.go
```diff
@@ -64,7 +65,18 @@ type secretEngine struct {
...
+	mountPaths := make([]string, 0, len(mounts))
+	for k := range mounts {
+		mountPaths = append(mountPaths, k)
+	}
+	sort.Slice(mountPaths, func(i, j int) bool {
+		return len(mountPaths[i]) > len(mountPaths[j])
+	})
+
+	for _, k := range mountPaths {
+		v := mounts[k]
```
csatib02: Same simplifications can be considered there too:

```go
func replaceAccessor(input string, mounts map[string]*api.MountOutput) string {
	// Sort mount paths by length (longest first) to avoid substring collisions
	// e.g., "__accessor__kubernetes_cluster" should be replaced before "__accessor__kubernetes"
	mountPaths := maps.Keys(mounts)
	slices.SortFunc(mountPaths, func(a, b string) int {
		return len(b) - len(a)
	})

	result := input
	for _, mountPath := range mountPaths {
		placeholder := fmt.Sprintf("__accessor__%s", strings.TrimSuffix(mountPath, "/"))
		if strings.Contains(result, placeholder) {
			slog.Info("replaced __accessor__ placeholder",
				"placeholder", placeholder,
				"accessor", mounts[mountPath].Accessor)
			result = strings.ReplaceAll(result, placeholder, mounts[mountPath].Accessor)
		}
	}

	return result
}
```

csatib02:
LGTM!
Thanks @dmount-clari!

PR #3795, opened 2026-05-05
feat: add key prefix support for Azure Key Vault backend

steve-sgnl:
## Overview

Add an optional `--azure-key-vault-prefix` flag that prepends a configurable string to all secret names stored in Azure Key Vault. This enables multiple independent consumers to share a single Key Vault instance without secret name collisions.

**Use case:** When running multiple Vault clusters that each store unseal/recovery keys in the same Azure Key Vault, the fixed secret names (`vault-recovery-0`, `vault-root`, etc.) collide. A prefix like `cluster-a-` produces `cluster-a-vault-recovery-0`, avoiding overwrites.

**Implementation:** Follows the exact pattern already used by the S3, GCS, OCI, and Alibaba OSS backends. The prefix is concatenated directly with the key name (no separator added automatically), giving users full control over the resulting secret name format.

When the prefix is empty (the default), behavior is identical to the current implementation.

### Changes

| File | Change |
|------|--------|
| `pkg/kv/azurekv/keyvault.go` | Add `prefix` field to struct, prepend in `Get`/`Set` |
| `cmd/bank-vaults/main.go` | Add `--azure-key-vault-prefix` CLI flag |
| `cmd/bank-vaults/common.go` | Pass prefix to `azurekv.New()` |

### Companion change needed

The [vault-operator](https://github.com/bank-vaults/vault-operator) `AzureUnsealConfig` struct will need a `keyPrefix` field and corresponding `--azure-key-vault-prefix` argument wiring. Happy to open that PR once this is reviewed.

## Notes for reviewer

- Azure Key Vault secret names only allow `[a-zA-Z0-9-]` (max 127 chars). Users are responsible for choosing a valid prefix. This matches how the S3 prefix works (no validation on the prefix value itself).
- No separator is injected between prefix and key, consistent with `objectNameWithPrefix` in the S3 and GCS backends.
- Default empty prefix means zero behavioral change for existing users.

### DIFF
--- a/cmd/bank-vaults/common.go
+++ b/cmd/bank-vaults/common.go
@@ -185,7 +185,7 @@ func kvStoreForConfig(ctx context.Context, cfg *viper.Viper) (kv.Service, error)
 		return multi.New(services), nil
 
 	case cfgModeValueAzureKeyVault:
-		akv, err := azurekv.New(cfg.GetString(cfgAzureKeyVaultName))
+		akv, err := azurekv.New(cfg.GetString(cfgAzureKeyVaultName), cfg.GetString(cfgAzureKeyVaultPrefix))
 		if err != nil {
 			return nil, errors.Wrap(err, "error creating Azure Key Vault kv store")
 		}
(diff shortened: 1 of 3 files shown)


steve-sgnl:
Companion operator PR: bank-vaults/vault-operator#1071

csatib02 (approved):

